PR #4, opened 2023-01-29
resizing fullscreen elements

wxwisiasdf:


schombert:
I actually had resize implemented like this at one point, but it isn't the direction I want to go. Also, the is_fullscreen flag doesn't mean "cover the entire screen" for the victoria 2 gui files. Yes, I know it's weird, but I didn't write them.

PR #17, opened 2023-02-04
Directly use the element under mouse on mouse actions

nuruvilu:


nuruvilu:
I just tested this again and it broke the minimap, will fix.

PR #44, opened 2023-02-18
Made zooming consistent (again) and adjusted zoom speed and speed of camera movement when zooming. Also (most likely) fixed 'inverted map' bug (again).

Masterchef118:


Masterchef118:
Additional note: This PR also makes some changes and additions to the Production window and related gui files.

PR #55, opened 2023-02-23
Use filesystem system to compose filenames

wxwisiasdf:


schombert:
This will result in the file being opened and closed just to create the name, which is very inefficient, especially since an antivirus will, if present, hook one of those calls.

PR #60, opened 2023-02-25
Non-vanilla flag support

wxwisiasdf:
GFM uses flags like `_special` and `_theocracy` and such wacky shit, so why not add support for those?


schombert:
There is one major problem and one minor problem with this. The major problem is the block you removed in culture.cpp
```
auto id = state.world.national_identity_get_government_flag_type(
		state.world.nation_get_identity_from_identity_holder(target_nation),
		gov_type);
	if(id != 0)
		return flag_type(id - 1);
```
This is important for making flags work. Some tags have which flag they used for which government type set to do something unique. If you ignore it, places like France will show up with the wrong flag. So not replacing this logic with something equivalent is a show-stopper.

Secondly, you can't just comment out lines in the tests when you make changes. You need to update the tests to work with whatever changes you want to make.

You may also want to consider whether just adding theocracy and special to the enum is easier. It is certainly going to be more performant

PR #63, opened 2023-02-27
Improve support for various mods

wxwisiasdf:
This should make most of the popular mods work properly - especially some which used heavy resources [MHM moment, GFM moment].

### DIFF
--- a/docs/tags_and_flags.md
+++ b/docs/tags_and_flags.md
@@ -8,6 +8,14 @@ Remember that a nation in the game is not, for us, the same thing as a tag. To g
 
 ### Flags
 
+First, the parser will add all the flag types that are referenced - this will result in `state::flag_type_map` having all of the possible flag_types that were loaded on this scenario. After that it will allocate the proper number of flags to accomodate all of the flag types.
+
+To obtain the dynamically-assigned Id for an specific type of flag, use the following function:
+```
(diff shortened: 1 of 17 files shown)


schombert:
I think it is time to impose some sanity limits: redirect monarchy2 and republic2 inside the parser to the standard monarchy and republic flags. Redirect cosmetic_1, cosmetic_2,  and spare to somewhere sensible.

wxwisiasdf:
> I think it is time to impose some sanity limits: redirect monarchy2 and republic2 inside the parser to the standard monarchy and republic flags. Redirect cosmetic_1, cosmetic_2, and spare to somewhere sensible.

The mods which use monarchy2 and republic2 have them **separate from republic and monarchy**

schombert:
Yes, I understand that, I am just not convinced that it is worth respecting. Would anything important be lost if those alternate monarchy and republic types just had to have the base monarchy and republic flags?

schombert:
I think that "attrition" wouldn't do anything if you added it to a leader trait in the base game. And unless the base game supports it, I am disinclined to add it

schombert:
The same goes for adding extra unit stats

wxwisiasdf:
Attrition is supported as a trait modifier in the base game iirc.

wxwisiasdf:
We can live without TCG support for now...

PR #75, opened 2023-03-06
Improve support for HFM/TGC

wxwisiasdf:
Now without unit modifiers! :D


schombert:
You can't just map years and days to months directly because then the difference between them would be lost. So instead of having an obvious error when loading the scenario, we would have a bunch of hidden breakage

wxwisiasdf:
That's fair, what they should be mapped into then?

schombert:
It depends on the context they are used in the files, which I cannot see. If different durations are used for the same class of modifier, they they should all be normalized to some common interval, as if they were all specified in months or some other, well-documented, common unit of time. I will also add that this work is much less important than fixing the failing test.

PR #79, opened 2023-03-08
Add stripes to map

bruno-cabrera98:
The main change is that the province color texture was changed to a Texture Array of 2 textures, the province color and the stripes color.

In gen_prov_color_texture (map.cpp) the prov_color vector was kept, but it contains the colors for both textures, First for province color and then stripe colors.

I'm not quite sure if using a texture array of only 2 textures is better or worse in terms of performance (compared to 2 separated textures).

Initially I thought that with 4 iterations of the stripe texture you could fill the image. I wanted to do 4 color textures to make the map coloring more generic, one for each stripe. But this doesn't work, because the stripes don't have the same size.

For now I modified the culture map to use the stripes, but I'll also need to modify the other map modes


bruno-cabrera98:
Hi! Thanks for the input. I would rather keep the texture for now. But next time I work on the shaders again I'll change it

PR #91, opened 2023-03-16
Add function to convert path slashes between OSes

nuruvilu:


nuruvilu:
Oops I forgot to change that one `std::string`

PR #106, opened 2023-03-23
Fixed a crash caused by the console window asking for a font that doesn't exist.

Masterchef118:


Masterchef118:
*When using classic fonts.

PR #122, opened 2023-03-28
Add support for various continents

wxwisiasdf:
Since most mods add new continents, like TTA and HFM.
It would be wise to have them be identified as IDs instead of adding the extreme overhead of storing them as strings since most mods use the same set of continents (TTA being the exception).


schombert:
There is a reason that I stored the existing continents explicitly that has nothing to do with the ui. The game has special logic concerning which nations automatically count as interested in a crisis that references the existing continents. So while there is no need to save any new continents in this way, I do need the base continents. Thus, I cannot accept a PR that removes them

schombert:
Additionally, there is no need to store a name to continent map or to filter out continents to a specific list. The modifier that each continent creates already stores its name, and so the needs of the ui can be met simply by, on creation of the element, scanning the modifiers for one with a matching text value inside its name *key* and storing that. Since this would only have to be done once upon creating the ui, the time cost is trivial.

PR #138, opened 2023-04-05
Added good filtering buttons to Production factory window.

Masterchef118:


schombert:
I'm happy to see that the bad filtering buttons have been upgraded

PR #158, opened 2023-04-11
fill diplomacy window and some extra stuff on technology

wxwisiasdf:
fill diplomacy window and some extra stuff on technology tab

### DIFF
--- a/src/gamestate/dcon_generated.txt
+++ b/src/gamestate/dcon_generated.txt
@@ -1764,6 +1764,11 @@ object {
 		type{ sys::date }
 		tag{ save }
 	}
+	property{
+		name{ name }
+		type{ text_sequence_id }
+		tag{ save }
+	}
 }
 object {
 	name{ wargoal }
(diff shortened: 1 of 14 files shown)


schombert:
I have some concerns about this:
```
template<uint16_t Rank>
class nation_gp_opinion_text : public standard_nation_text {
public:
	std::string get_text(sys::state& state) noexcept override {
		dcon::nation_id gp_id{};
		state.world.for_each_nation([&](dcon::nation_id id) {
			auto fat_id = dcon::fatten(state.world, id);
			if(fat_id.get_rank() == Rank)
				gp_id = id;
		});
		float influence = 0.f;
		state.world.for_each_gp_relationship([&](dcon::gp_relationship_id id) {
			auto fat_id = dcon::fatten(state.world, id);
			if(fat_id.get_great_power() == gp_id)
				if(fat_id.get_influence_target() == nation_id)
					influence += fat_id.get_influence();
		});
		return std::to_string(int32_t(influence));
	}
};
```
You can look  up the nth ranked GP directly by consulting the sorted list of GPs. And similarly, you can look up the gp relationship between a GP and a nation directly rather than iterating over all the GP relationships in the world

schombert:
Also, as a minor note, if you change `dcon_generated.txt` you typically *don't* have to bump the save file version. Its serialization is robust enough that, if you add a new property, loading an old file will result in the new property being zero initialized. In your case, it would leave the names as invalid text sequence handles. However, bumping the version doesn't really hurt anything, so on the other hand "better safe than sorry" is perfectly fine too

PR #165, opened 2023-04-12
unit_start_experience modifier, print scope types for easier debugging, make national focuses intake modifier stuff

wxwisiasdf:
- unit_start_experience modifier
hfm uses it, basically land_unit_start_experience_modifier and naval_unit_start_experience_modifier but with a single statment

- print scopes for easier debugging
self explainatory

- make national focuses intake modifier stuff
removes some overhead and allows the national_focuses to be able to be tooltip'ed by whatever make_modifier_tooltip or whatnot...

### DIFF
--- a/src/gamestate/dcon_generated.txt
+++ b/src/gamestate/dcon_generated.txt
@@ -3411,7 +3411,7 @@ object {
 	storage_type{ contiguous }
 	size{ 64 }
 	tag{ scenario }
-
+	
 	property{
 		name{ name }
 		type{ text_sequence_id }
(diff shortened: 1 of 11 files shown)


schombert:
A- Broke the build :)
B - Don't add a new modifier; if you really want unit_start_experience modifier to affect both, then when parsing it have it add both land_unit_start_experience_modifier and naval_unit_start_experience_modifier to the modifier under construction

schombert:
C- if you are going to implement tf_is_accepted_culture_nation, you should implement its evaluation condition properly (follow the tf_is_accepted_culture_state to see what needs to be done) OR make its display function display its actual condition (i.e. copy how "never" works). Saying that the trigger does one thing in a tooltip and implementing it to do another is not great

PR #175, opened 2023-04-14
remove tax efficiency duplicate modifier & use macros to avoid code duplication

wxwisiasdf:
less places to update for modifiers ;)


schombert:
I am going to leave this in limbo for a bit, because I already have two of the modifiers removed as part of a large-ish set of uncommitted changes. After I push that up, you can tweak this to make it compatible :)

schombert:
War name needs to be a `text_sequence_id`. As I pointed out previously, the names for the wars are generated from the content of the csv files, which are referenced by `text_sequence_id`

PR #189, opened 2023-04-15
use unit names indices for more space for unit names (fixes HFM crash)

wxwisiasdf:


schombert:
If you are going to remove my use of 16 byte alignment to save space, you should probably also adjust the code that makes the strings 16 byte aligned: 
```
unit_names.resize(start + length + (16 - length % 16), char(0));
```

schombert:
This won't actually work if you try rebuilding the scenario with it in place; you haven't left space for the terminating null

PR #190, opened 2023-04-16
inventions listbox and technology effects

wxwisiasdf:


schombert:
I guess I should have caught this earlier, but COMMODITY_MOD_DESCRIPTION is a really poor use of a macro. Use a local lambda expression if you must, but what you have now is virtually undebuggable if there is ever a problem with it.

PR #195, opened 2023-04-17
be able to pass mod files thru command line (linux only)

wxwisiasdf:


schombert:
While this is fine for your private use, it will create problems if the saved scenario file wasn't created with the same combination of mods. So I can't accept this as a general addition.

PR #197, opened 2023-04-17
fix mods placed somewhere outside the vanilla 2 folder

wxwisiasdf:


schombert:
Are you sure that this won't break the "replacement" logic? Also, why do we need to support mods that aren't in the mods folder?

wxwisiasdf:
I'm sure it doesn't break the replacement logic because I tested it beforehand with mods - I've recently discovered that mods can override other mods (that stack replacements upon each other) - for example:

game has folder F
mod A overrides folder F
mod B overrides folder F
hence, mod B gets to replace folder F

Supporting mods that aren't in the mods folder is useful mainly for not having to place mods on the game directory necessarily.

schombert:
I don't understand what this:
``
native_string mod_path = root;
mod_path += NATIVE_DIR_SEPARATOR;
mod_path += simple_fs::correct_slashes(simple_fs::utf8_to_native(context.path));
add_root(fs, mod_path);
``
is trying to accomplish. Unless the mod is located within every prior mod, this is going to create a lot of roots that point to nowhere in the file system

wxwisiasdf:
i think thats how it worked in the original game

schombert:
That can't be how it worked in the original game because that would result in the number of roots, and hence the number of places that need to be checked for each file, growing exponentially with the number of mods added. (each mod you add will double the number of roots)

PR #202, opened 2023-04-18
add national,province and major events window

wxwisiasdf:
no rotation for the left flag on the major event variant of the national event widget
```c++
auto new_elm = ui::make_element_by_type<ui::national_event_window<true>>(*this, "event_major_window");
Cyto::Any payload = ui::national_event_data_wrapper(dcon::national_event_id{42});
new_elm->impl_set(*this, payload);
new_elm->impl_on_update(*this);
ui_state.root->add_child_to_front(std::move(new_elm));
```
^sample instantiation


schombert:
But this `ui::make_element_by_type<ui::national_event_window<true>>(*this, "event_major_window");` can't work as a general solution. UI elements cannot be deleted, so if we make a new window every time there is a new event we will eventually end up with an infinite number of windows. Either: make only one event visible at a time and add left and right buttons on the event window to cycle through events OR create a pool that event windows can be drawn from before we try to create new ones, and write the machinery to close event windows by returning them to the pool

schombert:
There are two issues here. One minor problem is that "new" events are going to retain their old order: by just making them visible they won't necessarily appear on top. A second minor problem is that by leaving them in place, but invisible, we will still be wasting some cycles with a potentially unbounded number of invisible event windows. To fix both of these issues you could do the following: make the pool a vector of unique pointers as well and shuffle the unique pointer back and forth between the ui container and the pool when the window needs to be visible (the code on the ui side to do this--i.e. to pull the unique pointer out of the ui container so that you can store it somewhere else--already exists).

PR #205, opened 2023-04-20
fix console crashes + new commands

wxwisiasdf:


schombert:
I can't accept this with any commands that make changes to the game state. Changes to the game state have to be synchronized with the update function

PR #216, opened 2023-05-01
disable exceptions and rtti on main project (since they are never and…

wxwisiasdf:
… would never be used) - except for tbb


schombert:
Since I don't care about the linux build, I can pull this in as is. However, you may want to consider imitating the windows build, which has them off only on release

PR #220, opened 2023-05-07
discard bogus triggers to avoid crashes

wxwisiasdf:


schombert:
This won't do what you want. It will discard *all* triggers as soon as an error is encountered anywhere, whether it has to do with the trigger or not

schombert:
`err.fatal = trigger_err.fatal;` may unset a previous fatal error.

schombert:
I have also realized that, if you replace a trigger with another random trigger, that given that you may end up passing nation ids where the trigger expects provinces, you will probably cause a crash or corruption that way, so this has to be closed.

PR #221, opened 2023-05-07
Add linux cmake build presets for Visual Studio Code 

Makkkkus:
Requires the user to have the cmake-tools extension installed.

Made building the project (assuming you have all the dependencies) just be a matter of choosing a development profile and building. Currently there are two profiles, x64 release and x64 debug, both built with Clang.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,8 +6,9 @@ charset = utf-8
 insert_final_newline = true
 trim_trailing_whitespace = true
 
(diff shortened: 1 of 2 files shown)


schombert:
This may have to wait until Erik gets the build issues sorted out on linux

coderguy57:
Looks good. Can be used for the people that want to presets on linux and disabled for people that don't.

PR #223, opened 2023-05-07
Configure vscode c++ extension to use cmake

Makkkkus:
Don't know if this is okay for everyone, but I'll give it a shot.


schombert:
I have thought about it, but for the same reason I don't push anything in the .vs directory to git (the visual studio settings directory), I don't think we should push anything in the .vscode directory. You can add an additional exclusion rule and push that if our existing git rules aren't doing their job properly.

PR #224, opened 2023-05-08
extra flag types for coe-roi and aoe

wxwisiasdf:
some new flags to accommodate the new government types introduced
(and i thought i added most of the flag_types already - apparently not)


schombert:
So is `flag_count` useless now? It seems like it either needs to be updated or removed

PR #227, opened 2023-05-09
increment space for mods

wxwisiasdf:
fixes crash with HTL


schombert:
Increasing the culture group ids past 126 will result in them taking 2 bytes, because the sign bit needs to be reserved

PR #230, opened 2023-05-10
fix crash for mod

wxwisiasdf:
i am so sorry performance, but some sacrifices needs to be done, modding is the most important part of human civilization :>


schombert:
What is pop_satisfaction_wrapper_id ?

PR #245, opened 2023-05-14
fix message settings window

wxwisiasdf:


schombert:
` factory_icon->frame = uint16_t(fat_id.get_type().get_output().id.index());` won't produce the correct output icon. You need to used `commodity_get_icon`

PR #250, opened 2023-05-16
Fix unrecognized colour escape codes on tooltips

wxwisiasdf:


schombert:
I already have a better fix for `!` in flight that this will conflict with. You can add lower case `g` and such when it goes in next

PR #256, opened 2023-05-18
fix 64k text limit being overpassed and add gray,blue colours

wxwisiasdf:


src/text/text.cpp
```diff
@@ -33,71 +39,84 @@ namespace text {
...
 					if(uint8_t(*pos) == 0xA7) {
 						if(section_start != pos) {
 							auto added_key = state.add_to_pool(std::string_view(section_start, pos - section_start));
 							state.text_components.emplace_back( added_key );
 						}
-						section_start = pos + 2;
-						if(pos + 1 < seq_end)
-							state.text_components.emplace_back( char_to_color(*(pos + 1)) );
-						pos += 1;
+						section_start = pos += 1;
```
schombert, line 54: This sort of thing (doing assignment and then using the return value) is a real bad habit. I know it may sound petty, but it is. Break it up into two lines (and in the other places as well)

PR #260, opened 2023-05-18
Fix diplomacy window country name

coderguy57:
I think this is how generic_settable_element is suppose to work. In that case it is a fix for that too.


coderguy57:
Good job!

PR #263, opened 2023-05-19
general ui improvments

wxwisiasdf:


src/gui/topbar_subwindows/diplomacy_subwindows/gui_diplomacy_actions_window.hpp
```diff
@@ -447,7 +472,10 @@ class diplomacy_action_command_units_button : public generic_settable_element<bu
 	    } else {
 		    text::localised_format_box(state, contents, box, std::string_view("dip_enough_diplo"));
 		    text::add_line_break_to_layout_box(contents, state, box);
-		    text::localised_format_box(state, contents, box, std::string_view("diplomacy_ai_acceptance"));
+
+            text::substitution_map ai_map{};
+            text::add_to_substitution_map(ai_map, text::variable_type::country, std::string_view(text::produce_simple_string(state, state.world.nation_get_name(content))));
```
schombert, line 477: This, and the other places you have done this, will result in a dangling pointer: the temporary string will be destroyed before it can be used in the substitution map. (There is a reason it won't compile with a std::string by default -- that was to prevent you from doing this accidentally). Fortunately, you can stick nation names in substitution maps directly. Ask B--- for more details.

PR #265, opened 2023-05-19
better hline divisors that don't go around the tooltip mindlessly

wxwisiasdf:


Breijz:
No

src/text/text.hpp
```diff
@@ -220,8 +227,10 @@ namespace text {
 		alignment align = alignment::left;
 		text_color color = text_color::white;
 	};
+
 	struct layout {
 		std::vector<text_chunk> contents;
+		std::vector<layout_divider> dividers;
```
schombert, line 233: This won't work if the layout needs to rearrange the text, for example if the layout turns into more than one column

PR #270, opened 2023-05-20
add support fur state_province_id in wargoals

wxwisiasdf:


src/parsing/parsers_declarations.cpp
```diff
@@ -2410,6 +2410,11 @@ void history_war_goal::casus_belli(association_type, std::string_view value, err
 	}
 }
 
+void history_war_goal::state_province_id(association_type, int32_t value, error_handler& err, int32_t line, war_history_context& context) {
+	auto amid = context.outer_context.state.world.province_get_abstract_state_membership_as_province(dcon::province_id(dcon::province::value_base_t(value)));
```
schombert: This will not work. Province ids in the files do not convert 1 to 1 to internal province ids

src/parsing/parsers_declarations.cpp
```diff
@@ -2410,6 +2410,11 @@ void history_war_goal::casus_belli(association_type, std::string_view value, err
...
+	state_id = context.outer_context.state.world.abstract_state_membership_get_state(amid);
```
schombert, line 2415: Also this may not make sense since state instances are not necessarily created at this point

wxwisiasdf: state_province_id points to the capital province of the state for the cassus belli, namely the "obtain back core" cb which i found and prompted me to add this in the first place

PR #272, opened 2023-05-20
parser generator only generate prototypes if needed & fix identation

wxwisiasdf:
- only generate function prototypes to reduce redundancies
- fix identation
- a slight comptime increase

### DIFF
--- a/ParserGenerator/parser_generator.cpp
+++ b/ParserGenerator/parser_generator.cpp
@@ -582,27 +582,42 @@ std::string construct_match_tree_outer(auto const& vector, auto const& generator
 void file_write_out(std::fstream& stream, std::vector<group_contents>& groups) {
 	//	process the parsed content into the generated file
 	std::string output;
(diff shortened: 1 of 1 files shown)


schombert:
What are the benefits of this changes? I am very hesitant to change the parser generator since it seems to be working fine

wxwisiasdf:
Mainly that it stops generating otherwise unnecessary code, it isn't much, but it shaves off around 400 lines of code - which the compiler might like.

schombert:
Have you actually measured a difference with the compiler diagnostics?

wxwisiasdf:
the difference in timings is pretty insignificant and around the 1-10 ms - so the timings most likely got lost in the noise of `-ftime-trace`. not a lot of justification for adding this change however, but it just makes it slightly more readable for others

schombert:
Then it doesn't seem worth it to me. As a general rule of thumb, only mess with things that are already working when you have to

PR #278, opened 2023-05-21
OwO *Notices more tooltips*

Breijz:


src/economy/economy.cpp
```diff
@@ -60,8 +60,18 @@ bool nation_is_constructing_factories(sys::state& state, dcon::nation_id n) {
 	// TODO
 	return false;
 }
-bool nation_has_closed_factories(sys::state& state, dcon::nation_id n) {
-	// TODO
+bool nation_has_closed_factories(sys::state& state, dcon::nation_id n) {	// TODO - should be "good" now
+	auto nation_fat = dcon::fatten(state.world, n);
+	for(auto staat_owner : nation_fat.get_state_ownership()) {
```
schombert, line 65: You don't need to iterate over states. Since factories are secretly located in provinces, you can directly iterate over the owned provinces of a nation to find all the factories

src/economy/economy.cpp
```diff
@@ -60,8 +60,18 @@ bool nation_is_constructing_factories(sys::state& state, dcon::nation_id n) {
...
+		for(auto state : staat_owner.get_state().get_definition().get_abstract_state_membership()) {
+			for(auto factloc : state.get_province().get_factory_location()) {
```
schombert, line 67: Also, this won't work because an abstract state can be split between two or more state instances owned by different nations. Not all the provinces in a state may belong to the same nation. Fortunately, `province.hpp` provides a function to iterate over all the provinces in a particular state instance, if you ever need it

PR #285, opened 2023-05-22
decrease relation command

wxwisiasdf:


schombert:
`p.type = command_type::make_leader;`
Probably won't work properly 😅
```
if(rel && state.world.diplomatic_relation_get_value(rel) <= 200.f)
        return false; // Maxxed out
```
Will never allow you to decrease relations
This:
```
auto rel = state.world.get_diplomatic_relation_by_diplomatic_pair(source, target);
    if(!rel)
        rel = state.world.force_create_diplomatic_relation(source, target);
```
is unnecessary -- see how the existing adjust relations function works

PR #292, opened 2023-05-25
release nations window fixed, moveable windows, election window draft, hold elections button

wxwisiasdf:


src/gui/topbar_subwindows/diplomacy_subwindows/gui_diplomacy_actions_window.hpp
```diff
@@ -631,10 +631,12 @@ class diplomacy_action_discredit_button : public button_element_base {
 
 			// TODO: Conditions for enabling/disabling
 			disabled = false;
-			if(content == state.local_player_nation)
-				disabled = true;
-			else if(!nations::is_great_power(state, state.local_player_nation))
+			if(content == state.local_player_nation
+			|| !command::can_discredit_advisors(state, state.local_player_nation, content, dcon::fatten(state.world, content).get_in_sphere_of().id)
+			|| !nations::is_great_power(state, state.local_player_nation)
+			|| nations::is_great_power(state, content)) {
```
schombert, line 637: You don't need to check for nations being great powers, because the command already does

src/gui/topbar_subwindows/gui_production_window.hpp
```diff
@@ -12,6 +12,57 @@
 
 namespace ui {
 
+enum class prod_good_sort : uint8_t {
```
schombert: This has to be changed because mods can change which commodities there are

PR #297, opened 2023-05-26
various ui improvments and fixes

wxwisiasdf:
- fix alignment on projects window with inputs
- show construction factories, the upgrading of factories and the factories normally
- influence diplomacy actions now show a window instead of immediately doing action
- fixed bug with gp selection on diplo window not being used as the affected_gp in influence actions
- fixed crash with console
- fixed colours of budget window
- various production disable buttons if not used
- select/deselect all commodity filters

### DIFF
--- a/src/economy/economy.hpp
+++ b/src/economy/economy.hpp
@@ -137,9 +137,15 @@ struct new_factory {
 	float progress = 0.0f;
 	dcon::factory_type_id type;
 };
-
 template<typename F>
 void for_each_new_factory(sys::state& state, dcon::state_instance_id s, F&& func); // calls the function repeatedly with new_factory as parameters
+
+struct upgraded_factory {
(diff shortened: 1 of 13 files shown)


src/economy/economy.hpp
```diff
@@ -137,10 +137,16 @@ struct new_factory {
...
-
 template<typename F>
 void for_each_new_factory(sys::state& state, dcon::state_instance_id s, F&& func); // calls the function repeatedly with new_factory as parameters
 
+struct upgraded_factory {
+	float progress = 0.0f;
+	dcon::factory_type_id type;
+};
+template<typename F>
+void for_each_upgraded_factory(sys::state& state, dcon::state_instance_id s, F&& func); // calls the function repeatedly with new_factory as parameters
```
schombert, line 148: comment is inaccurate

PR #317, opened 2023-06-01
XAC* parsing

wxwisiasdf:
- Not fully decoded
- who thought this was a good idea


schombert:
While I could be convinced to merge in a fully working model solution, I am not willing to merge in code I suspect will remain vestigial. For now this should live in its own fork.

PR #322, opened 2023-06-02
fix scrolling issues and clicking issues on various widgets (esp. in the population window)

wxwisiasdf:
before -> the topbar was fixed, somewhat, but the issue persists through most of the tabs
after -> no issues with listboxes and a handful of annoying scrolling stuff fixed (and clicking too)


wxwisiasdf:
This also fixes for non-tooltip elements [see ledger]

PR #331, opened 2023-06-02
 fold and/or scopes

wxwisiasdf:


src/parsing/trigger_parsing.cpp
```diff
@@ -593,16 +614,13 @@ int32_t simplify_trigger(uint16_t *source) {
 	assert(
 	    (0 <= (*source & trigger::code_mask) && (*source & trigger::code_mask) < trigger::first_invalid_code) || (*source & trigger::code_mask) == trigger::placeholder_not_scope);
 	if ((source[0] & trigger::code_mask) >= trigger::first_scope_code) {
+		// simplify each member
+		auto source_size = fold_and_or_trigger(source);
```
schombert, line 618: This should not be run here

src/parsing/trigger_parsing.cpp
```diff
@@ -637,9 +655,6 @@ int32_t simplify_trigger(uint16_t *source) {
 				source_size -= 2;
 			}
 		}
-
-		// MISSING OPTIMIZATION: fold inner 'and' scope into outer 'and' scope, inner 'or' scope into outer 'or' scope
```
schombert, line 641: Instead it should run here. This is because it needs the removal of extraneous negation to run properly

PR #335, opened 2023-06-03
expand 64k limit for triggers by using indices

wxwisiasdf:


schombert:
How much more space does the data actually need? What is the performance cost of the additional indirection (There is a performance benchmark for how long events take to process, you could probably use that to do measurements. But don't forget that you need to measure **in release**)

PR #341, opened 2023-06-03
"which modifiers affect this" tooltip

wxwisiasdf:
only for research points atm

### DIFF
--- a/src/gui/gui_modifier_tooltips.cpp
+++ b/src/gui/gui_modifier_tooltips.cpp
@@ -50,12 +50,11 @@ void modifier_description(sys::state& state, text::layout_base& layout, dcon::mo
 
 	const auto& prov_def = fat_id.get_province_values();
 	for(uint32_t i = 0; i < prov_def.modifier_definition_size; ++i) {
-		if(prov_def.values[i] == 0.f)
-			continue;
+		if(!bool(prov_def.offsets[i]))
+			break;
 
 		auto offset = uint32_t(prov_def.offsets[i].index());
 		auto data = province_modifier_names[offset];
-
 		auto box = text::open_layout_box(layout, indentation);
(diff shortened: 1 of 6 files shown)


schombert:
I assume that this doesn't have the province modifiers in yet?

src/gui/gui_modifier_tooltips.cpp
```diff
@@ -50,41 +50,185 @@ void modifier_description(sys::state& state, text::layout_base& layout, dcon::mo
...
+			break;
+		if(prov_def.offsets[i] != pmid)
+			continue;
+		
+		auto data = province_modifier_names[pmid.index()];
+		auto box = text::open_layout_box(layout, indentation);
+		text::add_to_layout_box(layout, state, box, text::produce_simple_string(state, fat_id.get_name()), text::text_color::white);
+		text::add_to_layout_box(layout, state, box, std::string_view{":"}, text::text_color::white);
+		text::add_space_to_layout_box(layout, state, box);
+		auto value = prov_def.values[i] * scaled;
```
schombert, line 98: This is the actual value that you should show, but that is not what you send to the text box. Also, if the scale is 0 the modifier should be skipped completely

src/gui/gui_topbar.hpp
```diff
@@ -325,29 +325,33 @@ class topbar_nation_daily_research_points_text : public nation_daily_research_po
 			auto nation_id = any_cast<dcon::nation_id>(payload);
 
 			auto box = text::open_layout_box(contents, 0);
-			bool bClergyPassed = false; // Ugly fix but it should work -breizh
+			bool clergy_passed = false; // Ugly fix but it should work -breizh
```
schombert, line 328: renaming this is doing God's work

PR #347, opened 2023-06-04
military window work

wxwisiasdf:
- disable not activated units
- show units that are being constructed
- rework so `set` is gone foreva

### DIFF
--- a/src/gui/topbar_subwindows/military_subwindows/gui_build_unit_large_window.hpp
+++ b/src/gui/topbar_subwindows/military_subwindows/gui_build_unit_large_window.hpp
@@ -25,48 +25,54 @@ public:
 
 class unit_build_button : public button_element_base {
 public:
-	dcon::unit_type_id unit_type_id;
-	dcon::culture_id culture_id;
-	dcon::province_id province_id;
-	// false == army
-	// true == navy
 	bool is_navy = false;
 	void button_action(sys::state& state) noexcept override {
-		if(is_navy == false) {
-			command::start_land_unit_construction(state, state.local_player_nation, province_id, culture_id, unit_type_id);
-		} else {
-			command::start_naval_unit_construction(state, state.local_player_nation, province_id, unit_type_id);
+		if(parent) {
+			Cyto::Any n_payload = dcon::nation_id{};
+			parent->impl_get(state, n_payload);
+			dcon::nation_id n = Cyto::any_cast<dcon::nation_id>(n_payload);
+			Cyto::Any u_payload = dcon::unit_type_id{};
+			parent->impl_get(state, u_payload);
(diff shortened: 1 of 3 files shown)


src/gui/topbar_subwindows/military_subwindows/gui_build_unit_large_window.hpp
```diff
@@ -227,318 +165,54 @@ class build_units_listbox : public listbox_element_base<buildable_units, buildab
...
-										}
+		if(parent) {
+			Cyto::Any payload = dcon::unit_type_id{};
+			parent->impl_get(state, payload);
+			dcon::unit_type_id utid = Cyto::any_cast<dcon::unit_type_id>(payload);
+			if(is_navy == false) {
+				for(auto po : state.world.nation_get_province_ownership_as_nation(state.local_player_nation)) {
+					auto p = po.get_province();
+					state.world.for_each_culture([&](dcon::culture_id c) {
+						if(command::can_start_land_unit_construction(state, state.local_player_nation, p, c, utid)) {
```
schombert, line 179: Does this lag the window? You might try filtering by cultures that are actually present (via looking in the demographics) ...

src/gui/topbar_subwindows/military_subwindows/gui_build_unit_large_window.hpp
```diff
@@ -682,71 +244,106 @@ class build_unit_large_window : public window_element_base {
...
+			auto ptr = make_element_by_type<unit_folder_button>(state, id);
+			auto value = std::stoi(std::string(name.substr(12)));
+			for(uint8_t i = 0; i < state.military_definitions.unit_base_definitions.size(); i++) {
+				if(state.military_definitions.unit_base_definitions[dcon::unit_type_id(i)].icon == value) {
+					ptr->unit_type = dcon::unit_type_id(i);
+					break;
+				}
+			}
+			switch(value) {
+			case 1:
```
schombert, line 257: This won't work, because which values correspond to land or naval units could be changed by mods

schombert: Admittedly, that was a problem with the previous version as well

src/gui/topbar_subwindows/military_subwindows/gui_build_unit_large_window.hpp
wxwisiasdf: Doesn't lag window, even on debug

schombert: Well, I guess its fine then. Wildly wasteful, but fine

PR #366, opened 2023-06-08
Reverted Commit and General Cleanup (More Details)

Breijz:
Reverted Commit that reintroduced tooltip Cross-contamination
Added GUI Mode Setting & Added accompanying parity checks
Moved Modifiers out of research tooltip and finished research tooltip
Made Currency be postfixed instead of prefixed
Updated alice.csv and alice.gui
Added Comments and Docs for tooltips
Updated .clang-format


Breijz:
dont merge, needs some other fixes rq

schombert:
I know that you don't like the changes, but I would prefer them not to be described as "inaccurate". How about "classic" vs. "enhanced"?

Breijz:
I dont think enhanced is a good word for it either? Kinda insinuates the changes made are pure QoL changes without issues, Something like "Work in Progress" ?

schombert:
Everything is a work in progress at the moment.

Breijz:
True, although its more apt unless a better term can be found to describe that mode, and thats the best i can think of at the moment

Breijz:
Actually, how about something like "PA" as the mode name? given the changes are specific to PA, AKA not in Vic2

PR #369, opened 2023-06-09
Reorg of files, Stats win for mil. window, and unit reorg/change leader window

Breijz:
Added Stats window to Military window
Added roughdrafts of tooltips for military window
Fixed parameter order 'layout_dest& contents, sys::state& state` into `sys::state& state, layout_dest& contents`
Removed last tooltips from gui_common_elements.hpp, only three remain.
Updated reformat.sh


src/gui/gui_common_elements.hpp
```diff
@@ -1232,32 +1100,42 @@ class nation_ships_text : public standard_nation_text {
...
 	}
 
-	public:
+public:
 	std::string get_text(sys::state& state, dcon::nation_id nation_id) noexcept override {
 		return std::to_string(get_ship_count(state, nation_id));
 	}
 };
 
 class nation_brigades_text : public standard_nation_text {
```
schombert, line 1109: you have a class here that claims to contain the number of brigades but instead counts the number of armies. I think this is likely to be confusing

Breijz: ill rename it to armies, and likewise for nation_ships_text into navies

src/gui/gui_common_elements.hpp
```diff
@@ -237,25 +227,6 @@ class province_liferating_progress_bar : public standard_province_progress_bar {
 	float get_progress(sys::state& state, dcon::province_id prov_id) noexcept override {
 		return state.world.province_get_life_rating(prov_id) / 100.f;
 	}
-
-	tooltip_behavior has_tooltip(sys::state& state) noexcept override {
-		return tooltip_behavior::variable_tooltip;
-	}
-
-	void update_tooltip(sys::state& state, int32_t x, int32_t y, text::columnar_layout& contents) noexcept override {
```
schombert, line 245: Without this tooltip, how do we determine what the numerical liferating of a province is?

Breijz: gui/gui_province_window.hpp:17
it was moved into its own class

schombert: Then what is this class used for? As far as I can remember, the liferating bar only appears in the province window

Breijz: its meant to cover both potential and current scenario, z.B in future if we have a need to show two life rating bars, or if a mod does so then its covered by this. Otherwise it just keeps consistency with the other classes.

src/gui/gui_common_elements.hpp
Breijz: also, removed nation_ships_text, it seems nation_navies_text deprecated it

PR #372, opened 2023-06-12
refresh unit icons

wxwisiasdf:


src/gui/gui_map_icons.hpp
```diff
@@ -109,12 +109,14 @@ class unit_icon_window : public window_element_base {
 	}
 
 	void on_update(sys::state& state) noexcept override {
-		bool has_attrition = false; // TODO: Attrition
-		attr_icon->set_visible(state, has_attrition);
 		bool has_any = false;
 		state.world.province_for_each_army_location_as_location(content, [&](dcon::army_location_id id) { has_any = true; });
 		state.world.province_for_each_navy_location_as_location(content, [&](dcon::navy_location_id id) { has_any = true; });
-		set_visible(state, has_any);
+		for(const auto& c : children)
```
schombert, line 115: This will work, but it could be more efficient to have a single function outside the individual little windows that loops over all the provinces and determines whether each of them needs to display anything, turning on or off the icon entirely. As things become more complicated (battle icons, docked ships are shown as docked, multiple icons for when there are many units...) that approach will probably be more maintainable

PR #373, opened 2023-06-13
Ui fixes in production, diplomacy, and military windows. Fixed event crash.

Masterchef118:


src/gui/topbar_subwindows/gui_diplomacy_window.hpp
```diff
@@ -501,11 +501,17 @@ class diplomacy_country_facts : public window_element_base {
 			}
 			{
 				std::string text{};
+				uint32_t size = 0;
 				for(auto const culture_id : fat_id.get_accepted_cultures()) {
 					auto culture = dcon::fatten(state.world, culture_id);
-					text += text::produce_simple_string(state, culture.get_name()) + ", ";
+					if(size < fat_id.get_accepted_cultures().size()-1) {
+						text += text::produce_simple_string(state, culture.get_name()) + ", ";
```
schombert, line 508: This is a common and annoying patterns when text processing. Two other alternatives I have used are: add the first element on its own, and then prepend ", " to every following element (so the loops starts from one). And: add ", " after every element and then remove the last two characters after the loop. Neither is much prettier. Maybe someone will introduce me to an elegant solution one of these days.

PR #375, opened 2023-06-15
National value fix

Ivysaur-Mrquestionmarks:


Breijz:
Dont add tooltips to the gui_common_elements.hpp file, comment at the top of the file should explain, but the issue with doing this is that it can cause tooltip-cross contamination, this is where two tooltips have the same text when they should either have different ones, or shouldnt have anything, etc, instead the better thing to do is to inherit from the class you wish to add a tooltip to and add the tooltip to the inherited version

Breijz:
disregard, i misread and was stupid, the tooltip code isnt in the gui_common_elements.hpp so thanks for not doing that, looks good otherwise :3
might remove the commented code though

PR #376, opened 2023-06-18
Unit Panel and Subwindow Work

Breijz:


src/entry_point_nix.cpp
```diff
@@ -13,7 +13,7 @@ int main() {
 	if(!sys::try_read_scenario_and_save_file(*game_state, NATIVE("development_test_file.bin"))) {
 		// scenario making functions
 		game_state->load_scenario_data();
-		game_state->local_player_nation = dcon::nation_id{42};
+		game_state->local_player_nation = dcon::nation_id{116};
```
schombert, line 16: Is this ENG?

src/gamestate/system_state.cpp
```diff
@@ -166,6 +171,34 @@ void state::render() { // called to render the frame may (and should) delay retu
 		}
 	}
 
+
+	if(!mouse_probe.under_mouse) {
+		if(ui_state.last_tooltip != ui_state.map_tooltip_window) {
+			ui_state.last_tooltip = ui_state.map_tooltip_window;
+		} 
+		if(ui_state.last_tooltip && ui_state.tooltip->is_visible()) {
```
schombert, line 179: So what this evaluates is whether a tooltip had been created previously, and if so, whether the ui element that created it is still visible. Note that this is within a black that will only run if nothing is under the mouse, so I am not sure what the intent is

src/gamestate/system_state.cpp
```diff
@@ -563,6 +596,10 @@ void state::on_create() {
 		new_elm->set_visible(*this, false);
 		ui_state.root->add_child_to_front(std::move(new_elm));
 	}
+	{
+		auto new_elm = ui::make_element_by_type<ui::map_tooltip>(*this, "tooltip");
```
schombert, line 600: We really should reuse the existing tooltip window, not create a second new window. Having two windows to display tooltips provides no advantage that I can see while opening the door to bugs that simply couldn't exist if there was just one window

src/gamestate/system_state.cpp
Breijz: copy and paste, need to fix ig

src/gui/gui_unit_panel.hpp
```diff
@@ -256,39 +331,59 @@ class unit_selection_panel : public window_element_base {
...
-		if constexpr(std::is_same_v<T, dcon::army_id>) {
-			unitlocation_text->set_text(state,
-					text::produce_simple_string(state, dcon::fatten(state.world, content).get_location_from_army_location().get_name()));
-		} else {
-			unitlocation_text->set_text(state,
-					text::produce_simple_string(state, dcon::fatten(state.world, content).get_location_from_navy_location().get_name()));
+		if(parent) {
+			Cyto::Any payload = T{};
+			parent->impl_get(state, payload);
+			auto content = any_cast<T>(payload);
```
schombert, line 337: For example, you could use receive<T>(state, parent) here

src/gamestate/system_state.cpp
Breijz: problem is that im not sure how to avoid it in this case, since a custom update_tooltip() is needed, i suppose i could just make the "map" a element, although that appears hacky

schombert: That is why we should collaborate on this more. The tooltip doesn't have an internal "update" at all. Rather the render function drives updating the tooltip's content

Breijz: how do you mean? each element has a update_tooltip() function and the render code calls said function no?

PR #378, opened 2023-06-20
Map Tooltips, and Coloniser flag fix

Breijz:


src/gamestate/system_state.hpp
```diff
@@ -47,9 +47,9 @@ struct user_settings_s {
 	float interface_volume = 1.0f;
 	bool prefer_fullscreen = false;
 	bool map_is_globe = false;
-	bool fake_graphs = false;
-	bool use_new_ui = true;
-	bool use_classic_fonts = false;
+	bool fake_graphs = true;
+	bool use_new_ui = false;
+	bool use_classic_fonts = true;
```
schombert, line 52: Why have you changed the defaults?

Breijz: Parity, People expect the UI to look like Vic2s, it also doesnt have any downside (anyone who wants to use the new stuff doesnt get effected by it past, at most, having to set their settings again), and given the new ui has some issues and the classic fonts are somewhat easier to read, it makes sense for those to be used

schombert: The classic fonts are much harder to read on higher resolutions. Putting that aside, if the defaults were do change, they should do so by general consensus, not unilaterally

Breijz: they were changed before without general consensus though? the use_new_ui was false by default in past, also as for the classic font, im using a 4k monitor for it, i cant say its harder to read on higher resolutions, the standard font is though given its so thin

PR #381, opened 2023-06-25
Naval combat window should mostly be done

Breijz:


assets/alice.csv
```diff
@@ -511,50 +511,50 @@ reform_effect_desc;When activated:;;;;;;;;;;;;x
 special_rules;Special rules:;;;;;;;;;;;;x
 voting_rules;Voting rules:;;;;;;;;;;;;x
 hide_decision;Hide the notification for this decision;;;;;;;;;;;;x
-sup_point_gain;Monthly suppression point increase: �G+$val$;;;;;;;;;;;;x
+sup_point_gain;Monthly suppression point increase: �G+$val$;;;;;;;;;;;;x
```
schombert, line 514: Lines like this are places where a utf 8 conversion has been performed that may look the same, but will probably break the text

Breijz: No way of fixing or avoiding it i think, ive only edited the csv file with codium, instead of vim, so not sure

src/text/text.cpp
```diff
@@ -473,6 +474,8 @@ variable_type variable_type_from_name(std::string_view v) {
 		CT_STRING_ENUM(theirnum)
 		CT_STRING_ENUM(totalemi)
 		CT_STRING_ENUM(totalimm)
+		CT_STRING_ENUM(infantry)
+		CT_STRING_ENUM(cavalry)
```
schombert, line 478: this is only seven letters long, not 8

.gitignore
```diff
@@ -12,7 +12,6 @@ mod[s]/
 
 # Generated parsers
 *_generated.hpp
-*/local_user_settings.hpp
```
schombert, line 15: local_user_settings is excluded on purpose, and I would like to keep it that way

Breijz: causes compile errors for the CI

schombert: Specifically, if it is in the repo, then changes anyone makes to set their personal game directory will be pushed to everyone. Likewise, pulling will overwrite your personal game directory

assets/alice.csv
schombert: Well, you should roll back to an earlier version of the file then. And you will need to find an editor that doesn't perform automatic conversion. I don't edit the file in Visual studio for that reason

.gitignore
Breijz: Ja, issue is that the CI needs the file for compiling, as its defines are used in the user_settings struct, it could probably be relocated, although not sure where

schombert: Suggest putting the defines into a different pr then until we can come up with a solution, because obviously we can't put them in local settings then

Breijz: aight, ill just move them into src/special_define.hpp and adjust the contributing.md then

schombert: It will lead to the same problem. If you expect people to edit these, their changes will constantly get sent to the repo. Is that what you want?

Breijz: No, thats why it was in local_user_settings.hpp, since people should change them, but these defines are needed for compile time (For CI), so it'll have to do

Breijz: OH wait nevermind, we can just exclude the .gitignore for local_user_settings for a single commit then add it back later, that'll make sure no ones copy gets pushed but the repo will maintain a valid version

schombert: So, when I edit them, it is going to change your file the next time you pull ...

Breijz: your stuff would still abide by the .gitignore, afterall its done at pushing time ?

Breijz: oh, though other option is to add a file creation as part of the github action, so it'll create a local_user_settings.hpp file with the defines needed

src/test
```diff
@@ -0,0 +1,3 @@
+test
```
schombert, line 1: ???

Breijz: no clue, removing it

PR #385, opened 2023-07-23
Ai revision

wxwisiasdf:
- better decision making
- code duplication removal
- small documentation effort
- allow OPMs to play :3

### DIFF
--- a/assets/alice.csv
+++ b/assets/alice.csv
@@ -906,6 +906,7 @@ ai_alliance_1;The AI is currently seeking allies
 ai_alliance_2;Your nation is close enough to assist in a war
 ai_alliance_3;No conflicting ambitions
 ai_alliance_4;Your nation's military is sufficiently strong
+ai_alliance_5;Has shared mutual strategic interests
 ai_access_1;The AI sees one of your wars as advantageous to them
 col_start_title;Begin colonizing this state
 col_start_1;No colonizers are present or colonization is still in the initial stage
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -62,11 +62,7 @@ void update_ai_general_status(sys::state& state) {
 			}
 		}
 
-		float self_str = float(state.world.nation_get_military_score(n));
-		for(auto subj : n.get_overlord_as_ruler()) {
-			self_str += 0.75f * float(subj.get_subject().get_military_score());
-		}
-
+		float self_str = estimate_strength(state, n);
 		float defensive_str = self_str;
 		defensive_str += in_sphere_of ? float(state.world.nation_get_military_score(in_sphere_of)) : 0.0f;
 		for(auto d : state.world.nation_get_diplomatic_relation(n)) {
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -5,9 +5,8 @@ namespace ai {
 
 static inline float estimate_strength(sys::state& state, dcon::nation_id n) {
 	float value = state.world.nation_get_military_score(n);
-	for(auto subj : state.world.nation_get_overlord_as_ruler(n)) {
+	for(auto subj : state.world.nation_get_overlord_as_ruler(n))
 		value += subj.get_subject().get_military_score();
-	}
 	return value;
 }
 
@@ -58,19 +57,12 @@ void update_ai_general_status(sys::state& state) {
 		for(auto b : state.world.nation_get_nation_adjacency_as_connected_nations(n)) {
 			auto other = b.get_connected_nations(0) != n ? b.get_connected_nations(0) : b.get_connected_nations(1);
 			if(!nations::are_allied(state, n, other) && (!in_sphere_of || in_sphere_of != other.get_in_sphere_of())) {
-				greatest_neighbor = std::max(greatest_neighbor, float(state.world.nation_get_military_score(other)));
+				greatest_neighbor = std::max(greatest_neighbor, estimate_strength(state, other));
 			}
 		}
 
 		float self_str = estimate_strength(state, n);
-		float defensive_str = self_str;
-		defensive_str += in_sphere_of ? float(state.world.nation_get_military_score(in_sphere_of)) : 0.0f;
-		for(auto d : state.world.nation_get_diplomatic_relation(n)) {
-			if(d.get_are_allied()) {
-				auto other = d.get_related_nations(0) != n ? d.get_related_nations(0) : d.get_related_nations(1);
-				defensive_str += float(state.world.nation_get_military_score(other));
-			}
-		}
+		float defensive_str = estimate_defensive_strength(state, n);
 
 		bool threatened = defensive_str < safety_factor * greatest_neighbor;
 		state.world.nation_set_ai_is_threatened(n, threatened);
@@ -82,8 +74,8 @@ void update_ai_general_status(sys::state& state) {
 				auto other = adj.get_connected_nations(0) != n ? adj.get_connected_nations(0) : adj.get_connected_nations(1);
 				auto ol = other.get_overlord_as_subject().get_ruler();
 				if(!ol && other.get_in_sphere_of() != n && (!threatened || !nations::are_allied(state, n, other))) {
-					auto other_str = float(other.get_military_score());
-					if(self_str * 0.5f < other_str && other_str <= self_str * 1.25) {
+					auto other_str = estimate_strength(state, other);
+					if(self_str * 0.5f < other_str && other_str <= self_str * 1.5f) {
 						auto rel = state.world.diplomatic_relation_get_value(state.world.get_diplomatic_relation_by_diplomatic_pair(n, other));
 						if(rel < min_relation) {
 							min_relation = rel;
@@ -100,7 +92,7 @@ void update_ai_general_status(sys::state& state) {
 				n.set_ai_rival(potential);
 			}
 		} else {
-			auto rival_str = float(n.get_ai_rival().get_military_score());
+			auto rival_str = estimate_strength(state, n.get_ai_rival());
 			auto ol = n.get_ai_rival().get_overlord_as_subject().get_ruler();
 			if(ol || n.get_ai_rival().get_in_sphere_of() == n || rival_str * 2 < self_str || self_str * 2 < rival_str) {
 				n.set_ai_rival(dcon::nation_id{});
@@ -113,28 +105,30 @@ void form_alliances(sys::state& state) {
 	static std::vector<dcon::nation_id> alliance_targets;
 
 	for(auto n : state.world.in_nation) {
-		if(n.get_is_player_controlled() == false && n.get_ai_is_threatened() && !(n.get_overlord_as_subject().get_ruler())) {
+		if(!n.get_is_player_controlled() && n.get_ai_is_threatened() && !(n.get_overlord_as_subject().get_ruler())) {
 			alliance_targets.clear();
 
-			for(auto nb : n.get_nation_adjacency()) {
-				auto other = nb.get_connected_nations(0) != n ? nb.get_connected_nations(0) : nb.get_connected_nations(1);
-				if(other.get_is_player_controlled() == false && !(other.get_overlord_as_subject().get_ruler())  && !nations::are_allied(state, n, other) && !military::are_at_war(state, other, n) && ai_will_accept_alliance(state, other, n)) {
+			for(auto other : state.world.in_nation)
+				if(other != n && !other.get_is_player_controlled() && !(other.get_overlord_as_subject().get_ruler()) && !nations::are_allied(state, n, other) && !military::are_at_war(state, other, n) && ai_will_accept_alliance(state, other, n))
 					alliance_targets.push_back(other.id);
-				}
-			}
 
 			if(!alliance_targets.empty()) {
 				std::sort(alliance_targets.begin(), alliance_targets.end(), [&](dcon::nation_id a, dcon::nation_id b) {
-					if(state.world.nation_get_military_score(a) != state.world.nation_get_military_score(b))
-						return state.world.nation_get_military_score(a) > state.world.nation_get_military_score(b);
+					if(estimate_strength(state, a) != estimate_strength(state, b))
+						return estimate_strength(state, a) > estimate_strength(state, b);
 					else
 						return a.index() > b.index();
 				});
-
 				nations::make_alliance(state, n, alliance_targets[0]);
 			}
 		}
 	}
+}
+
+static inline bool ai_is_close_enough(sys::state& state, dcon::nation_id target, dcon::nation_id from) {
+	auto target_continent = state.world.province_get_continent(state.world.nation_get_capital(target));
+	auto source_continent = state.world.province_get_continent(state.world.nation_get_capital(from));
+	return (target_continent == source_continent) || bool(state.world.get_nation_adjacency_by_nation_adjacency_pair(target, from));
 }
 
 bool ai_will_accept_alliance(sys::state& state, dcon::nation_id target, dcon::nation_id from) {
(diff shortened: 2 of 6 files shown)


src/ai/ai.cpp
```diff
@@ -22,23 +57,12 @@ void update_ai_general_status(sys::state& state) {
...
 			auto other = b.get_connected_nations(0) != n ? b.get_connected_nations(0) : b.get_connected_nations(1);
 			if(!nations::are_allied(state, n, other) && (!in_sphere_of || in_sphere_of != other.get_in_sphere_of())) {
-				greatest_neighbor = std::max(greatest_neighbor, float(state.world.nation_get_military_score(other)));
+				greatest_neighbor = std::max(greatest_neighbor, estimate_strength(state, other));
 			}
 		}
 
-		float self_str = float(state.world.nation_get_military_score(n));
-		for(auto subj : n.get_overlord_as_ruler()) {
-			self_str += 0.75f * float(subj.get_subject().get_military_score());
```
schombert, line 31: I want to maintain this discount when deciding how secure the nation is, because puppets are somewhat less useful on the defense

src/ai/ai.cpp
```diff
@@ -63,12 +87,12 @@ void update_ai_general_status(sys::state& state) {
 
 			if(potential) {
 				if(!n.get_is_player_controlled() && nations::are_allied(state, n, potential)) {
-					command::cancel_alliance(state, n, potential);
+					command::execute_cancel_alliance(state, n, potential);
```
schombert, line 90: good catch

src/ai/ai.cpp
```diff
@@ -81,69 +105,64 @@ void form_alliances(sys::state& state) {
...
 
 	for(auto n : state.world.in_nation) {
-		if(n.get_is_player_controlled() == false && n.get_ai_is_threatened() && !(n.get_overlord_as_subject().get_ruler())) {
+		if(!n.get_is_player_controlled() && n.get_ai_is_threatened() && !(n.get_overlord_as_subject().get_ruler())) {
 			alliance_targets.clear();
 
-			for(auto nb : n.get_nation_adjacency()) {
-				auto other = nb.get_connected_nations(0) != n ? nb.get_connected_nations(0) : nb.get_connected_nations(1);
-				if(other.get_is_player_controlled() == false && !(other.get_overlord_as_subject().get_ruler())  && !nations::are_allied(state, n, other) && !military::are_at_war(state, other, n) && ai_will_accept_alliance(state, other, n)) {
+			for(auto other : state.world.in_nation)
```
schombert, line 111: Checking only the adjacent nations was intentional -- distant allies are much less useful

src/ai/ai.cpp
```diff
@@ -1220,19 +1239,19 @@ crisis_str estimate_crisis_str(sys::state& state) {
...
-		dtotal += float(state.world.nation_get_military_score(secondary_defender));
+		dtotal += estimate_defensive_strength(state, secondary_defender);
 	}
 	for(auto& i : state.crisis_participants) {
 		if(!i.id)
 			break;
 		if(!i.merely_interested) {
 			if(i.supports_attacker) {
-				atotal += float(state.world.nation_get_military_score(i.id));
+				atotal += estimate_defensive_strength(state, i.id);
```
schombert, line 1252: This is wrong: it will add in strength of allies that won't be called in to a crisis war

src/ai/ai.cpp
```diff
@@ -1677,11 +1696,10 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
 			continue;
 		if((bits & (military::cb_flag::po_demand_state | military::cb_flag::po_annex)) == 0)
 			continue;
-		if(military::cb_infamy(state, c) > 13.0f)
+		if(state.world.nation_get_infamy(from) + military::cb_infamy(state, c) > state.defines.badboy_limit)
```
schombert, line 1699: 13 was picked intentionally to prevent the ai from spending all its infamy in one place

src/ai/ai.cpp
```diff
@@ -1692,27 +1710,26 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
 	}
 }
 
-bool valid_construction_target(sys::state& state, dcon::nation_id from, dcon::nation_id target) {
+static inline bool valid_construction_target(sys::state& state, dcon::nation_id from, dcon::nation_id target) {
```
schombert, line 1713: why have you made this function static?

src/ai/ai.cpp
```diff
@@ -1692,27 +1710,26 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
...
 		return false;
-	if(state.world.nation_get_military_score(target) > state.world.nation_get_military_score(from))
+	if(military::are_at_war(state, target, from))
 		return false;
-	if(state.world.nation_get_owned_province_count(target) <= 3)
+	
+	if(estimate_defensive_strength(state, target) * 0.5f > estimate_defensive_strength(state, from))
 		return false;
-	if(military::are_at_war(state, target, from))
+	// Allows small one-province-minors to declare war on other states, and bigger countries to ignore small one-province-minors
```
schombert, line 1729: You have missed the point of the line you changed. That line didn't prevent small nations from fabricating -- it prevented nations from wasting infamy by fabricating on them

src/ai/ai.cpp
```diff
@@ -1721,11 +1738,11 @@ void update_cb_fabrication(sys::state& state) {
 		if(!n.get_is_player_controlled() && n.get_owned_province_count() > 0) {
 			if(n.get_is_at_war())
 				continue;
-			if(n.get_infamy() >= 12.0f)
+			if(n.get_infamy() > state.defines.badboy_limit / 1.2f)
```
schombert, line 1741: This value works with the 13 above

src/ai/ai.cpp
wxwisiasdf: 25/2 = 12.5

src/ai/ai.cpp
wxwisiasdf: it allows small nations to fight each other

src/ai/ai.cpp
schombert: Yes. But with your change, if the nation has 0 infamy, it could spend up to 25 fabricating a single cb. I would rather have the nation not spend all its infamy in one place

src/ai/ai.cpp
schombert: Nations wasting full infamy amounts for 1 or 2 provinces should be avoided. The way things were let the ai of small nations fabricate on people, but prevented nations for wasting buckets of infamy on a tiny number of provinces. Note that they can still end up fighting if they have cbs from other sources; they just won't spend infamy

wxwisiasdf: No no no I wanted to make it so only SMALL nations fight other SMALL nations

schombert: No thanks

wxwisiasdf: Why not????? It's the most logical thing to do, for e.x Indian non-sphered nations

schombert: Because if a small nation has the opportunity, it should try to take a chunk off a larger neighbor.

src/ai/ai.cpp
```diff
@@ -3,6 +3,41 @@
 
 namespace ai {
 
+static inline float estimate_strength(sys::state& state, dcon::nation_id n) {
```
schombert, line 6: Again, why are these functions static? Or inline for that matter, since they are in a cpp file?

src/ai/ai.cpp
```diff
@@ -22,23 +57,14 @@ void update_ai_general_status(sys::state& state) {
...
-		float defensive_str = self_str;
-		defensive_str += in_sphere_of ? float(state.world.nation_get_military_score(in_sphere_of)) : 0.0f;
-		for(auto d : state.world.nation_get_diplomatic_relation(n)) {
-			if(d.get_are_allied()) {
-				auto other = d.get_related_nations(0) != n ? d.get_related_nations(0) : d.get_related_nations(1);
-				defensive_str += float(state.world.nation_get_military_score(other));
-			}
-		}
+		for(auto subj : n.get_overlord_as_ruler())
+			self_str += 0.5f * float(subj.get_subject().get_military_score());
```
schombert, line 66: why have you decided to make this 0.5 instead of 0.75?

wxwisiasdf: puppets dont help a lot anyways

src/ai/ai.cpp
```diff
@@ -81,69 +107,66 @@ void form_alliances(sys::state& state) {
...
-
-	auto target_continent = state.world.province_get_continent(state.world.nation_get_capital(target));
-	auto source_continent = state.world.province_get_continent(state.world.nation_get_capital(from));
-
-	bool close_enough = (target_continent == source_continent) || bool(state.world.get_nation_adjacency_by_nation_adjacency_pair(target, from));
-
-	if(!close_enough)
+	
+	// Same rival equates to instantaneous alliance (we benefit from more allies against a common enemy)
+	if(state.world.nation_get_ai_rival(target) == state.world.nation_get_ai_rival(from))
```
schombert, line 146: needs a check for whether the rival is set. Otherwise nations without rivals will all ally each other

src/ai/ai.cpp
```diff
@@ -81,69 +107,66 @@ void form_alliances(sys::state& state) {
...
-
-	auto target_score = state.world.nation_get_military_score(target);
-	auto source_score = state.world.nation_get_military_score(from);
-
-	return source_score * 2 >= target_score;
+	
+	// And also if they're powerful enough to be considered for an alliance
+	auto target_score = estimate_strength(state, target);
+	auto source_score = estimate_strength(state, from);
+	return source_score * 2.5f >= target_score;
```
schombert, line 156: Is there a reason you have increased the range here? Since we can't see how well the AI is performing yet, please don't fiddle with numbers like this when we can't see whether the change will be better or worse. Otherwise, the change is just a change for the sake of change

src/ai/ai.cpp
```diff
@@ -81,69 +107,66 @@ void form_alliances(sys::state& state) {
...
-
-	auto target_continent = state.world.province_get_continent(state.world.nation_get_capital(target));
-	auto source_continent = state.world.province_get_continent(state.world.nation_get_capital(state.local_player_nation));
-
-	bool close_enough = (target_continent == source_continent) || bool(state.world.get_nation_adjacency_by_nation_adjacency_pair(target, state.local_player_nation));
-
-	text::add_line_with_condition(state, contents, "ai_alliance_2", close_enough, indent);
+	text::add_line_with_condition(state, contents, "ai_alliance_2", ai_is_close_enough(state, target, state.local_player_nation), indent);
 
 	text::add_line_with_condition(state, contents, "ai_alliance_3", state.world.nation_get_ai_rival(target) != state.local_player_nation && state.world.nation_get_ai_rival(state.local_player_nation) != target, indent);
```
schombert, line 166: Since you added an additional condition for when the ai will accept an alliance, you need to add the condition in this section and create text explaining it.

src/ai/ai.cpp
```diff
@@ -1692,27 +1714,24 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
...
 	auto ol = state.world.nation_get_overlord_as_subject(from);
 	if(state.world.overlord_get_ruler(ol) && state.world.overlord_get_ruler(ol) != target)
 		return false;
-	auto sl = state.world.nation_get_in_sphere_of(target);
-	if(sl == from)
-		return false;
-	if(nations::are_allied(state, target, from))
+	if(state.world.nation_get_in_sphere_of(target) == from)
 		return false;
-	if(nations::are_allied(state, sl, from))
```
schombert, line 1707: We still need this condition; otherwise you will end up with ai fabricating cbs it shouldn't use

src/ai/ai.cpp
```diff
@@ -1692,27 +1714,24 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
...
+	if(military::are_at_war(state, target, from))
 		return false;
-	if(state.world.nation_get_military_score(target) > state.world.nation_get_military_score(from))
+	
+	if(estimate_defensive_strength(state, target) * 0.5f > estimate_defensive_strength(state, from))
```
schombert, line 1731: Comparing my defensive strength to their defensive strength is not a good measure for when I might be able to use a cb -- I can't call my defensive allies in to attack necessarily

src/ai/ai.cpp
```diff
@@ -1692,27 +1714,24 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
...
 		return false;
 	if(state.world.nation_get_owned_province_count(target) <= 3)
 		return false;
-	if(military::are_at_war(state, target, from))
```
schombert, line 1713: We need this condition because you can't fabricate on someone you are currently at war with

src/ai/ai.cpp
schombert: Ok, but see below: since we don't know how things are working yet, changing constants like this is change for the sake of change

assets/alice.csv
```diff
@@ -873,7 +873,7 @@ msg_decision_1;$x$ has taken the $y$ decision
 msg_decision_2;Effect:
 msg_rebels_win_title;Rebel victory
 msg_rebels_win_1;$x$ rebels have won.
-msg_rebels_win_2;�R$x$�! prestige lost
+msg_rebels_win_2;�R$x$�! prestige lost
```
schombert, line 876: It looks like the encoding on this and similar lines was messed up

assets/alice.csv
```diff
@@ -906,14 +906,15 @@ ai_alliance_1;The AI is currently seeking allies
 ai_alliance_2;Your nation is close enough to assist in a war
 ai_alliance_3;No conflicting ambitions
 ai_alliance_4;Your nation's military is sufficiently strong
+ai_alliance_5;Has mutual rivals
```
schombert, line 909: Since "rival" is purely an internal name, let's go with something more like "Has shared strategic aims". Also, since this condition works as an OR (as either it or some of the other conditions have be be satisfied) more wording may need to be adjusted

src/ai/ai.cpp
```diff
@@ -81,69 +107,68 @@ void form_alliances(sys::state& state) {
...
-
-	auto target_continent = state.world.province_get_continent(state.world.nation_get_capital(target));
-	auto source_continent = state.world.province_get_continent(state.world.nation_get_capital(from));
-
-	bool close_enough = (target_continent == source_continent) || bool(state.world.get_nation_adjacency_by_nation_adjacency_pair(target, from));
-
-	if(!close_enough)
+	
+	// Same rival equates to instantaneous alliance (we benefit from more allies against a common enemy)
+	if(state.world.nation_get_ai_rival(target) && state.world.nation_get_ai_rival(from) && state.world.nation_get_ai_rival(target) == state.world.nation_get_ai_rival(from))
```
schombert, line 146: You only need to check that one of the two is valid. Since if one is valid and == is satisfied, the other must be valid as well

src/ai/ai.cpp
schombert: Still missing this condition

src/ai/ai.cpp
```diff
@@ -1693,26 +1719,23 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
...
-
+	if(state.world.nation_get_constructing_cb_type(from))
+		return false;
 	auto ol = state.world.nation_get_overlord_as_subject(from);
 	if(state.world.overlord_get_ruler(ol) && state.world.overlord_get_ruler(ol) != target)
 		return false;
-	auto sl = state.world.nation_get_in_sphere_of(target);
-	if(sl == from)
-		return false;
-	if(nations::are_allied(state, target, from))
```
schombert, line 1705: And now this condition appears to be missing

PR #387, opened 2023-07-24
fixes for test_project

wxwisiasdf:


src/gui/gui_element_base.hpp
```diff
@@ -118,13 +118,11 @@ class element_base {
 
 template<typename T>
 inline T retrieve(sys::state& state, element_base* parent) {
+	Cyto::Any payload = T{};
```
schombert, line 121: I don't understand why you made this change. It will still function the same, of course, but it is just slightly slower in the case where parent is null

src/gui/topbar_subwindows/gui_production_window.hpp
```diff
@@ -198,10 +191,10 @@ class factory_upgrade_button : public shift_button_element_base {
 	}
 
 	void update_tooltip(sys::state& state, int32_t x, int32_t y, text::columnar_layout& contents) noexcept override {
-		auto fid = retrieve<dcon::factory_id>(state, parent);
+		const dcon::factory_id fid = retrieve<dcon::factory_id>(state, parent);
```
schombert, line 194: You can write `const auto`

src/gui/topbar_subwindows/gui_production_window.hpp
```diff
@@ -333,67 +320,53 @@ class factory_subsidise_button : public button_element_base {
...
-				// TODO: classic tooltips ?
+		const dcon::factory_id fid = retrieve<dcon::factory_id>(state, parent);
+		const dcon::nation_id n = retrieve<dcon::nation_id>(state, parent);
+		if(n == state.local_player_nation) {
+			text::add_line(state, contents, "factory_delete_header");
+			if(disabled) {
+				text::add_line_break_to_layout(state, contents);
+				text::add_line(state, contents, "factory_delete_not_allowed");
 			}
+			// TODO: classic tooltips ?
```
schombert, line 369: I guess these comments can be removed, since the special setting for them is no more

PR #388, opened 2023-07-31
-fix topbar bug with politics window

wxwisiasdf:
-fix ai violating constraints when joinin' war
-fix war continuining even after combatant is ded

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -559,7 +559,7 @@ A war goal added in this way will result in a status quo war goal on that side b
 
 Any war ending probably requires checking that units aren't stuck in invalid combats, updating blockade status, etc.
 
-A nation being fully annexxed (i.e no longer exists) will automatically white peace all wars it was involved in and remove itself from war participations, it will also end all wars it was the primary attacker or defender.
+A nation being fully annexxed (i.e no longer exists) will automatically remove itself from all wars it was involved in, it will also end all wars it was the primary attacker or defender.
 
 #### Crisis offers
 
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -559,6 +559,8 @@ A war goal added in this way will result in a status quo war goal on that side b
 
 Any war ending probably requires checking that units aren't stuck in invalid combats, updating blockade status, etc.
 
+A nation being fully annexxed (i.e no longer exists) will automatically white peace all wars it was involved in and remove itself from war participations, it will also end all wars it was the primary attacker or defender.
+
 #### Crisis offers
 
 Crisis resolution offers function much in the same way as peace offers. Every refused crisis offer increases the temperature of the current crisis by define:CRISIS_TEMPERATURE_ON_OFFER_DECLINE.
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -1800,12 +1800,19 @@ void update_cb_fabrication(sys::state& state) {
 }
 
 bool will_join_war(sys::state& state, dcon::nation_id n, dcon::war_id w, bool as_attacker) {
+	// Prevent constraint violations
+	if(!military::standard_war_joining_is_possible(state, w, n, as_attacker))
+		return false;
+
 	if(!as_attacker)
 		return true;
 	for(auto par : state.world.war_get_war_participant(w)) {
 		if(par.get_is_attacker() == false) {
 			if(military::can_use_cb_against(state, n, par.get_nation()))
 				return true;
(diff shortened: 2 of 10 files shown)


src/ai/ai.cpp
```diff
@@ -1800,12 +1800,19 @@ void update_cb_fabrication(sys::state& state) {
 }
 
 bool will_join_war(sys::state& state, dcon::nation_id n, dcon::war_id w, bool as_attacker) {
+	// Prevent constraint violations
+	if(!military::standard_war_joining_is_possible(state, w, n, as_attacker))
```
schombert, line 1804: this check should be made in the diplomatic action acceptance, rather than here, otherwise players may be able to join invalid wars

src/ai/ai.cpp
```diff
@@ -1800,12 +1800,19 @@ void update_cb_fabrication(sys::state& state) {
...
+		return false;
+
 	if(!as_attacker)
 		return true;
 	for(auto par : state.world.war_get_war_participant(w)) {
 		if(par.get_is_attacker() == false) {
 			if(military::can_use_cb_against(state, n, par.get_nation()))
 				return true;
+		// Eager to absolutely demolish our rival uwu
+		} else if(state.world.nation_get_ai_rival(n) == par.get_nation()) {
```
schombert, line 1814: this will cause nations to join alongside their rivals, not against them

src/military/military.cpp
```diff
@@ -2900,6 +2900,11 @@ void implement_war_goal(sys::state& state, dcon::war_id war, dcon::cb_type_id wa
 	float prestige_gain = successful_cb_prestige(state, wargoal, from) * (war ? 1.0f : state.defines.crisis_wargoal_prestige_mult);
 	nations::adjust_prestige(state, from, prestige_gain);
 	nations::adjust_prestige(state, target, -prestige_gain);
+
+	// If the nation stops existing we shall cleanup all wars it was involved at
+	if(state.world.nation_get_owned_province_count(target) == 0)
```
schombert, line 2905: This check shouldn't be here; it should be done when a nation loses a province (provinces can also be lost via event or rebel defection, for example)

src/provinces/province.cpp
```diff
@@ -859,6 +859,7 @@ void change_province_owner(sys::state& state, dcon::province_id id, dcon::nation
 
 void conquer_province(sys::state& state, dcon::province_id id, dcon::nation_id new_owner) {
 	bool was_colonial = state.world.province_get_is_colonial(id);
+	auto old_owner = state.world.province_get_nation_from_province_ownership(id);
```
schombert, line 862: you forgot to remove this

docs/rules.md
```diff
@@ -559,6 +559,8 @@ A war goal added in this way will result in a status quo war goal on that side b
 
 Any war ending probably requires checking that units aren't stuck in invalid combats, updating blockade status, etc.
 
+A nation being fully annexxed (i.e no longer exists) will automatically white peace all wars it was involved in and remove itself from war participations, it will also end all wars it was the primary attacker or defender.
```
schombert, line 562: "automatically white peace all wars it was involved in" sounds like it would end the wars. But what you mean presumably is that it will end its participation in the wars. Fortunately, this is automatically handled by the fact that deleting the nation also destroys all of its relationships, including war participation

src/nations/nations.cpp
```diff
@@ -1245,6 +1245,19 @@ void cleanup_nation(sys::state& state, dcon::nation_id n) {
 		state.world.delete_wargoal(wg);
 	}
 
+	// If the nation stops existing we shall cleanup all wars it was involved at
+	// War participation may already be annulled by previous calls so we have to iterate
```
schombert, line 1249: This shouldn't happen. If the nation or the war hasn't been deleted, then everything should remain in place. Instead what you want to do is something like this:
```
par = ... the range or war participation ...
while(par.begin() != par.end()) {
	... do something that will result in the deletion of (*par.begin()) ...
}
```
However, there is a big problem here
Look at how a peace offer is implemented. We go over each wargoal in the offer one by one, implementing them. Suppose that annexing the war leader happens to be first on the list. Your code here would then, in implementing that wargoal, delete the war itself, thereby deleting all the wargoals attached to it, thereby preventing any other wargoals from being implemented, even if they were directed against other people. It also has the problem that the peace offer routine is going to continue and read out information from this now deleted war to possibly try to end it on its own. We could end up trying to delete the war twice, essentially

PR #396, opened 2023-08-04
AI work

wxwisiasdf:


src/military/military.cpp
```diff
@@ -3959,6 +3964,23 @@ void cleanup_army(sys::state& state, dcon::army_id n) {
...
 		state.world.delete_regiment((*regs.begin()).get_regiment());
 	}
+
+	auto controller = state.world.army_get_controller_from_army_control(n);
+	auto b = state.world.army_get_battle_from_army_battle_participation(n);
+	if(b) {
+		bool should_end = true;
+		// TODO: Do they have to be in common war or can they just be "hostile against"?
+		for(auto bp : state.world.land_battle_get_army_battle_participation_as_battle(b)) {
+			if(bp.get_army() != n && are_allied_in_war(state, controller, state.world.army_get_controller_from_army_control(bp.get_army()))) {
```
schombert, line 3974: This will break for battles against rebels

src/provinces/province.cpp
```diff
@@ -1567,7 +1573,7 @@ std::vector<dcon::province_id> make_land_path(sys::state& state, dcon::province_
 			if((bits & province::border::impassible_bit) == 0 && !origins_vector.get(other_prov)) {
 				if(other_prov == end) {
 					fill_path_result(nearest.province);
-					return path_result;
+					return assert_path_result(path_result);
```
schombert, line 1576: I believe that this will break NRVO

src/provinces/province.cpp
```diff
@@ -1593,7 +1599,7 @@ std::vector<dcon::province_id> make_land_path(sys::state& state, dcon::province_
 		}
 	}
 
-	return path_result;
+	return assert_path_result(path_result);
```
schombert, line 1602: as above

src/ai/ai.cpp
```diff
@@ -19,7 +19,7 @@ float estimate_defensive_strength(sys::state& state, dcon::nation_id n) {
 
 		auto other = dr.get_related_nations(0) != n ? dr.get_related_nations(0) : dr.get_related_nations(1);
 		if(other.get_overlord_as_subject().get_ruler() != n)
-			value += estimate_strength(state, other);
+			value += other.get_military_score();
```
schombert, line 22: Why are we not calling estimate strength here?

src/ai/ai.cpp
```diff
@@ -104,17 +104,32 @@ void update_ai_general_status(sys::state& state) {
...
 			alliance_targets.clear();
 
-			for(auto nb : n.get_nation_adjacency()) {
-				auto other = nb.get_connected_nations(0) != n ? nb.get_connected_nations(0) : nb.get_connected_nations(1);
-				if(other.get_is_player_controlled() == false && !(other.get_overlord_as_subject().get_ruler())  && !nations::are_allied(state, n, other) && !military::are_at_war(state, other, n) && ai_will_accept_alliance(state, other, n))
-					alliance_targets.push_back(other.id);
+			// Adjacency with us
+			get_alliance_targets_by_adjacency(state, n.id, n.id, alliance_targets);
+			// Adjacency with rival (useful for e.x, Chile allying Paraguay to fight bolivia)
+			get_alliance_targets_by_adjacency(state, n.id, n.get_ai_rival(), alliance_targets);
```
schombert, line 125: Running this twice could end up with duplicate nations in the list

src/ai/ai.cpp
wxwisiasdf: the estimate_strength function also addup the overlords themselves, so when an overlord != n then it will duplicate by the subjects

src/ai/ai.cpp
wxwisiasdf: handled below with std::unique

src/ai/ai.cpp
```diff
@@ -104,17 +104,32 @@ void update_ai_general_status(sys::state& state) {
...
+			// Adjacency with people who are at war with us
+			for(auto wp : state.world.nation_get_war_participant(n)) {
+				bool as_attacker = wp.get_is_attacker();
+				for(auto p : state.world.war_get_war_participant(wp.get_war())) {
+					if(p.get_is_attacker() == !as_attacker)
+						get_alliance_targets_by_adjacency(state, n.id, p.get_nation(), alliance_targets);
```
schombert, line 131: And this is not a good idea, because alliances are more permanent for the AI while wars are temporary. And the AI shouldn't be forming alliances just to get dragged into a war in any case. Furthermore, just because you are in a war, doesn't mean that you can call allies in anyways (not war leader, or not defending), so most of these alliances will be useless in any case.

src/provinces/province.cpp
wxwisiasdf: it shalln't in release, in debug it may, but that's the point - to assert all paths are being checked correctly

src/ai/ai.cpp
schombert: Also, the rival is not always set. You need to check whether the rival exists in the first place

src/ai/ai.cpp
```diff
@@ -124,7 +139,15 @@ void form_alliances(sys::state& state) {
 					else
 						return a.index() > b.index();
 				});
+				auto last = std::unique(alliance_targets.begin(), alliance_targets.end());
+				alliance_targets.erase(last, alliance_targets.end());
```
schombert, line 143: More efficient to not add duplicates

src/ai/ai.cpp
```diff
@@ -124,7 +139,15 @@ void form_alliances(sys::state& state) {
...
 				nations::make_alliance(state, n, alliance_targets[0]);
+
+				// Call our new allies into wars.... they may not accept but they may just may join!
+				for(auto wp : state.world.nation_get_war_participant(n)) {
+					if(!military::are_allied_in_war(state, n, alliance_targets[0]))
+						command::execute_call_to_arms(state, n, alliance_targets[0], wp.get_war());
```
schombert, line 149: And don't do this; at best this is a lot of unnecessary work since most of these CTAs won't be valid

src/ai/ai.cpp
wxwisiasdf: the AI should take advantage of the situation and not be "exactly" bound to some alliance, we already do ignore diplomatic relations to an extent so why don't we just let the AI don't care about their allies in the first place? this type of behaviour allows the AI to just take advantage instead of being taken advantage of

src/ai/ai.cpp
```diff
@@ -147,14 +170,27 @@ bool ai_will_accept_alliance(sys::state& state, dcon::nation_id target, dcon::na
 	if(state.world.nation_get_ai_rival(target) && state.world.nation_get_ai_rival(target) == state.world.nation_get_ai_rival(from))
 		return true;
 	
+	// Having a mutual, non-rival enemy - incentivizing AI to clump against a
+	// common enemy more easily
+	bool mutual_enemy = false;
+	state.world.for_each_nation([&](dcon::nation_id n) {
+		if(n != target && n != from) {
+			bool enemy_a = military::can_use_cb_against(state, from, n) || military::are_at_war(state, from, n);
```
schombert, line 178: can_use_cb_against is *expensive*. And this is an n^2 check. So you must remove this

src/ai/ai.cpp
```diff
@@ -163,7 +199,17 @@ void explain_ai_alliance_reasons(sys::state& state, dcon::nation_id target, text
 
 	text::add_line(state, contents, "kierkegaard_1", indent);
 
-	text::add_line_with_condition(state, contents, "ai_alliance_5", state.world.nation_get_ai_rival(target) && state.world.nation_get_ai_rival(target) == state.world.nation_get_ai_rival(state.local_player_nation), indent + 15);
+	// Having a mutual, non-rival enemy - incentivizing AI to clump against a
+	// common enemy more easily
+	bool mutual_enemy = false;
+	state.world.for_each_nation([&](dcon::nation_id n) {
+		if(n != target && n != state.local_player_nation) {
+			bool enemy_a = military::can_use_cb_against(state, state.local_player_nation, n) || military::are_at_war(state, state.local_player_nation, n);
```
schombert, line 207: see above

src/ai/ai.cpp
```diff
@@ -1801,13 +1847,15 @@ void update_cb_fabrication(sys::state& state) {
 }
 
 bool will_join_war(sys::state& state, dcon::nation_id n, dcon::war_id w, bool as_attacker) {
-	if(!as_attacker)
-		return true;
 	for(auto par : state.world.war_get_war_participant(w)) {
```
schombert, line 1850: Put the removed condition back. It ensures that the AI will answer defensive CTA. Without this guarantee, alliances are meaningless

src/ai/ai.cpp
```diff
@@ -1801,13 +1847,15 @@ void update_cb_fabrication(sys::state& state) {
...
-		if(par.get_is_attacker() == false) {
+		if(par.get_is_attacker() == !as_attacker) {
+			// Could use a CB against this nation?
 			if(military::can_use_cb_against(state, n, par.get_nation()))
 				return true;
-			// Eager to absolutely demolish our rival
+			// Already at war with this nation?
+			if(military::are_at_war(state, n, par.get_nation()))
```
schombert, line 1856: If you are already at war with X, war joining conditions prevent you from joining a second war against (or with X)

src/provinces/province.cpp
schombert: It will break in release too. NRVO requires the return statement to use the *name* (NRVO = named return value optimization). This is a case where you would need to use a macro. In fact, a macro with the loop wrapped in NDEBUG just in case

src/ai/ai.cpp
schombert: Just trust me on this one for now

src/ai/ai.cpp
wxwisiasdf: I'm troubled as to why we are allowing alliances to be so meaningful, even players break alliances all the time - bounding AI to be a faithful ally is detrimental to itself, it will help in wars it doesn't care about and really would any player do the same?

src/ai/ai.cpp
schombert: you are still letting duplicates be added

src/ai/ai.cpp
```diff
@@ -136,15 +159,30 @@ bool ai_is_close_enough(sys::state& state, dcon::nation_id target, dcon::nation_
...
 }
 
+static bool ai_has_mutual_enemy(sys::state& state, dcon::nation_id from, dcon::nation_id target) {
+	auto rival_a = state.world.nation_get_ai_rival(target);
+	auto rival_b = state.world.nation_get_ai_rival(from);
+	// Same rival equates to instantaneous alliance (we benefit from more allies against a common enemy)
+	if(rival_a && rival_a == rival_b)
+		return true;
+	// Our rivals are allied?
+	if(rival_a && rival_b && rival_a != rival_b && nations::are_allied(state, rival_a, rival_b))
```
schombert, line 169: this will result in way too many alliances

src/ai/ai.cpp
```diff
@@ -154,7 +192,7 @@ bool ai_will_accept_alliance(sys::state& state, dcon::nation_id target, dcon::na
 	// And also if they're powerful enough to be considered for an alliance
 	auto target_score = estimate_strength(state, target);
 	auto source_score = estimate_strength(state, from);
-	return source_score * 2.0f >= target_score;
+	return std::max<float>(source_score, 1.f) * ally_overestimate >= target_score;
```
schombert, line 195: why have you increased this to 3.5?

src/military/military.cpp
```diff
@@ -3959,6 +3964,33 @@ void cleanup_army(sys::state& state, dcon::army_id n) {
...
+		bool should_end = true;
+		auto controller = state.world.army_get_controller_from_army_control(n);
+		if(bool(controller)) {
+			// TODO: Do they have to be in common war or can they just be "hostile against"?
+			for(auto bp : state.world.land_battle_get_army_battle_participation_as_battle(b)) {
+				if(bp.get_army() != n && are_allied_in_war(state, controller, state.world.army_get_controller_from_army_control(bp.get_army()))) {
+					should_end = false;
+				}
+			}
+		} else {
```
schombert, line 3979: You are still not handling the case where A and B are together in a battle fighting rebels. If you remove A, B should still fight the rebels even if there is no alliance between B and A

PR #399, opened 2023-08-06
Multiplayer implementation

wxwisiasdf:


schombert:
I'm probably not going to merge this in until the harder work has been done. I suggest you keep it on a separate branch for now

src/gamestate/serialization.cpp
```diff
@@ -829,8 +829,13 @@ bool try_read_scenario_and_save_file(sys::state& state, native_string_view name)
 				[&](uint8_t const* ptr_in, uint32_t length) { read_scenario_section(ptr_in, ptr_in + length, state); });
 		buffer_pos = with_decompressed_section(buffer_pos,
 				[&](uint8_t const* ptr_in, uint32_t length) { read_save_section(ptr_in, ptr_in + length, state); });
-
-		state.game_seed = uint32_t(std::random_device()());
+		
+		if(state.network_state.is_present()) {
```
schombert, line 833: This is not a good idea. The way it should work is that the seed should be generated as part of making the "save" that represents the initial game state. That save could then be sent out to all players

PR #401, opened 2023-08-06
Pop Growth tooltip / ui thing.

NicholasLN:


src/nations/nations.cpp
```diff
@@ -28,7 +28,38 @@ int64_t get_monthly_pop_increase_of_nation(sys::state& state, dcon::nation_id n)
...
+	}
+
+	for(auto& province_id : province_list) {
+		auto fat_id = dcon::fatten(state.world, province_id);
+		fat_id.for_each_pop_location_as_province([&](dcon::pop_location_id id) {
+			auto pop = state.world.pop_location_get_pop(id);
+
+			auto growth = int64_t(demographics::get_monthly_pop_increase(state, pop));
+			auto promote = -int64_t(demographics::get_estimated_type_change(state, pop));
+			auto assimilation = -int64_t(demographics::get_estimated_assimilation(state, pop));
```
schombert, line 51: I don't think we need to subtract assimilation, type change, or the internal migrations, because while that pop will decrease in number, some other pop in the nation will increase by the same amount, cancelling it out.

PR #409, opened 2023-08-09
fix mods crashing due to purpousefully leaving out some dds files

wxwisiasdf:


src/map/map.cpp
```diff
@@ -666,6 +666,8 @@ void display_data::set_unit_arrows(std::vector<std::vector<glm::vec2>> const& ar
 
 GLuint load_dds_texture(simple_fs::directory const& dir, native_string_view file_name) {
 	auto file = simple_fs::open_file(dir, file_name);
+	if(!bool(file))
```
schombert, line 669: Are you sure this won't result in the game repeatedly trying to open the file? Doing that operation, even if it fails, can be slow

wxwisiasdf:
context: some mods purpousefully leave out DDS files, for example on the rivers and such, which causes them to crash due to forcefully loading them, so we simply set the texture id to an invalid handle if not found, which will automatically be used by the opengl driver and of course, display nothing - because no such texture would be present

src/map/map.cpp
wxwisiasdf: sorry, my bad, time to use peek file then

src/map/map.cpp
```diff
@@ -665,9 +665,9 @@ void display_data::set_unit_arrows(std::vector<std::vector<glm::vec2>> const& ar
 }
 
 GLuint load_dds_texture(simple_fs::directory const& dir, native_string_view file_name) {
+	if(!bool(simple_fs::peek_file(dir, file_name)))
```
schombert, line 668: Even peeking at the file can be time consuming if you are doing it every frame. You need to make sure that on the return from where load_dds_texture is called, the texture is set as already loaded, even if the loading "failed" and we have the null handle. (and then you can go back to just using load without peek)

schombert: That way we won't get stuck in a loop of calling into the file system every frame looking for the same missing texture over and over again

wxwisiasdf: They are only ever called when loading

wxwisiasdf: When initially building the map

schombert: Uh, that sounds impossible, since the map is built as part of the scenario, and obviously those textures need to be loaded when rendering

PR #413, opened 2023-08-09
placeholder trigger and effects

wxwisiasdf:
why:

some mods have hundreds and hundreds of invalid and malformed effects and triggers, this causes a **crash** in Project Alice, rather than crashing we would do:

a) emit warning and error messages
b) let the user continue if so they wish

I think a PR like this was submitted very early on, but it didn't have placeholder triggers/effects, here it does have them and it allows most invalid triggers and effects to simply call a "dummy" no-operation trigger.

it evaluates to false, so it never gets executed and doesn't cause a cascade of invalid errors and event chains, etc.

this is mainly for improving the user experience of "it works out of the box" - yes, it means some content will not be accessible due to it having errors - but it will also allow users to experience the content without having to wait for the mod developers to fix the mod for alice

also some mods are no longer updated, as such we should probably allow for potentially erroneous triggers and effects to be "dummy"-ified, so we are able to run popular, unupdated mods

### DIFF
--- a/src/gamestate/system_state.cpp
+++ b/src/gamestate/system_state.cpp
@@ -934,10 +934,16 @@ std::string_view state::to_string_view(dcon::unit_name_id tag) const {
 }
 
 dcon::trigger_key state::commit_trigger_data(std::vector<uint16_t> data) {
-	if(data.size() == 0) {
+	if(data.empty()) {
 		if(trigger_data_indices.empty())
 			trigger_data_indices.push_back(int32_t(0));
 		return dcon::trigger_key();
+	}
+	if(trigger_data_indices.empty()) { // Create placeholder for invalid triggers
+		trigger_data_indices.push_back(0);
+		trigger_data.push_back(uint16_t(trigger::integer_scope));
+		trigger_data.push_back(uint16_t(2));
+		trigger_data.push_back(uint16_t(1));
 	}
 
 	auto search_result = std::search(trigger_data.data(), trigger_data.data() + trigger_data.size(),
@@ -965,8 +971,14 @@ dcon::trigger_key state::commit_trigger_data(std::vector<uint16_t> data) {
 }
 
 dcon::effect_key state::commit_effect_data(std::vector<uint16_t> data) {
-	if(data.size() == 0)
-		return dcon::effect_key();
+	if(data.empty())
+ 		return dcon::effect_key();
+	
+	if(effect_data.empty()) { // Create placeholder for invalid effects
+		effect_data.push_back(uint16_t(trigger::integer_scope));
+		effect_data.push_back(uint16_t(2));
+		effect_data.push_back(uint16_t(1));
+	}
 
 	auto start = effect_data.size();
 	auto size = data.size();
--- a/src/gamestate/system_state.cpp
+++ b/src/gamestate/system_state.cpp
@@ -943,7 +943,7 @@ dcon::trigger_key state::commit_trigger_data(std::vector<uint16_t> data) {
 		trigger_data_indices.push_back(0);
 		trigger_data.push_back(uint16_t(trigger::integer_scope));
 		trigger_data.push_back(uint16_t(2));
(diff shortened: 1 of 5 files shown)


src/gamestate/system_state.cpp
```diff
@@ -934,11 +934,17 @@ std::string_view state::to_string_view(dcon::unit_name_id tag) const {
...
 dcon::trigger_key state::commit_trigger_data(std::vector<uint16_t> data) {
-	if(data.size() == 0) {
+	if(data.empty()) {
 		if(trigger_data_indices.empty())
 			trigger_data_indices.push_back(int32_t(0));
 		return dcon::trigger_key();
 	}
+	if(trigger_data_indices.empty()) { // Create placeholder for invalid triggers
+		trigger_data_indices.push_back(0);
+		trigger_data.push_back(uint16_t(trigger::integer_scope));
```
schombert, line 944: Why are you using an integer scope, which requires an extra byte?

src/parsing/effect_parsing.cpp
```diff
@@ -1102,9 +1102,12 @@ dcon::effect_key make_effect(token_generator& gen, error_handler& err, effect_bu
 		return dcon::effect_key{0};
 	}
 
-	auto const new_size = simplify_effect(context.compiled_effect.data());
-	context.compiled_effect.resize(static_cast<size_t>(new_size));
-
+	if(err.accumulated_errors.empty()) {
```
schombert, line 1105: This will stop making triggers after any error anywhere in the mod, no?

src/gamestate/system_state.cpp
wxwisiasdf: A dummy trigger and effect would:

a) run on any scope (province, nation, state or pop)
b) evaluate to false, always

I found the integer_scope to fullfill those two criteria

schombert:
I am not currently interested in this PR, because we should be improving mod compatibility before release.

src/gamestate/system_state.cpp
schombert: Then just use always = no. That would be a single uint16_t

wxwisiasdf:
That is true, but this allows developers to see the full errors and warnings when loading - so we are able to fix stuff as-is without having to pull out a debugger - otherwise we crash and print nothing

schombert:
That's fine, but then why are you skipping things once any error has been detected?

wxwisiasdf:
I also want to mention that erroneous triggers causes funny behaviours on various functions, doing this dummy trigger/effect thing allows us to have a mostly "correct state" even after we've received objectively bad input

Basically instead of crashing due to half-made bytecode, we will find out the real reasons why we crash

wxwisiasdf:
Ah, thats just an oversight on my part, not the intended behaviour

wxwisiasdf:
Oh yeah also, probably something to note:
When a trigger or an effect is of 0 size, it will also be handed over to the dummy trigger, whereas before it would run whatever happened to be the first trigger on the array, I think this could've been a bug we would've had to deal with if we ever had such predicament

src/parsing/effect_parsing.cpp
```diff
@@ -1094,17 +1094,23 @@ void recurse_over_effects(uint16_t* source, T const& f) {
...
 	if(context.compiled_effect.size() >= std::numeric_limits<uint16_t>::max()) {
 		err.accumulated_errors += "effect is " + std::to_string(context.compiled_effect.size()) +
 															" cells big, which exceeds 64 KB bytecode limit (" + err.file_name + ")";
 		return dcon::effect_key{0};
 	}
 
-	auto const new_size = simplify_effect(context.compiled_effect.data());
-	context.compiled_effect.resize(static_cast<size_t>(new_size));
-
+	if(e_err.accumulated_errors.empty()) {
```
schombert, line 1108: just take the size before and compare it with the size after, instead of creating a whole new error object

PR #417, opened 2023-08-10
allow rebels being specified in provinces without owners instead of c…

wxwisiasdf:
…rashing - emit an error

### DIFF
--- a/src/economy/demographics.cpp
+++ b/src/economy/demographics.cpp
@@ -2035,11 +2035,9 @@ dcon::pop_id find_or_make_pop(sys::state& state, dcon::province_id loc, dcon::cu
 		float totals = 0.0f;
 		state.world.for_each_ideology([&](dcon::ideology_id i) {
 			if(state.world.ideology_get_enabled(i)) {
-				auto const i_key = pop_demographics::to_key(state, i);
 				auto ptrigger = state.world.pop_type_get_ideology(ptid, i);
+				auto const i_key = pop_demographics::to_key(state, i);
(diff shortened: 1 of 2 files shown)


src/economy/demographics.cpp
```diff
@@ -2035,11 +2035,9 @@ dcon::pop_id find_or_make_pop(sys::state& state, dcon::province_id loc, dcon::cu
 		float totals = 0.0f;
 		state.world.for_each_ideology([&](dcon::ideology_id i) {
 			if(state.world.ideology_get_enabled(i)) {
-				auto const i_key = pop_demographics::to_key(state, i);
 				auto ptrigger = state.world.pop_type_get_ideology(ptid, i);
+				auto const i_key = pop_demographics::to_key(state, i);
 				auto owner = nations::owner_of_pop(state, np);
-
-				assert(ptrigger);
```
schombert, line 2042: Why have you taken this out? An ideology without an ideological drift trigger feels like a bug to me. And in any case, it is going to be evaluated below ... without a check for its validity, it would need an additional condition to skip the invalid case.

wxwisiasdf: dummy triggers are already evaluated to 0, so the effect is annulled

schombert: ok, but make sure that you double check that. The index for the invalid trigger is -1. If that is used to index into things ... bad things will happen

wxwisiasdf: Yes, checked that now - in the separate PR that behaviour is now accounted for - so it will not make bad things happen :D

PR #418, opened 2023-08-10
Expand effects to be beyond 64K limit

wxwisiasdf:
Fixes errors with various "big" mods like GFM and TGC. The issue is that most effects are indeed valid, however they are put way beyond the 64K barrier the effects normally have, so it "calls" whatever happens to be the key%64K, which often results in crashes. Basically, the effects are OK, they just need more "key-space".

### DIFF
--- a/src/gamestate/serialization.cpp
+++ b/src/gamestate/serialization.cpp
@@ -234,6 +234,7 @@ uint8_t const* read_scenario_section(uint8_t const* ptr_in, uint8_t const* secti
 	ptr_in = deserialize(ptr_in, state.trigger_data);
 	ptr_in = deserialize(ptr_in, state.trigger_data_indices);
 	ptr_in = deserialize(ptr_in, state.effect_data);
+	ptr_in = deserialize(ptr_in, state.effect_data_indices);
 	ptr_in = deserialize(ptr_in, state.value_modifier_segments);
 	ptr_in = deserialize(ptr_in, state.value_modifiers);
 	ptr_in = deserialize(ptr_in, state.text_data);
@@ -413,6 +414,7 @@ uint8_t* write_scenario_section(uint8_t* ptr_in, sys::state& state) {
 	ptr_in = serialize(ptr_in, state.trigger_data);
 	ptr_in = serialize(ptr_in, state.trigger_data_indices);
 	ptr_in = serialize(ptr_in, state.effect_data);
+	ptr_in = serialize(ptr_in, state.effect_data_indices);
 	ptr_in = serialize(ptr_in, state.value_modifier_segments);
 	ptr_in = serialize(ptr_in, state.value_modifiers);
 	ptr_in = serialize(ptr_in, state.text_data);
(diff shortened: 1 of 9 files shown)


src/gamestate/serialization.cpp
```diff
@@ -234,6 +234,7 @@ uint8_t const* read_scenario_section(uint8_t const* ptr_in, uint8_t const* secti
 	ptr_in = deserialize(ptr_in, state.trigger_data);
 	ptr_in = deserialize(ptr_in, state.trigger_data_indices);
 	ptr_in = deserialize(ptr_in, state.effect_data);
+	ptr_in = deserialize(ptr_in, state.effect_data_indices);
```
schombert, line 237: you need to bump up the version number

src/gui/gui_effect_tooltips.cpp
```diff
@@ -6243,7 +6243,7 @@ uint32_t internal_make_effect_description(EFFECT_DISPLAY_PARAMS) {
 
 void effect_description(sys::state& state, text::layout_base& layout, dcon::effect_key k, int32_t primary_slot, int32_t this_slot,
 		int32_t from_slot, uint32_t r_lo, uint32_t r_hi) {
-	effect_tooltip::internal_make_effect_description(state, state.effect_data.data() + k.index(), layout, primary_slot, this_slot,
+	effect_tooltip::internal_make_effect_description(state, state.effect_data.data() + state.effect_data_indices[k.index()], layout, primary_slot, this_slot,
```
schombert, line 6246: For example, if this is called with the invalid handle, k.index() will equal -1, which is obviously not a good idea

src/gamestate/system_state.cpp
```diff
@@ -986,19 +986,39 @@ dcon::trigger_key state::commit_trigger_data(std::vector<uint16_t> data) {
...
-	auto size = data.size();
+	if(effect_data_indices.empty()) { // Create placeholder for invalid effects
+		effect_data_indices.push_back(0);
+		effect_data.push_back(uint16_t(effect::no_payload));
+	}
 
-	effect_data.resize(start + size, uint16_t(0));
-	std::copy_n(data.data(), size, effect_data.data() + start);
-	return dcon::effect_key(uint16_t(start));
+	auto search_result = std::search(effect_data.data(), effect_data.data() + effect_data.size(),
```
schombert, line 1000: Any data on how often this actually finds something? While for triggers, there are lots of duplicate conditions (like all the triggers that check for a specific tag) I wasn't expecting there to be a lot of duplicate events, and looking through all the effect data for each effect can be slow

wxwisiasdf: for TGC it's around 32K, and effects are somewhere around a hundred or so

schombert: You mean that there are a hundred duplicate effects? If so, that may not be worth the search time

wxwisiasdf: Pardon me, on vanilla there are 260 duplicate effects, on TGC there are around 600

PR #427, opened 2023-08-11
generic solution for adding extra buildings

wxwisiasdf:
since naval bases, forts and railroads are hardcoded on the game's logic, the natural solution was to simply keep them, and any additional buildings would be added with more ease.

this PR does NOT add any new buildings, it rather makes it easier to add them, however - and it also simplifies a bunch of logic for the game - removing code duplication and whatnot.


schombert:
I'm not saying I won't pull this in, but this is a large number of changes to review, so it will probably take a while at best.

src/gamestate/dcon_generated.txt
```diff
@@ -2733,18 +2720,8 @@ object {
...
-	property {
-		name{ increase_fort }
-		type{ bitfield }
-		tag{ scenario }
-	}
-	property {
-		name{ increase_naval_base }
-		type{ bitfield }
+		name{ increase_building }
+		type{ array{economy::province_building_type}{uint8_t} }
```
schombert, line 2724: this should be an array of bits, not of uint8_t s

src/scripting/effects.cpp
```diff
@@ -2514,13 +2514,13 @@ uint32_t ef_rgo_size(EFFECT_PARAMTERS) {
 	return 0;
 }
 uint32_t ef_fort(EFFECT_PARAMTERS) {
-	auto& fort_level = ws.world.province_get_fort_level(trigger::to_prov(primary_slot));
-	fort_level = uint8_t(std::clamp(int32_t(fort_level) + int32_t(trigger::payload(tval[2]).signed_value), 0, 255));
```
schombert, line 2518: good catch on the 2

schombert:
I also suggest bumping the scenario file version as a courtesy to the other contributors

PR #434, opened 2023-08-12
basic error reporting and logging

wxwisiasdf:
"standarizes" error reporting so that we can report errors appropriatedly, it will also set breakpoints on error messages and fatal errors shall exit (relatively) gracefully from now on


schombert:
This is not desirable for two reasons. First, in the long run errors from building a scenario need to show up in the launcher. Secondly, writing to the console simply won't work on windows. The console isn't open while the game is running, so all error messages sent to it will be lost. That is why we use a message box

PR #437, opened 2023-08-12
flags render on text with @TAG

wxwisiasdf:


src/ogl/opengl_wrapper.cpp
```diff
@@ -456,12 +456,60 @@ void render_character(sys::state const& state, char codepoint, color_modificatio
...
+					tag[0] = char(codepoints[i++]);
+					if(i + 1 < count) {
+						tag[1] = char(codepoints[i++]);
+						if(i + 1 < count) {
+							tag[2] = char(codepoints[i++]);
+						}
+					}
+				}
+				i--;
+				GLuint flag_texture_handle = get_flag_texture_handle_from_tag(state, tag);
```
schombert, line 498: You probably want to do the following: are there more than 3 characters after the `@` and are those 3 a valid tag? If so, render the flag. Otherwise, render the text as text. (and then mirror this into the metrics) This allows misuses of the `@` to show up in text so that they are findable

PR #438, opened 2023-08-13
join/leave commands

wxwisiasdf:


schombert:
Both of these commands are probably unnecessary. Leave: just closing the game will close the connection, which the server can detect. Join: connecting probably should be done in the launcher. This is because only two games running the same scenario can connect to a single game. Thus connecting to a server should be done prior to / along with picking a scenario, so that the client can respond meaningfully to the scenario not matching (this cannot be done in game because changing the scenario would require unloading the ui). Thus joining probably has to be handled in the launcher.

PR #441, opened 2023-08-13
save editor write uncompressed savefiles

wxwisiasdf:


schombert:
that's going to make it harder for people to share and compare them. Wouldn't it be easier to write a comparison routine that does the decompression and then compares the result?

PR #446, opened 2023-08-14
lobby networking additions

wxwisiasdf:
- notification when a player enters and leaves (sets and unsets the "is_player_controlled" on the specified nation)
- disable selecting countries already selected by other players (QoL)
- notification for the ping of a player, this is specially useful for hosts and players to know if it's the fault of the game, or their internet just fucking sucks (QoL)
- i'll probably add the classic "ping bar" triangle icon to show the level of ping to the user in an intuitive way


src/gui/gui_nation_picker.hpp
```diff
@@ -200,10 +200,24 @@ class saves_window : public window_element_base {
...
+
 	void button_action(sys::state& state) noexcept override {
 		auto n = retrieve<dcon::nation_id>(state, parent);
-		state.local_player_nation = n;
-		state.ui_state.nation_picker->impl_on_update(state);
+		if(state.network_mode == sys::network_mode::single_player) {
+			state.local_player_nation = n;
+			state.ui_state.nation_picker->impl_on_update(state);
+		} else {
+			command::switch_nation(state, state.local_player_nation, state.world.nation_get_identity_from_identity_holder(n));
```
schombert, line 219: Have you accounted for the fact that the player don't have a valid nation id when the game opens? This command will already read as being sent from REB or whatever, and the normal version of this command IIRC sets the player controlled flag, which is not appropriate yet

src/common_types/constants.hpp
```diff
@@ -295,10 +295,16 @@ constexpr inline uint8_t standard_pause = log | popup | pause;
 
 enum class game_mode {
 	pick_nation,
-	single_player,
+	in_game,
```
schombert, line 298: The reason to keep single player here is that the game loop won't work the same way. In single player, the game loop advances the days automatically based on time. In multiplayer it must do so only when it receives a command to advance the day. So these are not differences only in networking

src/gamestate/commands.cpp
```diff
@@ -4061,6 +4081,30 @@ void execute_notify_player_leaves(sys::state& state, dcon::nation_id source) {
...
+	memset(&p, 0, sizeof(payload));
+	p.type = command_type::notify_player_ping;
+	p.source = source;
+	p.data.ping.ms = ms;
+	add_to_command_queue(state, p);
+}
+bool can_notify_player_ping(sys::state& state, dcon::nation_id source, uint32_t ms) {
+	return true;
+}
+void execute_notify_player_ping(sys::state& state, dcon::nation_id source, uint32_t ms) {
```
schombert, line 4095: What are you trying to accomplish with this command?

PR #447, opened 2023-08-14
Multiplayer PR

wxwisiasdf:


src/gamestate/system_state.cpp
```diff
@@ -2915,23 +2915,32 @@ void state::single_game_tick() {
...
-					single_game_tick();
+					if(network_mode == sys::network_mode::host) {
+						command::payload c;
+						c.type = command::command_type::advance_tick;
+						outgoing_commands.push(c);
+					} else {
+						single_game_tick();
+					}
 				} else {
-					command::execute_pending_commands(*this);
```
schombert, line 2934: why has this gone away?

wxwisiasdf: it went above

src/network/network.cpp
```diff
@@ -0,0 +1,74 @@
...
+void network_state::send_and_receive_commands(sys::state& state) {
+	if(state.network_mode == sys::network_mode::host) {
+		// put server commands onto the queues of the worker commands
+		auto* c = outgoing_commands.front();
+		while(c) {
+			// propagate to all the worker's queues
+			for(uint32_t i = 0; i < max_clients; i++)
+				if(clients[i].active)
+					clients[i].worker_commands.push(*c);
+			state.incoming_commands.push(*c); // and our local command queue
```
schombert, line 47: this could result in a deadlock. If the incoming commands queue is full, this will wait. And if it is waiting, there is nothing running that could take them out of the queue. It also isn't clear how any of this will eventually work, since nothing ever gets taken out of worker commands. Are you going to block on each network call until it completes or are you going to use asynchronous io?

src/gamestate/system_state.cpp
schombert: but above, it will only be executed if the tick is ready to advance, meaning that mid tick commands would have to wait, no?

wxwisiasdf: no, it will execute regardless of whenever it's mid tick or when it's supposed to go in a tick

src/network/network.cpp
wxwisiasdf: asnchronous of course, threads will pop elements from the worker queues and the server, indeed, will block if there is a lot of commands - but that is practically unachievable unless you have 100's of players

wxwisiasdf: the server will not wait for the worker queues to empty, it will continue to push, the worker threads will pop at their own rate and this ensures that all commands that are about to be executed, are sent first into the clients and then put into the incoming commands queue

schombert: You don't have to accept that possibility at all.Server can just receive command from client -> valid? -> execute and broadcast. It doesn't need to store it at all

schombert: This is also where you would filter out console commands from clients

wxwisiasdf: broadcasting is blocking

schombert: I thought that you were using async io?

wxwisiasdf: i was using raw sockets, should i've used asio??

schombert: You don't have to block on a socket send. You can use a non-blocking send. Or, if your OS layer doesn't support that, dispatch the sending to a thread pool or something else that can block

wxwisiasdf: that is what i was doing, i was putting a thread for each client so it would block there with the workerqueue for the send/recv parts of the sockets - in fact i added a wrapper for non_blocking send and recv

schombert: Also, you can create a non-blocking socket itself with
u_long mode = 1;  // 1 to enable non-blocking socket
ioctlsocket(sock, FIONBIO, &mode);

wxwisiasdf: that is what i was doing in fact lemme push the recv and send i was using for such

schombert: Again, the exact implementation isn't important for the moment. My point is simply that by doing what I described above you don't need to then store the incoming commands, and thus you don't have to risk deadlocking

src/network/network.cpp
```diff
@@ -64,9 +64,10 @@ void send_and_receive_commands(sys::state& state) {
...
 			if(command::is_console_command(c->type)) {
 				// propagate to all the worker's queues
-				for(uint32_t i = 0; i < max_clients; i++)
-					if(state.network_state.clients[i].active)
-						state.network_state.clients[i].worker_commands.push(*c);
+				for(auto& client : state.network_state.clients)
+					if(client.active) {
+						// TODO: send commands to client
+					}
 				state.incoming_commands.push(*c); // and our local command queue
```
schombert, line 71: you need to execute the command here, not store it. A command may affect the validity of the next command

PR #454, opened 2023-08-15
support `decision = <decision-name>` on country history files

wxwisiasdf:
used mainly for "setup" of countries (tgc,hpm,etc)

### DIFF
--- a/src/gamestate/system_state.cpp
+++ b/src/gamestate/system_state.cpp
@@ -29,6 +29,7 @@
 #include <thread>
 #include "rebels.hpp"
 #include "ai.hpp"
+#include "effects.hpp"
 #include "gui_leader_select.hpp"
 #include "gui_land_combat.hpp"
 #include "gui_nation_picker.hpp"
@@ -2327,7 +2328,7 @@ void state::load_scenario_data() {
 						it != context.map_of_ident_names.end()) {
 					auto holder = context.state.world.national_identity_get_nation_from_identity_holder(it->second);
 
-					parsers::country_history_context new_context{context, it->second, holder};
+					parsers::country_history_context new_context{context, it->second, holder, pending_decisions};
 
 					auto opened_file = open_file(country_file);
 					if(opened_file) {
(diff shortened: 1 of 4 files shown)


src/parsing/parsers_declarations.cpp
```diff
@@ -2679,6 +2679,25 @@ void country_history_file::ruling_party(association_type, std::string_view value
...
+		}
+		return dcon::text_sequence_id();
+	}();
+
+	dcon::decision_id d;
+	context.outer_context.state.world.for_each_decision([&](dcon::decision_id d) {
+		auto name = context.outer_context.state.world.decision_get_name(d);
+		if(name == value_key) {
+			if(auto e = context.outer_context.state.world.decision_get_effect(d); e)
+				effect::execute(context.outer_context.state, e, trigger::to_generic(context.holder_id), trigger::to_generic(context.holder_id), 0, uint32_t(context.outer_context.state.current_date.value), uint32_t(context.holder_id.index() << 4 ^ d.index()));
```
schombert, line 2696: Running the decisions at this point is unlikely to give great results since the scenario isn't finished being generated yet. You would need to add them to some list and then run them all after generation has finished

src/gamestate/system_state.cpp
```diff
@@ -2334,6 +2335,9 @@ void state::load_scenario_data() {
 						auto content = view_contents(*opened_file);
 						parsers::token_generator gen(content.data, content.data + content.file_size);
 						parsers::parse_country_history_file(gen, err, new_context);
+						
+						for(auto& e : new_context.pending_decisions)
```
schombert, line 2339: this loop could be done faster with just `pending_decisions = std::move(new_context.pending_decisions));` which will transfer the internal buffer of one vector to the other, no copies or new allocations required

schombert: Oh, I see, you have to do this for each file. In that case, you could put a reference to the outer vector in the context for some marginal savings

PR #462, opened 2023-08-20
Debug program to create crash dumps and whatnot

wxwisiasdf:
steps to make dump:

- run alice
- alice will output a message box on windows when abort() is called, this stops the application
- user runs DbgAlice
- crash dump generated


src/gamestate/system_state.cpp
```diff
@@ -2621,8 +2621,16 @@ void state::load_scenario_data() {
 
 	military::recover_org(*this);
 
-	if(err.accumulated_errors.length() > 0)
-		window::emit_error_message(err.accumulated_errors, err.fatal);
+	if(!err.accumulated_errors.empty()) {
+		if(err.fatal) {
```
schombert, line 2625: I really don't want this. In just two days I will be writing the launcher, and I will want it to be able to report errors in its own way

src/map/map.cpp
```diff
@@ -476,7 +474,20 @@ void display_data::render(glm::vec2 screen_size, glm::vec2 offset, float zoom, m
 				count.push_back(border.count);
 			}
 		}
-
+		glMultiDrawArrays(GL_TRIANGLES, &first[0], &count[0], GLsizei(count.size()));
+	}
+	if(zoom <= 5) { // Coastal borders
```
schombert, line 479: this will draw more than coasts. It will draw all the inland impassible borders as well

src/ogl/opengl_wrapper.cpp
```diff
@@ -4,6 +4,7 @@
 
 namespace ogl {
 
+#if 0
```
schombert, line 7: don't do this. Use /* ... */ if you want to comment out more than one line

PR #470, opened 2023-08-21
Added the pop size and deployed/possible regiments text to buildable units in build unit window. Added some other related things as well.

Masterchef118:


src/gui/topbar_subwindows/military_subwindows/gui_build_unit_large_window.hpp
```diff
@@ -269,6 +276,24 @@ class units_build_item : public listbox_row_element_base<buildable_unit_entry_in
 		} else if(name == "province") {
 			auto ptr = make_element_by_type<generic_name_text<dcon::province_id>>(state, id);
 			return ptr;
+		} else if(name == "time_to_build") {
+			auto ptr = make_element_by_type<simple_text_element_base>(state, id);
+			build_time = ptr.get();
+			return ptr;
+		} else if(name == "popsize") {
+			if(pass == 0) {
```
schombert, line 284: This is very funky. Don't do this. Instead of the recursive call to make child, move the contents of else block below to where the recursive call is (renaming the second ptr local variable to something distinct), and get rid of the pass variable. And then everything will work fine. You don't need to return the pointer you create in the second block because the only thing returning the pointer does is add it as a child element (if it is non-null, otherwise it creates a default placeholder). By calling add_child in the else block, you have already done all the necessary work.

PR #472, opened 2023-08-21
better liberal AI to reduce militancy sprouts

wxwisiasdf:
>tend to conservative IFF we need to jumpstart economy
>otherwise become liberal to be able to enact policies
>this highly appeases jacobin rebels
>less revolts :)

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -677,37 +677,37 @@ void update_ai_econ_construction(sys::state& state) {
 		int32_t max_projects = std::max(2, int32_t(treasury / 8000.0f));
 
 		auto rules = n.get_combined_issue_rules();
+
+		bool can_appoint = [&]() {
+			if(!politics::can_appoint_ruling_party(state, n))
+				return false;
+			auto last_change = state.world.nation_get_ruling_party_last_appointed(n);
+			if(last_change && state.current_date < last_change + 365)
+				return false;
+			if(politics::is_election_ongoing(state, n))
+				return false;
+			return true;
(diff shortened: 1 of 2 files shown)


src/ai/ai.cpp
```diff
@@ -4359,6 +4366,13 @@ void update_land_constructions(sys::state& state) {
 			}
 		}
 
+		// mobilize if we feel threatened
+		if(state.world.nation_get_ai_is_threatened(n)) {
+			military::start_mobilization(state, n);
```
schombert, line 4371: No. Mobilization really, really hurts your economy. Mobilizing when you don't immediately need the troops thus leaves you weaker and weaker over time. It also makes your pops more and more angry. *And* there are negative events that fire if you are mobilized while not at war.

src/ai/ai.cpp
```diff
@@ -70,6 +70,13 @@ void update_ai_general_status(sys::state& state) {
 		bool threatened = defensive_str < safety_factor * greatest_neighbor;
 		state.world.nation_set_ai_is_threatened(n, threatened);
 
+		// account for domestic rebel threats
+		auto militancy = state.world.nation_get_demographics(n, demographics::militancy);
+		auto total_pop = state.world.nation_get_demographics(n, demographics::total);
+		if(total_pop / militancy >= state.defines.mil_to_join_rebel * 0.75f) {
+			state.world.nation_set_ai_is_threatened(n, true);
```
schombert, line 77: Being threatened is just for making alliances. And allies won't fight your rebels. So this is counterproductive

PR #476, opened 2023-08-25
ai party fixes for liberals

wxwisiasdf:


src/ai/ai.cpp
```diff
@@ -678,65 +678,86 @@ void update_ai_econ_construction(sys::state& state) {
...
+						for(auto pi : state.culture_definitions.party_issues) {
+							auto issue_rules = state.world.political_party_get_party_issues(pid, pi).get_rules();
+							if((issue_rules & issue_rule::build_factory) != 0) {
+								target = pid;
+								break;
 							}
 						}
 					}
+				}
+			} else {
```
schombert, line 727: Why do this? Appointing any party will make some people mad, right. And there seems to be no limit to how often you are doing this other than possibility, so even if it makes a minority of people angry, you will be making some people angrier and angrier for no reason. I suggest using nation_get_ruling_party_last_appointed to make this condition happen less frequently. Once every 5 or 10 years, perhaps

src/ai/ai.cpp
```diff
@@ -1179,14 +1200,24 @@ void take_reforms(sys::state& state) {
 			continue;
 
 		if(n.get_is_civilized()) { // political & social
-			float max_support = 0.0f;
+			// Enact social policies to deter Jacobin rebels from overruning the country
+			// Reactionaries will popup in effect but they are MORE weak that Jacobins
```
schombert, line 1204: The reason this looked at movements before is that, if you don't pick the issue most supported by an ongoing movement, everyone in every other movement will get angry. I am willing to try your new version instead, but keep that in mind. You may eventually need to take both into account

src/ai/ai.cpp
wxwisiasdf: it is worth the cost for the return on investment, the ai sacrifices short term militancy for long term militancy decrease

wxwisiasdf: fine changing this

schombert: I said that you *could* do it, just less often than once a year. You didn't have to take it out completely

PR #485, opened 2023-08-27
tooltip settings

yrrrmmbl:


src/gui/gui_main_menu.cpp
```diff
@@ -113,6 +113,24 @@ void autosave_display::on_update(sys::state& state) noexcept {
...
+}
+void tooltip_mode_right::button_action(sys::state& state) noexcept {
+	state.user_settings.bind_tooltip_mouse = !state.user_settings.bind_tooltip_mouse;
+}
+void tooltip_mode_right::on_update(sys::state& state) noexcept {
+	send(state, parent, notify_setting_update{});
+}
+void tooltip_mode_display::on_update(sys::state& state) noexcept {
+	if (state.user_settings.bind_tooltip_mouse)set_text(state, "Stick to Mouse");
+	else set_text(state, "Center over UI Element");
```
schombert, line 132: don't forget to make these load from the csv file later

src/gui/gui_main_menu.cpp
```diff
@@ -113,6 +113,24 @@ void autosave_display::on_update(sys::state& state) noexcept {
 			break;
 	}
 }
+
+
+void tooltip_mode_left::button_action(sys::state& state) noexcept {
+	state.user_settings.bind_tooltip_mouse = !state.user_settings.bind_tooltip_mouse;
+}
+void tooltip_mode_left::on_update(sys::state& state) noexcept {
+	send(state, parent, notify_setting_update{});
```
schombert, line 122: this needs to go into the button action, not on update

schombert: You send the message up after changing something when the button is clicked to notify the main window that it needs to save and update. If you do it in on update, you will create a loop where, upadate -> notify parent to update -> parent updates again -> etc

yrrrmmbl:
TODO: add slider to change tooltip delay

schombert:
still need to pull the text to be displayed from the csv file.

yrrrmmbl:
i just make a new line or is this also somehow hardcoded?

schombert:
Yep, just go into alice.csv and add new lines as required. I suggest immediately prior to the last line, but it doesn't actually matter

PR #487, opened 2023-08-28
appoint RP according to most popular ones with the people

wxwisiasdf:


wxwisiasdf:
prioritized having a popular rp over jumpstarting the economy, since LF is already handled by the overly-aggressive capitalists

src/ai/ai.cpp
```diff
@@ -667,114 +667,84 @@ void take_ai_decisions(sys::state& state) {
...
+	return true;
 }
 
 void update_ai_econ_construction(sys::state& state) {
 	for(auto n : state.world.in_nation) {
 		// skip over: non ais, dead nations, and nations that aren't making money
-		if(n.get_is_player_controlled() || n.get_owned_province_count() == 0 || n.get_spending_level() < 1.0f || n.get_last_treasury() >= n.get_stockpiles(economy::money))
+		if(n.get_is_player_controlled() || n.get_owned_province_count() == 0)
 			continue;
+		//if(n.get_spending_level() < 1.0f || n.get_last_treasury() >= n.get_stockpiles(economy::money))
```
schombert, line 706: If you make them build more when they already can't afford it, that will just slow down construction further, making it take even longer for them to finish and yield profits

src/ai/ai.cpp
```diff
@@ -667,114 +667,84 @@ void take_ai_decisions(sys::state& state) {
...
+		//	continue;
+
+		// buy stuff from the global market if we need it
+		state.world.for_each_commodity([&](dcon::commodity_id c) {
+			n.set_stockpile_targets(c, 10000.f);
```
schombert, line 711: This will result in a lot of purchases that aren't necessary and it won't work as you expect. If you try to fill the stockpile from empty, it is almost certain all that demand won't be satisfied. So this will result in stockpiles flipping on and off AND starving factories and consumers of goods

wxwisiasdf: so i just reverse it? when demand is satisfied i buy it up, and when it is not, i sell it?

schombert:
And when you say that you have checked that the capitalists are enough, are you sure? The capitalists make much worse decisions than the nation's ai

wxwisiasdf:
> And when you say that you have checked that the capitalists are enough, are you sure? The capitalists make much worse decisions than the nation's ai

yeah they seem to do well into the 1860s

PR #490, opened 2023-08-28
Added a bool to turn on/off movement of the map when certain gui elements are visible.

Masterchef118:


src/gamestate/system_state.cpp
```diff
@@ -281,7 +281,7 @@ void state::on_mouse_wheel(int32_t x, int32_t y, key_modifiers mod, float amount
 		}
 	}
 }
-void state::on_key_down(virtual_key keycode, key_modifiers mod) {
+void state::on_key_down(virtual_key keycode, key_modifiers mod, bool can_move_map_while_visible) {
```
schombert, line 284: I don't see why this needs to be a parameter if it is already a member variable of the state object

src/gamestate/system_state.cpp
```diff
@@ -290,7 +290,7 @@ void state::on_key_down(virtual_key keycode, key_modifiers mod) {
 				ui::show_main_menu(*this);
 			}
 
-			map_state.on_key_down(keycode, mod);
+			map_state.on_key_down(keycode, mod, can_move_map_while_visible);
```
schombert, line 293: similarly, instead of passing it as a parameter here, you could test the variable and just not call the function at all when you don't want keypresses to be sent to the map

PR #493, opened 2023-09-01
ai that goes to 1910 without rebels

wxwisiasdf:
somehow doing this makes the AI not succumb to rebels, until the 1910s....

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -6,8 +6,14 @@ namespace ai {
 
 float estimate_strength(sys::state& state, dcon::nation_id n) {
 	float value = state.world.nation_get_military_score(n);
-	for(auto subj : state.world.nation_get_overlord_as_ruler(n))
-		value += subj.get_subject().get_military_score();
+	if(!state.world.nation_get_is_civilized(n)) // 75% weaker
+		value *= 0.25f;
+	for(auto subj : state.world.nation_get_overlord_as_ruler(n)) {
+		float subj_value = subj.get_subject().get_military_score();
+		if(!subj.get_subject().get_is_civilized()) // 75% weaker
+			subj_value *= 0.25f;
+		value += subj_value;
+	}
 	return value;
 }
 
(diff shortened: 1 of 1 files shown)


src/ai/ai.cpp
```diff
@@ -70,6 +76,11 @@ void update_ai_general_status(sys::state& state) {
 		bool threatened = defensive_str < safety_factor * greatest_neighbor;
 		state.world.nation_set_ai_is_threatened(n, threatened);
 
+		// provinces owned by rebels -> immediate threat
+		for(const auto po : n.get_province_ownership_as_nation())
+			if(bool(po.get_province().get_province_rebel_control().get_rebel_faction()))
+				state.world.nation_set_ai_is_threatened(n, true);
```
schombert, line 82: This will make the ai accept alliances when occupied by a rebel somewhere, but those allies will not help it out

src/ai/ai.cpp
```diff
@@ -101,6 +112,12 @@ void update_ai_general_status(sys::state& state) {
 				n.set_ai_rival(dcon::nation_id{});
 			}
 		}
+
+		if(n.get_ai_is_threatened() && !n.get_is_mobilized()) {
+			command::execute_toggle_mobilization(state, n);
```
schombert, line 117: A nation mobilizing while at peace is a real bad idea. Even when at war, it will kill your economy and can make you lose the war if you can't buy enough supplies to recover org

src/ai/ai.cpp
```diff
@@ -2499,6 +2525,15 @@ void make_peace_offers(sys::state& state) {
 }
 
 bool will_accept_peace_offer(sys::state& state, dcon::nation_id n, dcon::nation_id from, dcon::peace_offer_id p) {
+	// meta play
+	if(!state.world.nation_get_is_player_controlled(n) && !state.world.nation_get_is_player_controlled(from)) {
+		// smaller unciv gives in, to allow bigger unciv to be bigger
```
schombert, line 2530: I am not interested in having the ai throw the game

src/ai/ai.cpp
schombert: If you want the ai to do different things when rebels are activated, that is fine, but you cannot use the is_threatened flag for it

wxwisiasdf:
:(

PR #498, opened 2023-09-02
Fixes for keeping the map from moving while in topbar windows.

Masterchef118:


src/gui/gui_element_types.hpp
```diff
@@ -331,8 +331,12 @@ class generic_tabbed_window : public main_window_element_base {
 class generic_close_button : public button_element_base {
 public:
 	void button_action(sys::state& state) noexcept override {
-		if(parent)
+		if(parent) {
 			parent->set_visible(state, false);
+			if(!state.ui_state.can_move_map_while_visible) {
+				state.ui_state.can_move_map_while_visible = true;
```
schombert, line 337: This button is used in a lot of places. For example, it is used in the build factory subwindow. This would mean that if you are in the production window, and then open and close the build-a-factory window, that you would be able to move the map again while in the production window

schombert: (to be clear, if you *cancelled* out of building)

schombert:
I am willing to pull this fix in as it stands. However, as you can see, trying to keep track of which windows are open by updating a boolean at every place they might open or close is clearly no trivial task (this is the famous problem of cache invalidation). One way to make things easier on yourself might be instead to update the boolean in each top window's
```
virtual void on_visible(sys::state& state) noexcept { }
virtual void on_hide(sys::state& state) noexcept { }
```
functions instead rather than every place that they are made visible or hidden (although ... I don't think these functions have been used elsewhere, so they may not be well tested).

Another solution would be to get rid of the boolean altogether, and instead recreate it at the point of use by checking the visibility of each of the full-screen windows when a key is pressed, thereby eliminating the possibility that the cached value could get out of sync completely.

If you wish to pursue either of those solutions I can hold off on merging this. Or not; it is really up to you.

PR #502, opened 2023-09-10
save files independent from user-specific scenario so they can be shared

wxwisiasdf:
basically, allows sharing savefiles without having to share the scenario beforehand (but it must have the same "scenario save data", i.e the same game files)


schombert:
If you want to rely on checksums, we need to use a real checksum algorithm. Something from https://en.wikipedia.org/wiki/Category:Checksum_algorithms or https://en.wikipedia.org/wiki/Secure_Hash_Algorithms would be acceptable.

schombert:
and definitely more than 32 bits. 128 bits at the minimum

schombert:
You current checksum, by the way, is only 8 bits, since it xors byte by byte

wxwisiasdf:
> You current checksum, by the way, is only 8 bits, since it xors byte by byte

did you read the new checksum algorithm? i changed it

schombert:
you can try reporting the compiler error to the visual studio team. However, I obviously can't merge if it crashes the compiler

schombert:
unless, are all our build crashing because the CI compiler isn't up to date?

src/gamestate/serialization.hpp
```diff
@@ -128,19 +128,21 @@ uint8_t const* deserialize(uint8_t const* ptr_in, ankerl::unordered_dense::map<u
...
 }
 
-constexpr inline uint32_t save_file_version = 27;
-constexpr inline uint32_t scenario_file_version = 85 + save_file_version;
+constexpr inline uint32_t save_file_version = 28;
+constexpr inline uint32_t scenario_file_version = 86 + save_file_version;
 
 struct scenario_header {
 	uint32_t version = scenario_file_version;
 	uint32_t count = 0;
```
schombert, line 136: you can remove the count and timestamp fields both here and below if you aren't using them anymore

wxwisiasdf: I plan on still using count and timestamp for bookkeeping purpouses

PR #503, opened 2023-09-11
move key for lowering speed to numpad

yrrrmmbl:
implementation with one shortcut key per button seems kinda weird. this should be something handled in system_state.cpp so that both plus and minus keys on the keyboard work


src/gui/gui_topbar.hpp
```diff
@@ -975,7 +975,7 @@ class topbar_speeddown_button : public button_element_base {
 public:
 	void on_create(sys::state& state) noexcept override {
 		button_element_base::on_create(state);
-		base_data.data.button.shortcut = sys::virtual_key::MINUS;
+		base_data.data.button.shortcut = sys::virtual_key::SUBTRACT;
```
schombert, line 978: Instead of doing this, make one of the key codes mapped to the other when we send key through the ui. That way both keyboard keys can be used to adjust the speed (not all keyboards have a numpad)

PR #504, opened 2023-09-11
misc fixes

wxwisiasdf:
- fix chat message crash when long message
- improve multiplayer handling of savefiles (on nation picker, "PLAY" button will be greyed out if checksum doesn't match)
- provide checksumming everytime a save is loaded or when a client connects so we dont enter into an invalid game state that breaks everything

rationale for having two checksums:
1. is for scenario, scenario checks for savefiles and whatnot
2. the other is for the session, meant to be to check if a client can interact with the game state of a host properly

### DIFF
--- a/src/gamestate/commands.cpp
+++ b/src/gamestate/commands.cpp
@@ -4483,7 +4483,7 @@ void execute_notify_player_picks_nation(sys::state& state, dcon::nation_id sourc
 
 void execute_advance_tick(sys::state& state, dcon::nation_id source, sys::checksum_key& checksum) {
 #ifndef NDEBUG
-	sys::checksum_key current = state.get_network_checksum();
+	sys::checksum_key current = state.get_save_checksum();
 	if(!current.is_equal(checksum)) {
 #ifdef _WIN64
 		std::string msg = "Network has gotten out of sync: ";
@@ -4506,7 +4506,7 @@ void update_session_info(sys::state& state, dcon::nation_id source) {
 	p.type = command::command_type::update_session_info;
 	p.source = source;
 	p.data.update_session_info.seed = state.game_seed;
-	p.data.update_session_info.checksum = state.get_network_checksum();
+	p.data.update_session_info.checksum = state.get_save_checksum();
 	add_to_command_queue(state, p);
 }
 
--- a/src/gamestate/commands.cpp
+++ b/src/gamestate/commands.cpp
@@ -10,36 +10,37 @@ bool is_console_command(command_type t) {
 
 static void add_to_command_queue(sys::state& state, payload& p) {
 	switch(p.type) {
-		case command_type::notify_player_joins:
-		case command_type::notify_player_leaves:
-		case command_type::notify_player_picks_nation:
-		case command_type::notify_player_ban:
-		case command_type::notify_player_kick:
-			// Notifications can be sent because it's an-always do thing
-			break;
-		default:
-			// Normal commands are discarded iff we are not in the game
-			if(state.mode != sys::game_mode_type::in_game)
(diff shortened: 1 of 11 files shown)


src/gamestate/commands.cpp
```diff
@@ -4479,352 +4481,384 @@ void execute_notify_player_picks_nation(sys::state& state, dcon::nation_id sourc
...
-			break;
-		case command_type::even_split_navy:
-			execute_evenly_split_navy(state, c.source, c.data.navy_movement.n);
-			break;
+void execute_advance_tick(sys::state& state, dcon::nation_id source, sys::checksum_key& checksum) {
+#ifndef NDEBUG
+	sys::checksum_key current = state.get_network_checksum();
+	if(!current.is_equal(checksum)) {
+#ifdef _WIN64
+		std::string msg = "Network has gotten out of sync: ";
```
schombert, line 4489: We can't be serializing the game state out every tick to generate a checksum. Maybe in debug, but we can't be sending it out every day in release

wxwisiasdf: My bad, it was meant to be debug only

wxwisiasdf: Wait a minute, there is literally a #ifndef NDEBUG above it

schombert: Where is the code that *generates* the value, however?

wxwisiasdf: system_state.cpp:3387
```c++
sys::checksum_key state::get_network_checksum() {
	auto buffer = std::unique_ptr<uint8_t[]>(new uint8_t[sizeof_save_section(*this)]);

	dcon::load_record loaded = world.make_serialize_record_store_save();
	std::byte* start = reinterpret_cast<std::byte*>(buffer.get());
	world.serialize(start, loaded);

	auto buffer_position = reinterpret_cast<uint8_t*>(start);
	int32_t total_size_used = static_cast<int32_t>(buffer_position - buffer.get());

	checksum_key key;
	blake2b(&key, sizeof(key), buffer.get(), total_size_used, nullptr, 0);
	return key;
}
```

schombert: wrap that n #ifndef NDEBUG with a case to return a dummy key in release -- this prevents any mistakes from happening there. Oh, and put the ifndef around the body of the function that puts the message into the queue too. Finally, make sure that it isn't going to be running in single player -- making single player performance that much worse in debug will kill me since I need it to find and fix problems.

wxwisiasdf: I ensured it is only called in multiplayer contexts

wxwisiasdf: or when loading/saving

schombert: Just idiot proof things as much as possible, because calling that function by mistake anywhere it doesn't need to be called would be a huge performance drain -- that's 100s of MB of data that is being written out every time you call it.

wxwisiasdf: that would break savefiles and scenarios, if we return dummy keys that is

schombert:
Yeah, now that I look things over, your existing use of `header.checksum = state.get_network_checksum();` etc are broken. They are serializing out the state twice -- once to save and once to generate the checksum. Please fix that

schombert:
Furthermore `sys::checksum_key state::get_network_checksum()` returns a checksum based on the save portion, not the scenario portion. Thus, after a save file advances more than a day, it will no longer match the scenario checksum and be able to be loaded. Please fix that too.

wxwisiasdf:
There we go this should do!

PR #508, opened 2023-09-16
blackflag units after wars

wxwisiasdf:


wxwisiasdf:
fixes issue were ai wouldnt move units after wars

src/military/military.cpp
```diff
@@ -2599,6 +2599,8 @@ void cleanup_war(sys::state& state, dcon::war_id w, war_result result) {
 	}
 
 	state.world.delete_war(w);
+
+	update_blackflag_status(state);
```
schombert, line 2603: this won't actually do anything (see the details of the function). It is also redundant  -- as part of the war ending nations are removed from it. That triggers updating the blag flag status of all armies later in the day

schombert: I do see that I forgot to put the call to this function in after the military gc runs. That probably explains the issue you are seeing. Let me push that up and see if it fixes things for you

wxwisiasdf: doesnt seem to be the case through

wxwisiasdf:
seems kosher now

PR #511, opened 2023-09-20
Added label to list of puppets/overlord in nation picker. Added satellite/substate status to nation status in diplomacy window.

Masterchef118:


src/gui/topbar_subwindows/gui_diplomacy_window.hpp
```diff
@@ -1083,6 +1083,43 @@ class great_power_inf_detail : public window_element_base {
 	}
 };
 
+template<class T>
+class nation_status_diplomacy_window : public simple_text_element_base {
```
schombert, line 1087: This duplicates the contents of an element in gui_common_elements 585. If we need to use it more than once, the function status enum -> text should probably be factored out (in case it changes, so that one of the two ui elements won't end up out of sync)

PR #514, opened 2023-09-24
Genericize building logic

wxwisiasdf:


schombert:
Before I delve into reviewing this, could you explain the benefits of this change? Merging the logic for building factories and province buildings (only to then end up with ifs separating them out again) seems like it adds a fair amount of complexity. What are the benefits of this change?

wxwisiasdf:
> Before I delve into reviewing this, could you explain the benefits of this change? Merging the logic for building factories and province buildings (only to then end up with ifs separating them out again) seems like it adds a fair amount of complexity. What are the benefits of this change?

makes it easier to add new buildings, and to simplify the existing logic for both parsing and econ sim

src/economy/economy.cpp
```diff
@@ -1398,7 +1398,8 @@ void populate_private_construction_consumption(sys::state& state) {
 
 	for(auto c : state.world.in_province_building_construction) {
 		auto owner = c.get_nation().id;
-		if(owner && owner == c.get_province().get_nation_from_province_control() && c.get_is_pop_project() && province_building_type(c.get_type()) == economy::province_building_type::railroad) {
+		// Rationale for not checking building type: Its an invalid state; should not occur under normal circumstances
+		if(owner && owner == c.get_province().get_nation_from_province_control() && c.get_is_pop_project()) {
```
schombert, line 1402: we check building type because both public and private province building constructions can end up in this relationship. But only the privately built railroads are tallied up here

src/economy/economy.cpp
```diff
@@ -1880,8 +1881,11 @@ void advance_construction(sys::state& state, dcon::nation_id n) {
 
 	for(auto c : state.world.nation_get_province_building_construction(n)) {
 		if(c.get_province().get_nation_from_province_ownership() == c.get_province().get_nation_from_province_control()) {
+			auto t = economy::province_building_type(c.get_type());
+			// Rationale for not checking the building type:
+			// Pop projects created for forts and naval bases should NOT happen in the first place, so checking against them
```
schombert, line 1886: There is no check removed here, so I don't know what this comment means

src/economy/economy.cpp
schombert: ah I see, you are referring to not checking for being a railroad

src/gamestate/commands.cpp
```diff
@@ -430,7 +430,7 @@ void execute_begin_province_building_construction(sys::state& state, dcon::natio
 			si.set_naval_base_is_taken(true);
 	}
 
-	if(type == economy::province_building_type::railroad && source != state.world.province_get_nation_from_province_ownership(p)) {
+	if(type != economy::province_building_type::fort && type != economy::province_building_type::naval_base && source != state.world.province_get_nation_from_province_ownership(p)) {
```
schombert, line 433: probably should stick with checking for RR here, unless TGC has ui for doing foreign investment with the new buildings??

src/parsing/econ_parsing.cpp
```diff
@@ -34,12 +34,29 @@ void make_goods_group(std::string_view name, token_generator& gen, error_handler
 	}
 }
 
+void building_definition::type(association_type, std::string_view value, error_handler& err, int32_t line, scenario_building_context& context) {
```
schombert, line 37: this looks like the handler for a parser routine, but I don't see the corresponding function added to the .txt file

src/parsing/econ_parsing.cpp
```diff
@@ -34,12 +34,29 @@ void make_goods_group(std::string_view name, token_generator& gen, error_handler
...
+		err.accumulated_errors +=
+			"Unknown building type " + std::string(value) + " in file " + err.file_name + " line " + std::to_string(line) + "\n";
+	}
+}
+
 void building_file::result(std::string_view name, building_definition&& res, error_handler& err, int32_t line,
 		scenario_building_context& context) {
 	// no, this messes things up
 	// res.goods_cost.data.safe_get(dcon::commodity_id(0)) = float(res.cost);
 	switch(res.stored_type) {
```
schombert, line 58: I think this is now a switch with one case and one default: i.e. an if-else statement

src/parsing/econ_parsing.cpp
schombert: ah, I see, moved from below

src/gamestate/commands.cpp
wxwisiasdf: yes it does

PR #524, opened 2023-10-11
province_selector support

wxwisiasdf:
Problem: The province_selector is a building type included in various mods, to literally, "select" provinces, this behaviour, while very clumsy, has been included in a handful of mods.

Why: The purpouse of this is being able to "select" multiple provinces as immigration/emigration targets on mods, or as generally being able to "select" multiple provinces for other purpouses.

Solution: Have a bit on the province object indicating if it is selected or not, while this behaviour might not be the best - the intention of "selecting" a province is henceforth reflected properly. Mods provide UI for "selecting" provinces via a checkbox, however a better solution would perhaps be entering into a "select" mapmode so we select various provinces for whichever purpouse we want.

### DIFF
--- a/src/economy/economy.hpp
+++ b/src/economy/economy.hpp
@@ -53,9 +53,7 @@ inline std::string_view province_building_type_get_level_text(economy::province_
 
 struct global_economy_state {
 	building_information building_definitions[max_building_types];
-	bool province_selector_defined = false;
 	dcon::modifier_id selector_modifier{};
-	bool province_immigrator_defined = false;
 	dcon::modifier_id immigrator_modifier{};
 	float craftsmen_fraction = 0.8f;
 };
@@ -177,5 +175,8 @@ int32_t most_recent_price_record_index(sys::state& state);
 int32_t previous_price_record_index(sys::state& state);
 
 void prune_factories(sys::state& state); // get rid of closed factories in full states
+
+dcon::modifier_id get_province_selector_modifier(sys::state& state);
+dcon::modifier_id get_province_immigrator_modifier(sys::state& state);
 
 } // namespace economy
--- a/src/economy/economy.hpp
+++ b/src/economy/economy.hpp
@@ -53,6 +53,7 @@ inline std::string_view province_building_type_get_level_text(economy::province_
 
 struct global_economy_state {
 	building_information building_definitions[max_building_types];
+	bool province_selector_defined = false;
 	float craftsmen_fraction = 0.8f;
 };
 
--- a/src/economy/economy.hpp
+++ b/src/economy/economy.hpp
@@ -54,6 +54,9 @@ inline std::string_view province_building_type_get_level_text(economy::province_
 struct global_economy_state {
 	building_information building_definitions[max_building_types];
 	bool province_selector_defined = false;
+	dcon::modifier_id selector_modifier{};
+	bool province_immigrator_defined = false;
+	dcon::modifier_id immigrator_modifier{};
 	float craftsmen_fraction = 0.8f;
 };
 
--- a/src/gamestate/commands.cpp
+++ b/src/gamestate/commands.cpp
@@ -3865,10 +3865,12 @@ void execute_toggle_select_province(sys::state& state, dcon::nation_id source, d
 
 	for(auto m : state.world.province_get_current_modifiers(prov)) {
 		if(m.mod_id == state.economy_definitions.selector_modifier) {
+			// Remove the already existing modifier if it was found
 			sys::remove_modifier_from_province(state, prov, m.mod_id);
 			return;
 		}
 	}
+	// Otherwise, if it is not present on the province, add it
 	sys::add_modifier_to_province(state, prov, state.economy_definitions.selector_modifier, sys::date{});
 }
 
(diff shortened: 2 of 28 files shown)


src/gui/gui_trigger_tooltips.cpp
```diff
@@ -7581,6 +7581,12 @@ void tf_has_building_university(TRIGGER_DISPLAY_PARAMS) {
 	display_with_has_comparison(tval[0], text::produce_simple_string(ws, "a_university"), ws, layout, box);
 	text::close_layout_box(layout, box);
 }
+void tf_has_building_province_selector(TRIGGER_DISPLAY_PARAMS) {
+	auto box = text::open_layout_box(layout, indentation);
+	make_condition(tval, ws, layout, primary_slot, this_slot, from_slot, indentation, show_condition, box);
+	display_with_has_comparison(tval[0], text::produce_simple_string(ws, "a_province_selector"), ws, layout, box);
```
schombert, line 7587: shouldn't this read "is / is not selected" ?

schombert:
I don't see the logic that would make the button in the province window for this actually work. Wouldn't that need an entirely new command?

src/economy/economy.hpp
```diff
@@ -53,6 +53,10 @@ inline std::string_view province_building_type_get_level_text(economy::province_
 
 struct global_economy_state {
 	building_information building_definitions[max_building_types];
+	bool province_selector_defined = false;
+	dcon::modifier_id selector_modifier{};
+	bool province_immigrator_defined = false;
+	dcon::modifier_id immigrator_modifier{};
```
schombert, line 59: you don't need the booleans; you can just check whether, for example, selector_modifier is a valid id (by casting it to bool)

src/gamestate/commands.cpp
```diff
@@ -3844,6 +3844,62 @@ void execute_toggle_rebel_hunting(sys::state& state, dcon::nation_id source, dco
...
+	payload p;
+	memset(&p, 0, sizeof(payload));
+	p.type = command_type::toggle_select_province;
+	p.source = source;
+	p.data.generic_location.prov = prov;
+	add_to_command_queue(state, p);
+}
+bool can_toggle_select_province(sys::state& state, dcon::nation_id source, dcon::province_id prov) {
+	if(state.world.province_get_nation_from_province_control(prov) != source)
+		return false;
```
schombert, line 3857: while the control condition does match the condition for buildings, my guess is that it is unwanted by the mod makers. They probably want to be able to select occupied provinces

src/gamestate/commands.cpp
```diff
@@ -3844,6 +3844,62 @@ void execute_toggle_rebel_hunting(sys::state& state, dcon::nation_id source, dco
...
+	return true;
+}
+void execute_toggle_select_province(sys::state& state, dcon::nation_id source, dcon::province_id prov) {
+	if(!can_toggle_select_province(state, source, prov))
+		return;
+
+	for(auto m : state.world.province_get_current_modifiers(prov)) {
+		if(m.mod_id == state.economy_definitions.selector_modifier) {
+			sys::remove_modifier_from_province(state, prov, m.mod_id);
```
schombert, line 3868: remove modifier from province already does this loop to find the modifier

src/gamestate/commands.cpp
```diff
@@ -3844,6 +3844,62 @@ void execute_toggle_rebel_hunting(sys::state& state, dcon::nation_id source, dco
...
+			return;
+		}
+	}
+	sys::add_modifier_to_province(state, prov, state.economy_definitions.selector_modifier, sys::date{});
```
schombert, line 3872: similarly, add modifier to province is going to loop over them to ensure that the modifier is not already present

src/gui/gui_effect_tooltips.cpp
```diff
@@ -6321,6 +6321,43 @@ uint32_t ef_university(EFFECT_DISPLAY_PARAMS) {
 	return 0;
 }
 
+uint32_t ef_add_selector(EFFECT_DISPLAY_PARAMS) {
```
schombert, line 6324: you don't need these additional effect functions. They can be turned into add/remove province modifier effects

src/gui/gui_trigger_tooltips.cpp
```diff
@@ -7572,13 +7572,26 @@ void tf_religion_nation_this_pop(TRIGGER_DISPLAY_PARAMS) {
...
 }
 void tf_has_building_university(TRIGGER_DISPLAY_PARAMS) {
 	auto box = text::open_layout_box(layout, indentation);
 	make_condition(tval, ws, layout, primary_slot, this_slot, from_slot, indentation, show_condition, box);
-	display_with_has_comparison(tval[0], text::produce_simple_string(ws, "a_university"), ws, layout, box);
+	display_with_has_comparison(tval[0], text::produce_simple_string(ws, "tt_a_university"), ws, layout, box);
+	text::close_layout_box(layout, box);
+}
+
+void tf_has_province_selector(TRIGGER_DISPLAY_PARAMS) {
```
schombert, line 7585: see above about reusing has_province_modifier

src/gamestate/commands.cpp
schombert: Yes, I understand that. However, by not going through those functions, you can run the loop just once instead of having them run it redundantly

PR #530, opened 2023-10-15
Fog of war

wxwisiasdf:


src/map/map.hpp
```diff
@@ -100,6 +101,7 @@ class display_data {
 
 	// map pixel -> province id
 	std::vector<uint16_t> province_id_map;
+	bool fow_enabled = false;
```
schombert, line 104: why not make this a persistent setting accessible through the esc->graphics menu?

src/map/map_state.cpp
```diff
@@ -98,6 +98,7 @@ void map_state::update(sys::state& state) {
 		last_update_time = now;
 
 	update_unit_arrows(state, map_data);
+	map_data.update_fog_of_war(state);
```
schombert, line 101: Doing this here results in the fog being recalculated every frame. It only needs to be recalculated when the game state changes

PR #532, opened 2023-10-15
UI fixups

wxwisiasdf:
- add brown colour ($I) used in some mods(!!!)
- use retrieve and send idioms
- checkbox instead of using a lot of buttons in the main menu

### DIFF
--- a/assets/alice.gui
+++ b/assets/alice.gui
@@ -110,7 +110,7 @@ guiTypes = {
 			}
 
 			instantTextBoxType={
-				position = {x = 58 y = 115 }
+				position = {x = 68 y = 115 }
 				name = "window_mode_label"
 				text = "window_mode_label"
 				font = "Arial16"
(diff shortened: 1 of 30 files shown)


assets/alice.gui
```diff
@@ -110,73 +110,37 @@ guiTypes = {
...
-				maxWidth = 236
-				maxHeight = 18
-				orientation = "UPPER_LEFT"
-				format = centre
+				format = left
 			}
-
 			guiButtonType = {
-				name = "projection_left"
+				name = "projection_checkbox"
```
schombert, line 141: you can't reduce the projections to a checkbox because I probably eventually want to add more of them

src/text/text.cpp
```diff
@@ -34,6 +34,8 @@ text_color char_to_color(char in) {
 		return text_color::dark_red;
 	case 'J':
 		return text_color::dark_green;
+	case 'I':
```
schombert, line 37: someone uses I??? for brown. I'll accept it, but I won't like it

PR #533, opened 2023-10-15
Add sound effects

wxwisiasdf:


src/gamestate/system_state.hpp
```diff
@@ -152,6 +152,8 @@ struct user_settings_s {
 		message_response::standard_popup,//cb_fab_finished = 95,
 		message_response::standard_popup,//cb_fab_cancelled = 96,
 		message_response::ignore,//crisis_voluntary_joi_on = 97,
+		message_response::log,//army_built = 98,
+		message_response::log,//navy_built = 99,
```
schombert, line 156: unfortunately, you can't do this -- existing settings are memcopied out of the file, and adding new settings in the middle will break everything very badly

PR #537, opened 2023-10-15
Fix colonial icon tooltip

wxwisiasdf:


src/gui/gui_topbar.hpp
```diff
@@ -1518,27 +1518,27 @@ class topbar_colony_icon : public standard_nation_button {
...
-				});
-			} else {
-				text::add_line(state, contents, "countryalert_no_colonial");
+		auto nation_id = retrieve<dcon::nation_id>(state, parent);
+		auto nation_fat_id = dcon::fatten(state.world, nation_id);
+
+		bool is_empty = true;
+		nation_fat_id.for_each_colonization([&](dcon::colonization_id colony) {
+			auto sdef = state.world.colonization_get_state(colony);
+			if(province::can_invest_in_colony(state, nation_id, sdef)) {
```
schombert, line 1527: you also need a check for when the nation can end the colonial race and turn it into a colony. This is not covered by can invest (check get_colonization_stage == 3)

PR #539, opened 2023-10-15
Added cancel all units button in military build unit window.

Masterchef118:


schombert:
would you mind adding a tooltip? You could use `MIL_CONSTRUCTION_CANCEL_ALL` which is `Cancel all constructions`

PR #549, opened 2023-10-17
Land and Naval battles SFX

wxwisiasdf:


src/gui/gui_land_combat.hpp
```diff
@@ -984,6 +984,12 @@ class land_combat_window : public window_element_base {
 
 	}
 
+	void render(sys::state& state, int32_t x, int32_t y) noexcept override {
+		window_element_base::render(state, x, y);
+		// Play sound effects
+		sound::play_effect(state, sound::get_random_land_battle_sound(state), state.user_settings.effects_volume * state.user_settings.master_volume);
```
schombert, line 990: this would have the effect of trying to play the sound every frame that the window remains open. How about just when it is opened instead?

PR #553, opened 2023-10-18
provide more info on decision window (ai_will_do + potential trigger)…

wxwisiasdf:
… and fix substitutions missing


schombert:
this was a good idea

PR #560, opened 2023-10-20
cheat commands

wxwisiasdf:
- instant research
- instant construction
- always accept deals (yesman)
- always enact reforms
- always add wargoals
- toggle ai on/off

### DIFF
--- a/src/culture/culture.cpp
+++ b/src/culture/culture.cpp
@@ -915,6 +915,21 @@ void update_research(sys::state& state, uint32_t current_year) {
 			}
 		}
 	}
+
+	if(state.cheat_data.instant_research && state.world.nation_get_current_research(state.local_player_nation)) {
+		auto n = dcon::fatten(state.world, state.local_player_nation);
+		apply_technology(state, n, n.get_current_research());
+		notification::post(state, notification::message{
+			[t = n.get_current_research()](sys::state& state, text::layout_base& contents) {
+				text::add_line(state, contents, "msg_tech_1", text::variable_type::x, state.world.technology_get_name(t));
+				ui::technology_description(state, contents, t);
+			},
+			"msg_tech_title",
+			n,
+			sys::message_setting_type::tech
+		});
+		n.set_current_research(dcon::technology_id{});
+	}
 }
 
 void discover_inventions(sys::state& state) {
--- a/src/culture/culture.cpp
+++ b/src/culture/culture.cpp
@@ -915,21 +915,6 @@ void update_research(sys::state& state, uint32_t current_year) {
 			}
 		}
 	}
-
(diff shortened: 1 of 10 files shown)


src/culture/culture.cpp
```diff
@@ -915,6 +915,21 @@ void update_research(sys::state& state, uint32_t current_year) {
 			}
 		}
 	}
+
+	if(state.cheat_data.instant_research && state.world.nation_get_current_research(state.local_player_nation)) {
```
schombert, line 919: instant research doesn't need to be its own command because filling up the rp finishes it instantly anyways

schombert: (and it comes at the cost of additional complexity here)

wxwisiasdf: Difference is that:
- `instant research` is active (it will continue to apply as long as its ON)
- whereas `rp` only does a one-time fillup

src/economy/economy.cpp
```diff
@@ -3812,6 +3812,10 @@ void resolve_constructions(sys::state& state) {
 				break;
 			}
 		}
+
+		if(state.world.province_building_construction_get_nation(c) == state.local_player_nation && state.cheat_data.instant_construction)
```
schombert, line 3816: similarly, I would rather see a "finish buildings" cheat that fills up the purchased construction goods once in the command rather than an additional check here

src/culture/culture.cpp
schombert: yeah, but at the cost of this additional complexity here. Which is not worth it in my  opinion

src/gamestate/cheats.cpp
```diff
@@ -121,6 +121,25 @@ void c_end_game(sys::state& state, dcon::nation_id source) {
...
+	add_to_command_queue(state, p);
+}
+void execute_c_complete_constructions(sys::state& state, dcon::nation_id source) {
+	for(uint32_t i = state.world.province_building_construction_size(); i-- > 0;) {
+		dcon::province_building_construction_id c{ dcon::province_building_construction_id::value_base_t(i) };
+		auto t = province_building_type(state.world.province_building_construction_get_type(c));
+		auto const& base_cost = state.economy_definitions.building_definitions[int32_t(t)].cost;
+		auto& current_purchased = state.world.province_building_construction_get_purchased_goods(c);
+		for(uint32_t j = 0; j < commodity_set::set_size; ++j)
+			current_purchased.commodity_amounts[j] = base_cost.commodity_amounts[j];
```
schombert, line 139: the target amount to finish the building may be increased by a number of factors. For province buildings, you can get around this by multiplying the base cost by 2, which is the maximum it could ever be increased to

PR #565, opened 2023-10-22
Better capitalist AI

wxwisiasdf:


src/ai/ai.cpp
```diff
@@ -660,13 +668,36 @@ void update_focuses(sys::state& state) {
...
 				} else {
-					auto cfrac = state.world.state_instance_get_demographics(ordered_states[i], demographics::to_key(state, state.culture_definitions.clergy)) / state.world.state_instance_get_demographics(ordered_states[i], demographics::total);
+					auto total = state.world.state_instance_get_demographics(ordered_states[i], demographics::total);
+					auto cfrac = state.world.state_instance_get_demographics(ordered_states[i], demographics::to_key(state, state.culture_definitions.clergy)) / total;
+					auto pwfrac = state.world.state_instance_get_demographics(ordered_states[i], demographics::to_key(state, state.culture_definitions.primary_factory_worker)) / total;
+					auto swfrac = state.world.state_instance_get_demographics(ordered_states[i], demographics::to_key(state, state.culture_definitions.secondary_factory_worker)) / total;
 					if(cfrac < state.defines.max_clergy_for_literacy * 0.8f) {
 						state.world.state_instance_set_owner_focus(ordered_states[i], state.national_definitions.clergy_focus);
 						--num_focuses_total;
+					} else if(int32_t(pwfrac * 100.f) > int32_t(swfrac * 100.f)) {
```
schombert, line 678: first, why are you casting to ints ?
secondly, this would aim towards a 50-50 ratio, which is *not* the ideal ratio

src/ai/ai.cpp
```diff
@@ -660,13 +668,36 @@ void update_focuses(sys::state& state) {
...
+						// Keep balance between ratio of factory workers
+						state.world.state_instance_set_owner_focus(ordered_states[i], state.national_definitions.secondary_factory_worker_focus);
+						--num_focuses_total;
+					} else if(int32_t(swfrac * 100.f) > int32_t(pwfrac * 100.f)) {
+						// Keep balance between ratio of factory workers
+						state.world.state_instance_set_owner_focus(ordered_states[i], state.national_definitions.primary_factory_worker_focus);
+						--num_focuses_total;
+					} else {
```
schombert, line 686: And this won't work at all
If you have an unconditional else, this will result in the AI only using focuses on the N most populated states (where N is their number of focuses)

PR #571, opened 2023-10-22
Technology queue

wxwisiasdf:
right -> remove tech
shift -> add tech to queue

### DIFF
--- a/src/gamestate/system_state.cpp
+++ b/src/gamestate/system_state.cpp
@@ -754,9 +754,16 @@ void state::render() { // called to render the frame may (and should) delay retu
 	if(game_state_was_updated) {
 		if(!ui_state.tech_queue.empty()) {
 			if(!world.nation_get_current_research(local_player_nation)) {
-				auto tech = ui_state.tech_queue.back();
-				command::start_research(*this, local_player_nation, tech);
-				ui_state.tech_queue.pop_back();
+				auto it = ui_state.tech_queue.begin();
+				if(command::can_start_research(*this, local_player_nation, *it)) {
+					// can research, so research it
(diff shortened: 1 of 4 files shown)


src/gamestate/system_state.cpp
```diff
@@ -752,6 +752,14 @@ void state::render() { // called to render the frame may (and should) delay retu
 	}
 
 	if(game_state_was_updated) {
+		if(!ui_state.tech_queue.empty()) {
+			if(!world.nation_get_current_research(local_player_nation)) {
+				auto tech = ui_state.tech_queue.back();
+				command::start_research(*this, local_player_nation, tech);
```
schombert, line 758: this will not permit queuing up future techs: they will fail to start and get popped from the queue

src/gamestate/system_state.cpp
```diff
@@ -754,9 +754,16 @@ void state::render() { // called to render the frame may (and should) delay retu
...
 			if(!world.nation_get_current_research(local_player_nation)) {
-				auto tech = ui_state.tech_queue.back();
-				command::start_research(*this, local_player_nation, tech);
-				ui_state.tech_queue.pop_back();
+				auto it = ui_state.tech_queue.begin();
+				if(command::can_start_research(*this, local_player_nation, *it)) {
+					// can research, so research it
+					command::start_research(*this, local_player_nation, *it);
+					ui_state.tech_queue.erase(it);
+				} else {
```
schombert, line 762: don't do this, as it reorders the queue
instead: search for the first tech you can start, from the front to the back
on finding a tech, erase it from the queue, preserving its order
else, do nothing

PR #575, opened 2023-10-23
Made units reinforce to max strength always.

Masterchef118:


schombert:
you should undo leaf's ui additions as part of this pr

Masterchef118:
Oh yeah

schombert:
you can probably find them via the commit history and undo them that way

PR #576, opened 2023-10-23
make full reinforcement moddable

wxwisiasdf:
I like the option of being able to toggle it on/off


schombert:
I really don't want to support both as an option. And won't all of your ui be confusing when the plain V2 way is in effect?

wxwisiasdf:
> I really don't want to support both as an option. And won't all of your ui be confusing when the plain V2 way is in effect?

i updated the UI accordingly, and this is not a boolean, this just determines the MINIMUM amount of reinforcement "fraction" that is done at the start, so for example you can make a mod that caps it at 0.5, so all regiments will be >1,500 and <=3,000 - it is very useful for mods like CWE where regiment size becomes an issue

PR #581, opened 2023-10-25
support spreaded diplomacy files

wxwisiasdf:
supports diplomacy files like `history/diplomacy/Japan.txt` yknow, since, yknow, HFM, and HPM, yknow


schombert:
`parser_defs_generated.hpp:48045:2: error: unknown type name 'diplomacy_file'
        diplomacy_file cobj;`

schombert:
It still doesn't like it

PR #583, opened 2023-10-25
partialy retreat armies from battle

wxwisiasdf:


schombert:
this isn't going to make the ai cycle because they won't move the units back into battle when reinforced

PR #594, opened 2023-10-28
Release puppet action

wxwisiasdf:
image

### DIFF
--- a/src/gamestate/commands.cpp
+++ b/src/gamestate/commands.cpp
@@ -3719,12 +3719,7 @@ void release_subject(sys::state& state, dcon::nation_id source, dcon::nation_id 
 	add_to_command_queue(state, p);
 }
 bool can_release_subject(sys::state& state, dcon::nation_id source, dcon::nation_id target) {
-	auto ovr = state.world.nation_get_overlord_as_subject(target);
-	if(!bool(ovr))
-		return false;
-	if(state.world.overlord_get_ruler(ovr) != source)
-		return false;
-	return true;
+	return state.world.overlord_get_ruler(state.world.nation_get_overlord_as_subject(target)) != source;
 }
 void execute_release_subject(sys::state& state, dcon::nation_id source, dcon::nation_id target) {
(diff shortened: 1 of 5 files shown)


src/gamestate/commands.cpp
```diff
@@ -3710,6 +3710,28 @@ void execute_toggle_immigrator_province(sys::state& state, dcon::nation_id sourc
...
+	payload p;
+	memset(&p, 0, sizeof(payload));
+	p.type = command_type::release_subject;
+	p.source = source;
+	p.data.diplo_action.target = target;
+	add_to_command_queue(state, p);
+}
+bool can_release_subject(sys::state& state, dcon::nation_id source, dcon::nation_id target) {
+	auto ovr = state.world.nation_get_overlord_as_subject(target);
+	if(!bool(ovr))
```
schombert, line 3723: you don't need to make this check

src/gamestate/commands.cpp
```diff
@@ -3710,6 +3710,28 @@ void execute_toggle_immigrator_province(sys::state& state, dcon::nation_id sourc
...
+		return false;
+	if(state.world.overlord_get_ruler(ovr) != source)
```
schombert, line 3725: because without having an overload, it would return the invalid handle here, which will not be equal to source

src/gamestate/commands.cpp
```diff
@@ -3719,12 +3719,7 @@ void release_subject(sys::state& state, dcon::nation_id source, dcon::nation_id
 	add_to_command_queue(state, p);
 }
 bool can_release_subject(sys::state& state, dcon::nation_id source, dcon::nation_id target) {
-	auto ovr = state.world.nation_get_overlord_as_subject(target);
-	if(!bool(ovr))
-		return false;
-	if(state.world.overlord_get_ruler(ovr) != source)
-		return false;
-	return true;
+	return state.world.overlord_get_ruler(state.world.nation_get_overlord_as_subject(target)) != source;
```
schombert, line 3722: uh ... you mean == I think

PR #595, opened 2023-10-28
disallow ai from joining a war where the defender has high infamy

wxwisiasdf:


schombert:
This would still result in the AI dishonoring a defensive CTA, which will be a problem when we implement penalties for that

wxwisiasdf:
> This would still result in the AI dishonoring a defensive CTA, which will be a problem when we implement penalties for that

thats the point

PR #596, opened 2023-10-28
commands ignore diplo points if we're an AI and reuse execute for em to reduce code duplication

wxwisiasdf:


schombert:
The point of not calling execute in these cases is not *just* to avoid checking diplo points; it is to avoid all the command validity checks. The AI routines should already be doing those checks in deciding what action to do, and so checking them again when executing that choice is redundant, and potentially expensive depending on the nature of the checks

wxwisiasdf:
> The point of not calling execute in these cases is not _just_ to avoid checking diplo points; it is to avoid all the command validity checks. The AI routines should already be doing those checks in deciding what action to do, and so checking them again when executing that choice is redundant, and potentially expensive depending on the nature of the checks

Seems fair enough, removed them then

schombert:
reusing the commands isn't a bad idea. What I was saying is that you would want to use is not player controlled to skip the entire command validity check in command execution

schombert:
Or, if you want to really be awesome, *move* the command validity check to where the commands are about to be executed in the loop that pulls them off the queue. This would thus make direct calls bypass the check without any extra conditions added

wxwisiasdf:
actually that sounds better lemme do that yeah

wxwisiasdf:
Tested, poked around a bit, commands now can be used by AI while avoiding the checks :D

PR #597, opened 2023-10-28
allow france country name to show up

wxwisiasdf:


schombert:
As I said earlier, you should make the threshold proportional to the zoom level, so that as you zoom in smaller names start showing up

wxwisiasdf:
text is only regenerated when a province changes and it skips over text that is too small to be shown...

schombert:
Yeah, I know that is how it currently works. I am saying that that needs to change. And when someone makes that change (this is why we nag Erik) the thresholds for the visible size will end up being reworked anyways

PR #599, opened 2023-10-28
highlight desired factory types for the player

wxwisiasdf:


schombert:
I would like a screenshot to make sure this isn't annoying looking

PR #609, opened 2023-10-30
reset all nations to ai upon loading

wxwisiasdf:


src/gamestate/system_state.cpp
```diff
@@ -3064,6 +3064,10 @@ void state::preload() {
 }
 
 void state::fill_unsaved_data() { // reconstructs derived values that are not directly saved after a save has been loaded
+	// reset all nations to AI
+	for(const auto n : world.in_nation)
+		n.set_is_player_controlled(false);
```
schombert, line 3069: won't this break ordinary save loading? Is there something that will set is player controlled for the local player nation to true after this runs?

PR #610, opened 2023-10-30
properly identify artillery

wxwisiasdf:


src/parsing/military_parsing.cpp
```diff
@@ -110,75 +110,79 @@ void make_unit(std::string_view name, token_generator& gen, error_handler& err,
 		context.state.military_definitions.irregular = new_id;
 		err.accumulated_warnings += "Fallbacking to detecting an irregular from name " + err.file_name + "\n";
 	}
+	if(!bool(context.state.military_definitions.artillery) && name == "artillery") {
```
schombert, line 113: But ... you didn't remove the way artillery is actually detected, meaning that this assignment won't do anything

schombert:
again, the issue is that artillery is not serialized; it is picked as part of restoring unsaved data. Thus, the work you are doing to identify it is probably wasted. This may also explain why the game you ran with this in place crashed: there was no unit type for art. and thus the ai was building invalid units

PR #611, opened 2023-10-31
Upgrade released nation states to non-colonial.

NicholasLN:


src/gamestate/commands.cpp
```diff
@@ -967,6 +967,26 @@ void execute_release_and_play_as(sys::state& state, dcon::nation_id source, dcon
 	} else if(state.local_player_nation == holder) {
 		state.local_player_nation = source;
 	}
+
+	std::vector<dcon::state_instance_fat_id> state_list{};
+	for(auto si : state.world.nation_get_state_ownership(holder))
+		state_list.push_back(si.get_state());
+
+	for(dcon::state_instance_fat_id& state_id : state_list) {
```
schombert, line 975: you don't need to save the states in a collection only to then iterate over the members of the collection. Instead you can have just one loop over the owned states

PR #613, opened 2023-11-02
fixes for chat window

wxwisiasdf:


src/gui/gui_chat_window.hpp
```diff
@@ -93,7 +94,12 @@ class chat_message_listbox : public listbox_element_base<chat_message_entry, cha
 	}
 
 	mouse_probe impl_probe_mouse(sys::state& state, int32_t x, int32_t y, mouse_probe_type type) noexcept override {
-		return mouse_probe{ nullptr, ui::xy_pair{0,0} };
```
schombert, line 96: make sure that this change won't cause the window to intercept mouse events even when it is hidden. There was a problem previously where invisible chat messages would end up grabbing the mouse in single player, making it impossible to click on the map in the top left region

wxwisiasdf: > make sure that this change won't cause the window to intercept mouse events even when it is hidden. There was a problem previously where invisible chat messages would end up grabbing the mouse in single player, making it impossible to click on the map in the top left region

and that made a problem where the lobby window had a nonfunctional chat scrollbar, this fixes that while also having the previouis fix

PR #615, opened 2023-11-02
fix highlight not being a button and making weird issues with interac…

wxwisiasdf:
…tibility

### DIFF
--- a/src/gui/topbar_subwindows/gui_trade_window.hpp
+++ b/src/gui/topbar_subwindows/gui_trade_window.hpp
@@ -295,18 +295,18 @@ struct trade_details_select_commodity {
 	dcon::commodity_id commodity_id{};
 };
 
-class trade_commodity_entry_button : public button_element_base {
-	bool visible = true;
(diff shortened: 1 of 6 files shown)


src/gui/topbar_subwindows/gui_trade_window.hpp
```diff
@@ -492,7 +459,7 @@ class trade_commodity_entry : public window_element_base {
 		} else if(name == "selling_indicator") {
 			return make_element_by_type<commodity_stockpile_indicator>(state, id);
 		} else if(name == "automation_indicator") {
-			return make_element_by_type<commodity_ai_indicator>(state, id);
```
schombert, line 495: Wasn't this used to indicate whether the stockpile was filling or not? Or something like that?

wxwisiasdf: > Wasn't this used to indicate whether the stockpile was filling or not? Or something like that?

nope

PR #617, opened 2023-11-03
Improve spacing for main menu

wxwisiasdf:
makes spacing more uniform for all the tabs


assets/alice.gui
```diff
@@ -122,47 +118,78 @@ guiTypes = {
 			}
 			guiButtonType = {
 				name = "window_mode_checkbox"
-				position = { 51 115 }
```
schombert, line 125: this was intentional so that the text would be vertically centered with respect to the button

PR #620, opened 2023-11-03
fix user settings behaving weirdly

wxwisiasdf:


src/gamestate/system_state.cpp
```diff
@@ -1727,48 +1727,42 @@ void state::load_user_settings() {
...
 	if(settings_file) {
 		auto content = view_contents(*settings_file);
-		auto ptr = content.data;
-
+		char buffer[sizeof(user_settings_s)];
+		new (&buffer) user_settings_s{};
+		std::memcpy(buffer, content.data, content.file_size);
+		char const* ptr = &buffer[0];
 #define US_LOAD(x) \
-		if(ptr > content.data + content.file_size - sizeof(user_settings.x)) break; \
```
schombert, line 1733: you still want this or you are going to be loading 0 into any new settings. 0 may not be the correct default

wxwisiasdf: we are not memsetting, see above, we are new emplacing

schombert: it is a buffer of char. I would have to look up whether that will actually initialize, but at best you are going to get a buffer of 0s. Assuming the file isn't big enough, you will then memcpy those 0s into the settings

wxwisiasdf: > it is a buffer of char. I would have to look up whether that will actually initialize, but at best you are going to get a buffer of 0s. Assuming the file isn't big enough, you will then memcpy those 0s into the settings

its a new emplacement

schombert: yes ... I am aware

schombert: oh, I see what you are doing

schombert: In any case, this won't fix the problem. The problem is the line `simple_fs::write_file(settings_location, NATIVE("user_settings.dat"), &buffer[0], **uint32_t(sizeof(buffer)));`

schombert: The issue here is that you are writing the members item by item the exact number of bytes they need. However, the size of the structure is larger than the total size of the items. Thus, when the file is written, some additional junk data is added at the end. You are now reading back that junk data, which is always going to produce invalid values regardless of what you do in the reading section

PR #627, opened 2023-11-04
Misc. improvements to UI

wxwisiasdf:


src/gui/gui_effect_tooltips.cpp
```diff
@@ -2027,16 +2027,11 @@ uint32_t ef_change_tag_culture(EFFECT_DISPLAY_PARAMS) {
 	{
 		auto box = text::open_layout_box(layout, indentation);
 		text::substitution_map m;
-		if(primary_slot != -1) {
```
schombert, line 2030: these checks are to handle the case where the description of an effect needs to be displayed in the abstract (i.e. not applied to any particular nation, but what the effect is in general). That is done by passing -1 in the slots to indicate that the effect needs to be displayed for a generic target

src/gui/gui_effect_tooltips.cpp
```diff
@@ -2056,16 +2051,11 @@ uint32_t ef_change_tag_no_core_switch_culture(EFFECT_DISPLAY_PARAMS) {
...
```
schombert, line 2059: as above

PR #629, opened 2023-11-04
Artillery now is part of scenario serialization + report when artillery cant be found so we dont run into weird shit

wxwisiasdf:


src/gamestate/system_state.cpp
```diff
@@ -2295,6 +2295,9 @@ void state::load_scenario_data(parsers::error_handler& err) {
 		if(!bool(military_definitions.irregular)) {
 			err.accumulated_errors += "No irregular (or equivalent unit type) found\n";
 		}
+		if(!bool(military_definitions.artillery)) {
+			err.accumulated_errors += "No artillery (or equivalent unit type) found\n";
```
schombert, line 2299: this will always trigger, I expect, because artillery isn't identified as part of loading the unit files. It is identified as part of the load unsaved data routines

schombert:
make sure you test this some ... last time we went down this road we ended up breaking the ai's building in some mods

wxwisiasdf:
I mean it is better than previously done, because before, if arty was not defined, it will go silently without us knowing, causing unspeakable errors

PR #630, opened 2023-11-04
Macro builder version one point oh

wxwisiasdf:


schombert:
I see an issue here: your file can have at most 256 templates in it, ever. Which may sound like more than enough, but every version of every mod  launched under every different version of alice will need will need its own set of templates. I believe that it will run out of space rather quickly ... and then what?

wxwisiasdf:
> I see an issue here: your file can have at most 256 templates in it, ever. Which may sound like more than enough, but every version of every mod launched under every different version of alice will need will need its own set of templates. I believe that it will run out of space rather quickly ... and then what?

I will expand it to 8192, does that sound kosher?

schombert:
That will make the problem less likely, for certain, but why not just fix it properly? Let every scenario file have its own template save file. This will allow those old template files to be cleaned up with the old scenario files when they are no longer needed, rather than hanging around forever.

wxwisiasdf:
> That will make the problem less likely, for certain, but why not just fix it properly? Let every scenario file have its own template save file. This will allow those old template files to be cleaned up with the old scenario files when they are no longer needed, rather than hanging around forever.

isnt it better to just have every template be in a single file?

schombert:
Sure, that could work too, although opening a file is somewhat time consuming, and it would be less than ideal to have to open all the template files for all the mods to figure out which one was usable.

PR #632, opened 2023-11-04
Better trigger NOT/YES icons (and smaller)

wxwisiasdf:


src/ogl/opengl_wrapper.cpp
```diff
@@ -115,7 +115,7 @@ void load_special_icons(sys::state& state) {
 		auto content = simple_fs::view_contents(*money_dds);
 		uint32_t size_x, size_y;
 		state.open_gl.money_icon_tex = GLuint(ogl::SOIL_direct_load_DDS_from_memory(reinterpret_cast<uint8_t const*>(content.data),
-				content.file_size, size_x, size_y, 0));
+				content.file_size, size_x, size_y, ogl::SOIL_FLAG_TEXTURE_REPEATS));
```
schombert, line 118: Why are you changing these textures to repeat? You surely want them to only appear once when rendered, not in a tiled pattern

wxwisiasdf: Its not that we want them repeated, is that telling this to SOIL allows me to use compressed textures - for some reason I cannot yet fathom

PR #633, opened 2023-11-04
Compress assets

wxwisiasdf:


wxwisiasdf:
note: didnt make assets into DDS if it increased their size

PR #634, opened 2023-11-04
Diplomacy buttons akin to OpenV2

wxwisiasdf:


schombert:
could you describe what this change is actually doing?

PR #638, opened 2023-11-05
reduce file size of OpenV2 map mod so my internet doesnt die

wxwisiasdf:


schombert:
the only things this seem to do is (a) delete the zip, which I was using as a way of hosting the mod (see the mod compat. channel) and (b) move things around breaking any links to tem

PR #640, opened 2023-11-05
Fix decisions not triggering for countries

wxwisiasdf:


schombert:
I don't understand how this will change anything

wxwisiasdf:
remove some variable shadowing which caused some issues

schombert:
Removing the shadowing is a good practice, but it wasn't doing anything in this situation. Whatever the bug is, this isn't the cause

wxwisiasdf:
ok but its good practice

schombert:
find and fix the real bug, and I will merge away

src/parsing/parsers_declarations.cpp
```diff
@@ -2168,18 +2168,40 @@ void rebel_body::demands_enforced_effect(dcon::effect_key value, error_handler&
...
+	}
+}
+
+void decision::finish(decision_context& context) {
+	// set a default "always = yes" if no allow is given
+	if(!context.outer_context.state.world.decision_get_allow(context.id)) {
+		std::vector<uint16_t> data;
+		data.push_back(uint16_t(trigger::always | trigger::no_payload | trigger::association_eq));
+		auto key = context.outer_context.state.commit_trigger_data(data);
+		context.outer_context.state.world.decision_set_allow(context.id, key);
```
schombert, line 2203: this will bloat the scenario files unnecessarily, Instead, find where allow is checked (once in commands for the player, and once in the ai code) and make allow being the invalid handle == to yes

src/ai/ai.cpp
```diff
@@ -667,12 +667,13 @@ void take_ai_decisions(sys::state& state) {
 
 		ve::execute_serial_fast<dcon::nation_id>(state.world.nation_size(), [&](auto ids) {
 			ve::vbitfield_type filter_a = potential
-				? (ve::compress_mask(trigger::evaluate(state, potential, trigger::to_generic(ids), trigger::to_generic(ids), 0)) & !state.world.nation_get_is_player_controlled(ids))
+				? (ve::compress_mask((!potential || trigger::evaluate(state, potential, trigger::to_generic(ids), trigger::to_generic(ids), 0))) & !state.world.nation_get_is_player_controlled(ids))
```
schombert, line 670: this is redundant because potential was already checked via the line above

src/ai/ai.cpp
```diff
@@ -667,12 +667,13 @@ void take_ai_decisions(sys::state& state) {
...
 				: !state.world.nation_get_is_player_controlled(ids);
 
 			if(filter_a.v != 0) {
+				// empty allow assumed to be an "always = yes"
 				ve::mask_vector filter_c = allow
-					? (trigger::evaluate(state, allow, trigger::to_generic(ids), trigger::to_generic(ids), 0) && (state.world.nation_get_owned_province_count(ids) != 0)) && filter_a
+					? ((!allow || trigger::evaluate(state, allow, trigger::to_generic(ids), trigger::to_generic(ids), 0)) && (state.world.nation_get_owned_province_count(ids) != 0)) && filter_a
```
schombert, line 676: same thing here

schombert: the real problem is the second validity check, below

schombert: and then don't forget to make sure that the player is also allowed to take the decision in the command code, if you haven't already

PR #641, opened 2023-11-05
Include a fallback for DPI on older systems

wxwisiasdf:
ignore branch name 😅


src/entry_point_win.cpp
```diff
@@ -113,7 +113,13 @@ int WINAPI wWinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPWSTR
 		signal(SIGABRT, signal_abort_handler);
 	}
 
-	SetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
+	// Workaround for old machines
+	HINSTANCE hUser32dll = LoadLibrary(L"User32.dll");
+	if(hUser32dll) {
+		auto pSetProcessDpiAwarenessContext = (decltype(&SetProcessDpiAwarenessContext))GetProcAddress(hUser32dll, "SetProcessDpiAwarenessContext");
+		pSetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
```
schombert, line 120: this isn't going to fix anything. On old windows, the dll won't contain the function (and even if it did, wouldn't understand the constant value). Nor can you simply omit it without things behaving badly. You would need to fall back to an older version of dpi awareness and rewrite the per-monitor sizing routine for that instead.

src/entry_point_win.cpp
```diff
@@ -113,7 +114,29 @@ int WINAPI wWinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPWSTR
...
+	HINSTANCE hUser32dll = LoadLibrary(L"User32.dll");
+	if(hUser32dll) {
+		auto pSetProcessDpiAwarenessContext = (decltype(&SetProcessDpiAwarenessContext))GetProcAddress(hUser32dll, "SetProcessDpiAwarenessContext");
+		if(pSetProcessDpiAwarenessContext != NULL) {
+			pSetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
+		} else {
+			// windows 8.1 (not present on windows 8 and only available on desktop apps)
+			HINSTANCE hShcoredll = LoadLibrary(L"Shcore.dll");
+			if(hShcoredll) {
+				auto pSetProcessDpiAwareness = (decltype(&SetProcessDpiAwareness))GetProcAddress(hShcoredll, "SetProcessDpiAwareness");
```
schombert, line 127: We need someone to test whether these alternate paths actually work. The issue is that we draw over the whole window. Previously before PER_MONITOR_AWARE_V2 we might have had to call https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-enablenonclientdpiscaling on windows 10. So I have no idea whether this will actually work or not on window 8 as we expect

schombert:
My point was that EnableNonClientDpiScaling was used on older versions of windows 10. I don't think it is available on 8. So, what happens on 8? I don't know, and it needs a machine with windows 8 to test

schombert:
I have thought about it and concluded that it would be foolish to merge something that I can't test or bugfix

PR #645, opened 2023-11-06
Visual state selection for wargoals

wxwisiasdf:
image
image
image
image


assets/alice.csv
```diff
@@ -1260,6 +1260,7 @@ w_rgo_prod;Global RGO production: $x$;;;;;;;;;;;;x
 w_artisan_prod;Global artisan production: $x$;;;;;;;;;;;;x
 w_fac_prod;Global factory production: $x$;;;;;;;;;;;;x
 alice_lose_gp;We are losing �YGreat Power�W status!;;;;;;;;;;;;x
+alice_state_select_title;Select states (ESC);;;Seleccionar estados (ESC);;;;;;;;;x
```
schombert, line 1263: shouldn't that be (ESC to cancel)?

src/common_types/constants.hpp
```diff
@@ -499,7 +499,8 @@ constexpr inline uint8_t standard_pause = log | sound | popup | pause;
 enum class game_mode_type {
 	pick_nation,
 	in_game,
-	end_screen
+	end_screen,
+	select_states
```
schombert, line 503: does this interfere with time advancing? (in single player, not a big issue, but could create problems for multiplayer)

wxwisiasdf: fixed

PR #646, opened 2023-11-06
Add crash handler to launcher

wxwisiasdf:


schombert:
I will merge this to make you happy, but I am probably not going to save the launcher pdb from releases or look at any crash dumps. So far the crash dumps we generate from alice have been unhelpful.

PR #653, opened 2023-11-06
sanity check navies & armies

wxwisiasdf:


src/gamestate/system_state.cpp
```diff
@@ -2978,6 +2978,26 @@ void state::load_scenario_data(parsers::error_handler& err) {
 	if(gov_error)
 		return;
 
+	// Sanity checking navies & armies
+	for(auto n : world.in_navy) {
+		auto p = n.get_navy_location().get_location();
```
schombert, line 2983: while you are there, why not loop over the ships in the navy and assert that they all have a valid unit type (and for regiments below)

src/gamestate/system_state.cpp
```diff
@@ -2978,6 +2978,26 @@ void state::load_scenario_data(parsers::error_handler& err) {
...
+		if(p.id.index() >= province_definitions.first_sea_province.index()) {
+			//...
+		} else { //land province
+			auto pp = world.province_get_port_to(p);
+			auto adj = world.get_province_adjacency_by_province_pair(p, pp);
+			if(!pp || !adj) {
```
schombert, line 2989: checking for port province *should* be sufficient ...

wxwisiasdf: you never know

PR #654, opened 2023-11-06
improve error reporting

wxwisiasdf:
- show filename where invalid unit type was found for unit
- oob relationships within range

### DIFF
--- a/src/parsing/military_parsing.cpp
+++ b/src/parsing/military_parsing.cpp
@@ -162,12 +162,30 @@ void make_oob_army(token_generator& gen, error_handler& err, oob_file_context& c
 	context.outer_context.state.world.force_create_army_control(id, context.nation_for);
 	oob_file_army_context new_context{context.outer_context, id, context.nation_for};
 	parse_oob_army(gen, err, new_context);
+
+	// and check they have correct unit types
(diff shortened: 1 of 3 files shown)


src/parsing/military_parsing.cpp
```diff
@@ -162,12 +162,30 @@ void make_oob_army(token_generator& gen, error_handler& err, oob_file_context& c
...
 void make_oob_navy(token_generator& gen, error_handler& err, oob_file_context& context) {
 	auto id = context.outer_context.state.world.create_navy();
 	context.outer_context.state.world.force_create_navy_control(id, context.nation_for);
 	oob_file_navy_context new_context{context.outer_context, id, context.nation_for};
 	parse_oob_navy(gen, err, new_context);
+
+	// and check they have correct unit types
+	auto p = context.outer_context.state.world.navy_get_location_from_navy_location(id);
+	for(auto m : context.outer_context.state.world.navy_get_navy_membership(id)) {
+		if(!bool(m.get_regiment().get_type())) {
```
schombert, line 184: i think you mean get ship

wxwisiasdf: whoopsie

PR #658, opened 2023-11-07
Force fow on multiplayer

wxwisiasdf:


src/gamestate/system_state.cpp
```diff
@@ -1776,6 +1776,9 @@ void state::load_user_settings() {
 		user_settings.master_volume = std::clamp(user_settings.master_volume, 0.0f, 1.0f);
 		if(user_settings.antialias_level > 16)
 			user_settings.antialias_level = 0;
+		/* Always force fog of war on multiplayer - even tho it's an honour system */
+		if(network_mode != sys::network_mode_type::single_player)
```
schombert, line 1780: this is not a good solution because if the user then changes a setting, fow will be saved as on (as on settings change, all current settings are saved), Instead, you should find the "is fow on" check(s) and turn them into "is fow on or in multiplayer"

PR #667, opened 2023-11-10
reduce openv2 mod filesize

wxwisiasdf:


schombert:
rather than this, if you really want to save 10s of MB, find a way to make loading the 3 map coloring dds files from pngs first. Then we could just include pngs, which would be much smaller

PR #668, opened 2023-11-11
Update multiplayer.md

wxwisiasdf:


docs/multiplayer.md
```diff
@@ -44,7 +44,35 @@ On debug builds, a checksum will be generated every tick to ensure synchronisati
 
 Otherwise, the goal is no more oos :D
 
-### Tutorial (Client)
+### User guide
+
+In the launcher you may be presented with the following:
+
+![image](https://github.com/schombert/Project-Alice/assets/39974089/004fa194-487b-43d5-b1eb-e1185388476e)
```
schombert, line 51: don't add imaged to the assets directory (assets is what we copy into every install) -- put them in a folder inside docs instead

wxwisiasdf: thats github assets, not the actual asset folder

schombert: Ok, but why aren't the images just in the repo normally, so that we can update them, etc?

PR #670, opened 2023-11-12
properly zero out padding on various structures

wxwisiasdf:


src/military/military.hpp
```diff
@@ -131,13 +152,35 @@ struct global_military_state {
 
 	bool pending_blackflag_update = false;
 };
+static_assert(sizeof(global_military_state) ==
```
schombert, line 155: this is unnecessary -- global military state doesn't get memcopied (it can't; it has a vector member)

PR #675, opened 2023-11-13
Reimplement rebel armies

wxwisiasdf:


src/military/military.cpp
```diff
@@ -6437,6 +6431,29 @@ void update_siege_progress(sys::state& state) {
 			// is controler != owner ...
 			// event::fire_fixed_event(state, );
 		}
+		if(auto nr = new_rebel_controller.get(prov); nr) {
+			province::set_province_controller(state, prov, nr);
+			eject_ships(state, prov);
+
+			for(auto ar : state.world.province_get_army_location(prov)) {
+				auto a = ar.get_army();
```
schombert, line 6439: so, this is going to loop over all the armies in the province that has just been sieged ... presumably there are only rebel armies there, so none of them will be rebel hunters, which are exclusively player nation controlled...

schombert:
Additionally, the automatic rebel hunters will need logic to ensure that they don't suicidally throw themselves against rebel stacks that are too big.

src/culture/rebels.cpp
```diff
@@ -771,9 +771,90 @@ void daily_update_rebel_organization(sys::state& state) {
...
+						state.world.army_set_dig_in(a, 0);
+
+						rebel_hunters[i] = rebel_hunters.back();
+						rebel_hunters.pop_back();
+						break;
+					}
+				}
+
+				rebel_provs[0] = rebel_provs.back();
+				rebel_provs.pop_back();
```
schombert, line 855: I see some potential issues with this logic. First, while it won't break if pathfinding fails, it may leave some provinces unable to be targeted by rebel hunters. (For example, if the rebel province is nearest to a rebel hunter who is overseas from it, that rebel hunter will not be able to path to it, but the province will still be removed from the list. This could result in provinces that are simply never take back). You should at least consider using the connected region values to filter this check with. Secondly, it doesn't take the size of the rebel stacks into account, and so may result in a suicidal use of rebel hunters, especially if they have become depleted by prior combats. Thirdly, it doesn't prioritize rebel armies over rebel occupations, meaning that you could end up with a rebel hunter forever trailing behind rebel armies, recapturing one province while the rebels capture a new one.

schombert:
I also note that the "return home" logic will function inconsistently. If a rebel hunter finishes a battle and ends up in an unoccupied province without any more movement left to do, the return home logic won't be triggered and it won't go back to its starting province.

PR #676, opened 2023-11-13
increment army size (zombie mod, related to rebels)

wxwisiasdf:
allows me to actually play the game for a while after 1844,


schombert:
adding rebels back shouldn't have caused army size to need that much expansion. You aren't making each rebel regiment its own army are you?

wxwisiasdf:
> adding rebels back shouldn't have caused army size to need that much expansion. You aren't making each rebel regiment its own army are you?

of course not

PR #687, opened 2023-11-19
PDXScript modding extensions

wxwisiasdf:
- Compactly specify POPs in CSV
- change_variable/add_variable/sub_variable, etc


schombert:
base game also supports assigning rebel type, and maybe we should support literacy? In any case, I will merge this when a mod expresses a desire to use it as their primary means of defining the pops

PR #692, opened 2023-11-19
show end screen when player is defeated

wxwisiasdf:


src/nations/nations.cpp
```diff
@@ -1309,8 +1309,10 @@ void cleanup_nation(sys::state& state, dcon::nation_id n) {
 	state.national_definitions.gc_pending = true;
 	state.diplomatic_cached_values_out_of_date = true; // refresh stored counts of allies, vassals, etc
 
-	if(n == state.local_player_nation) { // TODO: player defeated; notify and end game
+	if(n == state.local_player_nation) {
+		// Player was defeated, show end screen
 		state.local_player_nation = dcon::nation_id{};
+		state.mode = sys::game_mode_type::end_screen;
```
schombert, line 1315: does this break multiplayer?

wxwisiasdf: once a player is on the end screen, the game stops receiving and sending network commands, definitely halting the game basically

wxwisiasdf: once a player is on the end screen, the game stops receiving and sending network commands, definitely halting the game basically

schombert: but does this break the game for everyone else who isn't dead? Will it create an issue if you eventually implement better connectivity detection / lagging player detection?

wxwisiasdf: it should not, because the player essentially stops "responding", eventually disconnecting when they press the "exit" button

schombert:
Well, multiplayer is "not my problem" but I do think this may eventually cause you issues when you do something better to handle players who are running slow, etc

PR #697, opened 2023-11-20
Fix military battles leading to invalid regiments being accounted for

wxwisiasdf:


schombert:
the second removal logic in change province ownership is probably necessary -- although you must *also* remove the regiment from its slot in the reserves for the battle if it is there. The first, however, should not be. Regiments take damage from only two sources: attrition and combat. Attrition damage is not applied to regiments in combat. And when a regiment is reduced to 0 strength in combat, it is removed from the battle at that point. Feel free to double check that my reasoning matches what is actually happening, but if I am correct then regiments eliminated from damage should be handled that way, and hence should not require this extra logic to remove them from battles.

src/provinces/province.cpp
```diff
@@ -863,6 +863,19 @@ void change_province_owner(sys::state& state, dcon::province_id id, dcon::nation
...
+			for(auto& e : state.world.land_battle_get_attacker_back_line(b))
+				e = (e == r) ? dcon::regiment_id{} : e;
+			for(auto& e : state.world.land_battle_get_attacker_front_line(b))
+				e = (e == r) ? dcon::regiment_id{} : e;
+			for(auto& e : state.world.land_battle_get_defender_back_line(b))
+				e = (e == r) ? dcon::regiment_id{} : e;
+			for(auto& e : state.world.land_battle_get_defender_front_line(b))
+				e = (e == r) ? dcon::regiment_id{} : e;
+			for(auto& e : state.world.land_battle_get_reserves(b))
+				e.regiment = (e.regiment == r) ? dcon::regiment_id{} : e.regiment;
```
schombert, line 877: you should probably just remove it from the list of reserves completely (unless you are sure that the logic for pulling them in checks for invalid ids) -- iterate from the last index in the reserves down and do the swap and pop idiom to remove

PR #703, opened 2023-11-21
prevent crash from ocurring (yes it occurs if i dont do this)

wxwisiasdf:


schombert:
You should figure out where the assumption about sources of damage I described to you is incorrect. Fixing the symptom without understanding the cause is generally a bad idea

PR #705, opened 2023-11-26
military battle logic fixes

wxwisiasdf:


src/military/military.cpp
```diff
@@ -3961,46 +3968,24 @@ void army_arrives_in_province(sys::state& state, dcon::army_id a, dcon::province
...
-				add_army_to_battle(state, a, new_battle, !bool(owner_nation) ? war_role::attacker : war_role::defender);
-				add_army_to_battle(state, o.get_army(), new_battle, bool(owner_nation) ? war_role::attacker : war_role::defender);
-
-				gather_to_battle = new_battle.id;
-				break;
-			} else if(auto par = internal_find_war_between(state, owner_nation, other_nation); par.role != war_role::none) {
+			military::participation par{ dcon::war_id{}, war_role::none };
+			if(owner_nation && other_nation) { // nation vs. nation
+				par = internal_find_war_between(state, owner_nation, other_nation);
+			} else { // nation vs. rebels
```
schombert, line 3984: this would result in rebels fighting themselves and each other

src/culture/rebels.cpp
```diff
@@ -878,45 +877,36 @@ void rebel_risings_check(sys::state& state) {
...
 					auto used_count = int32_t(cregs.end() - cregs.begin());
-
 					if(used_count < max_count) {
 						auto pop_location = pop.get_pop().get_province_from_pop_location();
-
-						auto new_reg = military::create_new_regiment(state, dcon::nation_id{}, state.military_definitions.irregular);
 						auto a = [&]() {
 							for(auto ar : state.world.province_get_army_location(pop_location)) {
-								if(!(ar.get_army().get_battle_from_army_battle_participation()) && ar.get_army().get_controller_from_army_rebel_control() == rf)
+								if(ar.get_army().get_controller_from_army_rebel_control() == rf && !ar.get_army().get_black_flag() && !ar.get_army().get_is_retreating() && !ar.get_army().get_battle_from_army_battle_participation())
```
schombert, line 898: rebels armies can't be black flagged. And it is fine to insert a new regiment into a retreating army (cheaper than creating a new army, after all)

src/culture/rebels.cpp
```diff
@@ -878,45 +877,36 @@ void rebel_risings_check(sys::state& state) {
...
 									return ar.get_army().id;
 							}
-							auto new_army = fatten(state.world, state.world.create_army());
+							auto new_army = dcon::fatten(state.world, state.world.create_army());
```
schombert, line 901: you don't need `dcon::` here because of ADL

wxwisiasdf: readability, not adl

src/military/military.cpp
```diff
@@ -3940,7 +3947,7 @@ void army_arrives_in_province(sys::state& state, dcon::army_id a, dcon::province
 
 	state.world.army_set_location_from_army_location(a, p);
 	auto regs = state.world.army_get_army_membership(a);
-	if(!state.world.army_get_black_flag(a) && !state.world.army_get_is_retreating(a) && regs.begin() != regs.end()) {
+	if(army_can_join_land_battle(state, a) && regs.begin() != regs.end()) {
```
schombert, line 3950: note that this will be slower: you are now checking for more things, one of which should have been eliminated by the assert (and the other shouldn't be relevant because this function isn't called for armies being transported -- unless it is called while they are moving off the boats but before that relationship has been broken, in which case this change will be a bug)

src/military/military.cpp
```diff
@@ -3961,46 +3968,24 @@ void army_arrives_in_province(sys::state& state, dcon::army_id a, dcon::province
 		}
 
 		// start battle
-		dcon::land_battle_id gather_to_battle;
-		dcon::war_id battle_in_war;
-
+		auto gather_to_battle = dcon::land_battle_id{};
+		auto battle_in_war = dcon::war_id{};
```
schombert, line 3972: This is fine, but it is redundant because their default constructor already does this

src/military/military.cpp
```diff
@@ -3961,46 +3968,24 @@ void army_arrives_in_province(sys::state& state, dcon::army_id a, dcon::province
...
 		for(auto o : state.world.province_get_army_location(p)) {
 			if(o.get_army() == a)
 				continue;
-			if(o.get_army().get_is_retreating() || o.get_army().get_black_flag() || o.get_army().get_navy_from_army_transport() || o.get_army().get_battle_from_army_battle_participation())
+			if(!army_can_join_land_battle(state, o.get_army()))
 				continue;
 
 			auto other_nation = o.get_army().get_controller_from_army_control();
 
-			if(bool(owner_nation) != bool(other_nation)) { // battle vs. rebels
```
schombert, line 3975: This logic you have removed was not redundant. It works slightly differently for rebels

wxwisiasdf:
ok reverted military.cpp

PR #711, opened 2023-11-27
take care of vassals and substates

wxwisiasdf:
prevents prussia from losing their substates to rebels [in TGC]
and china too

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -4145,9 +4145,26 @@ void assign_targets(sys::state& state, dcon::nation_id n) {
 	if(ready_armies.empty())
 		return; // nothing to attack with
 
+	struct army_target {
+		float minimal_distance;
+		dcon::province_id location;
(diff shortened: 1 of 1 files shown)


src/ai/ai.cpp
```diff
@@ -4191,6 +4189,23 @@ void assign_targets(sys::state& state, dcon::nation_id n) {
 			}
 		}
 	}
+	/* Our allies (mainly our substates, vassals) - we need to care of them! */
+	std::vector<dcon::nation_id> allied_with;
+	allied_with.reserve(state.world.nation_size());
+	for(const auto ovr : state.world.nation_get_overlord_as_ruler(n))
+		allied_with.push_back(ovr.get_subject());
+	for(auto w : allied_with) {
```
schombert, line 4197: rather than iterating to fill the vector and then iterating over its contents, you could merge these two loops and get rid of the vector

PR #712, opened 2023-11-27
compiling source files text.cpp font.cpp and parsers.cpp seperately and moving template definition to header

yrrrmmbl:


schombert:
This way of doing things is not an accident, it is a technique called a unity build or a jumbo build https://en.wikipedia.org/wiki/Unity_build

schombert:
This is done because our headers are quite large. Our main generated header is over 100,000 lines at this point.

yrrrmmbl:
i am fine with keeping it this way but it takes a very long time to build on my machine and is constantly swapping stuff to disk.
it also takes forever to recompile when i change just a little bit of one source file since it goes and rebuilds effectively all source files. the most frustrating part of it is that only one cpu core gets pushed not even to the limit so it seems like nothing at all is happening.

as it says on the wikipedia page:
"One of the drawbacks of unity builds is a larger memory footprint due to larger translation units. Larger translation units can also negatively affect parallel builds, since a small number of large compile jobs is generally harder or impossible to schedule to saturate all available parallel computing resources effectively. Unity builds can also deny part of the benefits of incremental builds, that rely on rebuilding as little code as possible, i.e. only the translation units affected by changes since the last build."

schombert:
That is true. Currently what we are trying to optimize for is the time of a full rebuild, which is what takes the most time. While building things as they stand takes me perhaps a minute or so in debug, compiling in release takes well over half an hour, which I have to do twice for a PGO build (and soon, 3 or 4 times to support more instruction sets). If you want to change things for the purposes of debug builds, I would suggest making a new build target that doesn't include the main.cpp file and which instead builds things more normally (which you could approach incrementally). This would give you the performance you want for incremental builds without interfering with the process for the release build.

PR #715, opened 2023-12-02
more useful warnings and errors

wxwisiasdf:


src/parsing/effect_parsing.hpp
```diff
@@ -2973,7 +2977,7 @@ struct effect_body {
 			else if(context.from_slot == trigger::slot_contents::province)
 				context.compiled_effect.push_back(effect::remove_casus_belli_from_province);
 			else {
-				err.accumulated_errors += "remove_casus_belli = from effect used in an incorrect scope type (" + err.file_name +
+				err.accumulated_errors += "remove_casus_belli = from effect used in an incorrect scope type " + slot_contents_to_string(context.main_slot) + " (" + err.file_name +
```
schombert, line 2980: so, I will merge this pr, because it probably won't hurt anything, but note that some of these errors, such as this one in particular, can come from mismatches in scopes other than the main one, in which case displaying the main one, which may very well make sense, is probably going to be even more confusing

PR #718, opened 2023-12-03
merge map with main.cpp

wxwisiasdf:


schombert:
that would make erik sad

wxwisiasdf:
map is already worked enough on + its better since were getting closer to 1.0

PR #719, opened 2023-12-03
conservative ideology fallback with name

wxwisiasdf:


src/gamestate/system_state.cpp
```diff
@@ -2725,8 +2725,14 @@ void state::load_scenario_data(parsers::error_handler& err) {
 		}
 	}
 	if(!culture_definitions.conservative) {
-		err.accumulated_errors += "NO CONSERVATIVE IDEOLOGY (fatal error)\n";
-		err.fatal = true;
+		if(auto it = context.map_of_ideologies.find("conservative"); it != context.map_of_ideologies.end()) {
+			culture_definitions.conservative = it->second.id;
+			err.accumulated_warnings += "conservative ideology will not reduce militancy\n";
```
schombert, line 2730: that's ... not true

wxwisiasdf: there

PR #721, opened 2023-12-04
fix end screen displaying rebel flag instead of player

wxwisiasdf:


schombert:
why are you adding random exes?

schombert:
and why do we need all these variations on the reb flag?

PR #729, opened 2023-12-05
Add if and else_if

wxwisiasdf:


schombert:
since you are adding codepoints, you need to bump the serialization version

src/scripting/effects.cpp
```diff
@@ -4,8 +4,14 @@
...
 
+struct effect_local_state {
+	static constexpr inline uint32_t max_if_depth = 16;
+	bool if_is_triggered[max_if_depth] = { false };
+	uint32_t if_depth = 0;
+};
+
 #define EFFECT_PARAMTERS                                                                                                         \
-	uint16_t const *tval, sys::state &ws, int32_t primary_slot, int32_t this_slot, int32_t from_slot, uint32_t r_hi, uint32_t r_lo
+	uint16_t const *tval, sys::state &ws, int32_t primary_slot, int32_t this_slot, int32_t from_slot, uint32_t r_hi, uint32_t r_lo, effect_local_state els
```
schombert, line 14: Shouldn't this be passed by reference?

wxwisiasdf: oversight, whoops

src/scripting/effects.cpp
```diff
@@ -15,14 +15,44 @@ inline uint32_t apply_subeffects(EFFECT_PARAMTERS) {
...
+}
+uint32_t es_else_if_scope(EFFECT_PARAMTERS) {
+	uint32_t ret = 0;
+	if((tval[0] & effect::scope_has_limit) != 0) {
+		auto limit = trigger::payload(tval[2]).tr_id;
+		if(!els && trigger::evaluate(ws, limit, primary_slot, this_slot, from_slot)) {
+			ret = apply_subeffects(tval, ws, primary_slot, this_slot, from_slot, r_hi, r_lo, els);
+			els = true;
+		}
+	} else {
```
schombert, line 51: needs to test for !els here, otherwise else will always run if there is no limit in it

PR #730, opened 2023-12-05
repeat backspace, del and left, right, up and down arrow keypress events

yrrrmmbl:
this makes the listed keypress events repeated events which is needed to have conventional text editing


schombert:
a) what about linux?
b) has this been tested not to break the map? Holding left, for example, would repeatedly send "left is now down" message to the map, which I imagine it wasn't designed to handle

yrrrmmbl:
a) not sure, actually
b) map_state::on_key_down() and ::on_key_up work by setting booleans of which keys are depressed and the actual doing things happens in the update function and relies on those booleans

PR #731, opened 2023-12-06
Adjusts the consciousness algorithm to be consistent with Vic 2

ma44:
Noticed that certain European nations were getting maxed plurality just a single decade into the game, UI and calculations seem to make consciousness gain from plurality/literacy be 10x as effective as what the UI states in Vic 2 so I adjusted it.

New adjustments
image

Vs Vic 2
image


wxwisiasdf:
do not write it as `10`, write it as `10.0f`

PR #736, opened 2023-12-08
tooltip swallowing finetuning

yrrrmmbl:


schombert:
there should be code somewhere you can use to make the mouse check the transparency of the background pixels

schombert:
Maybe I am thinking of the code from openv2. In any case, you can look at `void image_element_base::render(sys::state& state, int32_t x, int32_t y) noexcept` to see how to get the texture id. Once you have that, you can use it via `state.open_gl.asset_textures[id].data` to get the raw texture data (assuming it is marked in the gfx files as partial transparent, which I think it is). From there, you just need to test whether the pixel under the mouse is sufficiently transparent

PR #744, opened 2023-12-08
add divider to fix colour for event immediate

wxwisiasdf:


schombert:
what problem are you trying to fix with this?

schombert:
In V2 the description of the effect goes after the main event description, by the way

wxwisiasdf:
> what problem are you trying to fix with this?

colour being fucked because it is set to white and somehow this causes the immediate effect to become white

schombert:
And does it actually fix that? Because I think the tooltip display forces the text to be white

PR #745, opened 2023-12-08
properly display if/else_if on tooltips for effects

wxwisiasdf:


assets/alice.csv
```diff
@@ -1111,6 +1111,8 @@ et_add_selector;enable province selector;;;habilita el selector provincial;;;;;;
 et_remove_selector;disable province selector;;;deshabilita el selector provincial;;;;;;;;;x
 et_add_immigrator;enable province immigrator;;;habilita el inmigrador provincial;;;;;;;;;x
 et_remove_immigrator;disable province immigrator;;;deshabilita el inmigrador provincial;;;;;;;;;x
+et_else;If;;;Si;;;;;;;;;x
+et_else_if;Else if;;;O si;;;;;;;;;x
```
schombert, line 1115: what if it is an else scope with no limit? Then it will read "else if increase prestige by 5" or something like that. I suggest just "else", which will make when it does have a limit read as "else where .... increase prestige by 5", which still isn't great, but I guess it is a bit better

assets/alice.csv
```diff
@@ -1111,6 +1111,8 @@ et_add_selector;enable province selector;;;habilita el selector provincial;;;;;;
 et_remove_selector;disable province selector;;;deshabilita el selector provincial;;;;;;;;;x
 et_add_immigrator;enable province immigrator;;;habilita el inmigrador provincial;;;;;;;;;x
 et_remove_immigrator;disable province immigrator;;;deshabilita el inmigrador provincial;;;;;;;;;x
+et_else;If;;;Si;;;;;;;;;x
```
schombert, line 1114: I think you mean "et_if" as the key here

PR #748, opened 2023-12-08
fix rebels being on a restricted area (for ex. when hungary defects f…

wxwisiasdf:
…rom austria)

### DIFF
--- a/src/culture/rebels.cpp
+++ b/src/culture/rebels.cpp
@@ -1256,10 +1256,9 @@ std::string rebel_name(sys::state& state, dcon::rebel_faction_id reb) {
 	return text::resolve_string_substitution(state, state.world.rebel_faction_get_type(reb).get_name(), sub);
 }
 
-bool allow_in_area(sys::state& state, dcon::province_id p, dcon::province_id l, dcon::rebel_faction_id reb) {
+bool allow_in_area(sys::state& state, dcon::province_id p, dcon::rebel_faction_id reb) {
 	auto rf = dcon::fatten(state.world, reb);
 	auto prov = dcon::fatten(state.world, p);
-	auto location = dcon::fatten(state.world, l);
 	switch(culture::rebel_area(rf.get_type().get_area())) {
 	case culture::rebel_area::all:
 		return true;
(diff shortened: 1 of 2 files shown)


src/culture/rebels.cpp
```diff
@@ -1270,11 +1269,11 @@ bool allow_in_area(sys::state& state, dcon::province_id p, dcon::province_id l,
 	case culture::rebel_area::religion:
 		return prov.get_dominant_religion() == rf.get_religion();
 	case culture::rebel_area::nation_religion:
-		return prov.get_dominant_religion() == rf.get_defection_target().get_religion() && prov.get_province_ownership().get_nation() == location.get_province_ownership().get_nation();
+		return prov.get_dominant_religion() == rf.get_defection_target().get_religion() && prov.get_province_ownership().get_nation() == rf.get_rebellion_within().get_ruler();
 	case culture::rebel_area::nation_culture:
-		return prov.get_dominant_culture() == rf.get_defection_target().get_primary_culture() && prov.get_province_ownership().get_nation() == location.get_province_ownership().get_nation();
+		return prov.get_dominant_culture() == rf.get_defection_target().get_primary_culture() && prov.get_province_ownership().get_nation() == rf.get_rebellion_within().get_ruler();
 	case culture::rebel_area::nation:
-		return prov.get_province_ownership().get_nation() == location.get_province_ownership().get_nation();
```
schombert, line 1277: shouldn't this have prevented rebels from leaving the nation already?

schombert: maybe they were crossing borders because they were area all?

src/culture/rebels.cpp
```diff
@@ -1301,8 +1300,41 @@ void update_armies(sys::state& state) {
 		auto area = arc.get_controller().get_type().get_area();
 		auto location = ar.get_location_from_army_location();
 		/* If on an unsieged province, siege it! */
-		if(location.get_nation_from_province_control() && !location.get_rebel_faction_from_province_rebel_control())
+		if(location.get_nation_from_province_control() && !location.get_rebel_faction_from_province_rebel_control()) {
+			/* On a restricted area, so move the fuck out of it */
+			if(!allow_in_area(state, location, arc.get_controller())) {
```
schombert, line 1305: this doesn't check if the rebels are already moving, which would then cause them to try to move every time they are updated, which could theoretically cause them to get stuck in the province. And if they are way inside a forbidden area, they could end up just bouncing back and forth. I suggest just omitting this section altogether and accept that if the rebel has been forced into an invalid area somehow it will just continue to be there. As long as we make sure that rebels causing a province or nation to break free get deleted, we should be fine.

src/culture/rebels.cpp
wxwisiasdf: they do not leabve the nation, defections cause them to

schombert: Then you should make sure that, when a defection happens, the rebels in the province go away

PR #751, opened 2023-12-09
rotating autosaves (5)

wxwisiasdf:


schombert:
you will need some way to put the most recent one on top of the loading list; right now that is done by file name, which obviously won't work for this

schombert:
why did you close this?

PR #752, opened 2023-12-09
multi-language support

wxwisiasdf:
without needing to recreate scenario


schombert:
I am reluctant to add this because (a) it will bloat scenario files and (b) it isn't compatible with my long-term vision for how localization will work. Eventually, to move over to a proper localization system, languages will have to be identified by the canonical locale strings, such as "en-us". This means moving beyond a system where there are some fixed number of languages and to a system where there can be any number of languages.

schombert:
That is a post 1.0 sort of change. I imagine that it would go alongside switching to utf8 in the files.

PR #753, opened 2023-12-09
FIX PADDING ON GFX OBJECTS CAUSING ISSUES

wxwisiasdf:
I TOOK A HEX EDITOR AND COMAPRED TWO DUMPS (OF RAW SCENARIO DATA NOT JUST THE DCON PART) AND THEN I FOUND OUT THAT THE GFX WAS GOING WILD SHIT AJSDKLGJNDFGKH

### DIFF
--- a/src/gui/gui_graphics.hpp
+++ b/src/gui/gui_graphics.hpp
@@ -47,7 +47,6 @@ struct gfx_object {
 	uint16_t type_dependent = 0; // secondary texture handle or border size -- 8bytes
 	uint8_t flags = 0; // 9bytes
 	uint8_t number_of_frames = 1; // 10bytes
-	uint8_t padding[6] = {}; // 16bytes
 
 	object_type get_object_type() const {
 		return object_type(flags & type_mask);
@@ -65,7 +64,7 @@ struct gfx_object {
 		return (flags & do_transparency_check) != 0;
 	}
 };
-static_assert(sizeof(gfx_object) == 16);
+static_assert(sizeof(gfx_object) == 10);
 
 enum class element_type : uint8_t { // 3 bits
 	button = 0x01,
--- a/src/gui/gui_graphics.hpp
+++ b/src/gui/gui_graphics.hpp
@@ -32,6 +32,7 @@ struct xy_pair {
 	int16_t x = 0;
 	int16_t y = 0;
 };
+static_assert(sizeof(xy_pair) == 4);
 
 struct gfx_object {
 	constexpr static uint8_t always_transparent = 0x10;
@@ -108,7 +109,6 @@ struct text_base_data {
 	dcon::text_sequence_id txt; // 4bytes
 	uint16_t font_handle = 0; // 6bytes
 	uint8_t flags = 0; // 7bytes
-	uint8_t padding = 0; // 8bytes
 
 	alignment get_alignment() const {
 		return alignment(flags & alignment_mask);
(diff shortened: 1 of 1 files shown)


schombert:
WHY ARE WE SHOUTING?

src/gui/gui_graphics.hpp
```diff
@@ -65,6 +64,7 @@ struct gfx_object {
 		return (flags & do_transparency_check) != 0;
 	}
 };
+static_assert(sizeof(gfx_object) == 16);
```
schombert, line 67: Are you sure that you need padding for this one? It looks to me like it should be 2 byte aligned, and thus that 10 bytes will fill it completely. If it isn't 2 byte aligned, try just forcing its alignment to that

src/gui/gui_graphics.hpp
```diff
@@ -166,15 +170,16 @@ struct text_data : public text_base_data {
...
 	static constexpr uint8_t is_mask_mask = 0x80;
 
-	float scale = 1.0f;
-	dcon::gfx_object_id gfx_object;
-
-	uint8_t flags = 0;
+	float scale = 1.0f; // 4bytes
+	dcon::gfx_object_id gfx_object; // 6bytes
+	uint8_t flags = 0; // 7bytes
+	uint8_t padding = 0; // 8bytes
```
schombert, line 182: This doesn't need padding, because it is going into a union. You need to make sure that the union as a whole is memset to 0 before anything is assigned to it.

src/gui/gui_graphics.hpp
```diff
@@ -183,18 +188,23 @@ struct image_data {
...
 	}
 };
+static_assert(sizeof(image_data) == 8);
 
 struct overlapping_data {
-	float spacing = 1.0f;
-	alignment image_alignment = alignment::left;
+	float spacing = 1.0f; // 4bytes
+	alignment image_alignment = alignment::left; // 5bytes
+	uint8_t padding[3] = {}; // 8bytes
```
schombert, line 196: as above

src/gui/gui_graphics.hpp
```diff
@@ -183,18 +188,23 @@ struct image_data {
...
 struct list_box_data {
-	xy_pair border_size;
-	xy_pair offset;
-	dcon::gfx_object_id background_image;
-	uint8_t spacing = 0;
+	xy_pair border_size; // 4bytes
+	xy_pair offset; // 8bytes
+	dcon::gfx_object_id background_image; // 10bytes
+	uint8_t spacing = 0; // 11bytes
+	uint8_t padding[5] = {}; // 16bytes
```
schombert, line 205: as above

src/gui/gui_graphics.hpp
```diff
@@ -262,27 +273,27 @@ struct element_data {
...
-		list_box_data list_box;				// +11
-		scrollbar_data scrollbar;			//+10
-		window_data window;						// +4
-		position_data position;				//+0
+		list_box_data list_box; // +11
+		scrollbar_data scrollbar; //+10
+		window_data window; // +4
+		position_data position; //+0
 
 		internal_data() {
```
schombert, line 291: and this is the union that needs to be zeroed. Doing it in the constructor should be fine

src/gui/gui_graphics.hpp
```diff
@@ -262,27 +273,27 @@ struct element_data {
...
 			position = position_data{};
 		}
 	} data; // +12 = 24
 
-	uint8_t flags = 0;		// 25
+	uint8_t flags = 0; // 25
 	uint8_t ex_flags = 0; // 26
```
schombert, line 298: you will need 2 bytes of padding here -- this is a 28 byte size struct and ex_flags finishes at byte 26

src/gui/gui_graphics.hpp
```diff
@@ -136,6 +139,7 @@ struct button_data : public text_base_data {
 		return (text_base_data::flags & is_checkbox_mask) != 0;
 	}
 };
+static_assert(sizeof(button_data) == 4);
```
schombert, line 142: this inherits from text base data, so ...

src/gui/gui_graphics.hpp
```diff
@@ -166,15 +170,16 @@ struct text_data : public text_base_data {
 		return (text_base_data::flags & is_edit_mask) != 0;
 	}
 };
+static_assert(sizeof(text_data) == 4);
```
schombert, line 173: also inherits from text base data

wxwisiasdf:
Great stuff - now I can reuse scenarios!

schombert:
you didn't answer my questions about the changes to gfx_object

wxwisiasdf:
> you didn't answer my questions about the changes to gfx_object

which

wxwisiasdf:
> Are you sure that you need padding for this one? It looks to me like it should be 2 byte aligned, and thus that 10 bytes will fill it completely. If it isn't 2 byte aligned, try just forcing its alignment to that

well it makes the scenarios match now

PR #754, opened 2023-12-09
allow configurable fonts

wxwisiasdf:


schombert:
I like the font we have now, so I am not going to merge this. If you want to just add a third font slot and some way to make them easier to change via a file, I would accept that.

PR #755, opened 2023-12-09
allow work for non-avx2 compilations

wxwisiasdf:


schombert:
this looks like you are trying to support 32bit, which I am definitely not interested in

wxwisiasdf:
> this looks like you are trying to support 32bit, which I am definitely not interested in

looks like doesnt mean it is

schombert:
Ok, then why the changes to detect WIN32 and casting things to size_t?

wxwisiasdf:
> Ok, then why the changes to detect WIN32 and casting things to size_t?

😭

schombert:
You don't have to sob, I just want to understand why the changes are being made

wxwisiasdf:
the most easy way for me to ensure a non-AVX2 can be done, is allowing alice to be compiled on 32-bit; so i can check for stray intrinsics

schombert:
The problem doesn't appear to be stray intrinsics. It looks like the library implementation of one of the atomic functions uses an AVX instruction to me

PR #759, opened 2023-12-09
Fix "restoration of the"

wxwisiasdf:


src/military/military.cpp
```diff
@@ -2028,6 +2028,12 @@ void populate_war_text_subsitutions(sys::state& state, dcon::war_id w, text::sub
 
 	dcon::nation_id primary_attacker = state.world.war_get_primary_attacker(war);
 	dcon::nation_id primary_defender = state.world.war_get_original_target(war);
```
schombert, line 2030: if this isn't set, it is a bug. Can you tell me why it isn't being set?

PR #761, opened 2023-12-09
properly update province window when owner of said province changes

wxwisiasdf:


src/gui/gui_province_window.hpp
```diff
@@ -1881,6 +1881,13 @@ class province_view_window : public window_element_base {
 		}
 	}
 
+	void on_update(sys::state& state) noexcept override {
+		/* Properly updates the window when ownership changes - prevents weird bugs */
+		if(state.province_ownership_changed.load(std::memory_order_acquire) == true) {
```
schombert, line 1886: this is not a great fix, because the main loop is going to clear that ASAP, meaning that this will only work if the ui update happens to run in a quite narrow window

src/gui/gui_province_window.hpp
```diff
@@ -1882,10 +1875,11 @@ class province_view_window : public window_element_base {
...
 	void on_update(sys::state& state) noexcept override {
-		/* Properly updates the window when ownership changes - prevents weird bugs */
-		if(state.province_ownership_changed.load(std::memory_order_acquire) == true) {
-			set_active_province(state, active_province);
-		}
+		header_window->impl_on_update(state);
+		foreign_details_window->impl_on_update(state);
+		local_details_window->impl_on_update(state);
+		local_buildings_window->impl_on_update(state);
+		colony_window->impl_on_update(state);
```
schombert, line 1882: Does this actually work? They are child windows, so they should be getting update messages anyways. Is something incorrectly calling `on_update` on this windows instead of `impl_on_update`?

wxwisiasdf: they are set invisible, which excludes them from the update process
here they be explicitly updated so they set visibility back again (if needed)

schombert: ok, makes sense, although set_visible true as appropriate would be better. Whoever wrote the code that makes them visible in their update functions must not have understood that invisible elements don't get updated.

PR #763, opened 2023-12-10
More v2-looking map

wxwisiasdf:
image

unsure if you are willing to merge this as-is through, if not, then could we make this a setting? 🥺 

note: font unrelated, pr only change map, not font or text


schombert:
besides showing the rivers all the time (fine I guess, if your graphics card can handle it), what is this changing?

wxwisiasdf:
> besides showing the rivers all the time (fine I guess, if your graphics card can handle it), what is this changing?

border width basically

PR #764, opened 2023-12-10
proper casing for some csv entries

wxwisiasdf:


schombert:
Are you sure that none of these appear in the middle of sentences? Check the trigger and effects tooltips files to make sure that they aren't being used as pieces of a single line

wxwisiasdf:
> Are you sure that none of these appear in the middle of sentences? Check the trigger and effects tooltips files to make sure that they aren't being used as pieces of a single line

doesnt seem to be the case, to the contrary they all seem to appear at the beginning of,

PR #765, opened 2023-12-10
display a white piechart instead of a black one

wxwisiasdf:
original OpenV2 did this, why don't we?


schombert:
racist

wxwisiasdf:
noooooooooo

PR #768, opened 2023-12-11
Made it so the build unit buttons in the Military window are disabled when the player is unable to build more units.

Masterchef118:


wxwisiasdf:
AND when you dont want to cancel any other in-pending units holy fuck that was fast i didnt finish to type and schomby already merged it

PR #771, opened 2023-12-11
States use capital name as fallback

wxwisiasdf:


src/text/text.cpp
```diff
@@ -841,6 +841,8 @@ std::string get_dynamic_state_name(sys::state const& state, dcon::state_instance
 			return adj + " " + get_name_as_string(state, fat_id.get_definition());
```
schombert, line 841: you also need to add a case here -- for if it is a shared state that is missing its name. Otherwise, this solution was exactly what I was going to do.

PR #772, opened 2023-12-11
Load up guantlet cursor as custom cursor

wxwisiasdf:


schombert:
but what about linux?

wxwisiasdf:
will handle later when i get a linux machine, this is a harmless feature anyways so

PR #774, opened 2023-12-12
Nationalism AI; changes for encouraging more nationalism to successfully suceed

ma44:
Big list of changes
- Rebel AI can now properly form pan nationalist factions, although nationalists still appear to be broken, this should drastically increase the chance of an Italy appearing alongside the occasional Gutter crown for the germans
- Nations will try to unally and desphere countries they have wargoals against; stops certain situations like USA sphering mexico causing them to not be able to manifest a destiny
- Modified some tech weights, effective outcome should be a much higher focus on national focus techs, culture and army
- Redid the sphere AI; now tries to not sphere countries they don't have a neighboring sphere or a coastal state to access. Countries will now hyperfocus on sphering countries within their culture union
- AI will now intervene in wars even if they're already in a war if it means they can protect a non-GP culture union; this could use some military strength checks, but this is mainly to ensure the USA will keep Texas safe from the Mexicans and not be pseudo locked out from manifest destiny
- Wars that extend for > 6 years and have minor amounts of war score will now be peaced out easily; stops forever stalemate wars particularly with Europeans doing "cut down to size" wars against asian countries. 
- Countries with very high war exhaustion will peace out after 1 year in war; countries locked in rebel death spiral will try to escape the situation, especially when the aggressor cannot occupy the lands because of rebels
- GPs will now look globally to figure out where they can use their wargoals; helps Prussia decide in attacking Denmark and would allow for other free wargoals to be used like 'pay debts' and 'cut down to size'
- Undid the rebel size reduction, rebels winning a revolution will now negate militancy; this should help nations not ener a permanent death spiral just because the population is permanently unhappy

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -8,6 +8,11 @@ float estimate_strength(sys::state& state, dcon::nation_id n) {
 	float value = state.world.nation_get_military_score(n);
 	for(auto subj : state.world.nation_get_overlord_as_ruler(n))
 		value += subj.get_subject().get_military_score();
+	//Leaders currently make minor nations (especially spherelings) seem much more powerful than they are, so it's getting axed here
+	auto gen_range = state.world.nation_get_leader_loyalty(n);
+	auto num_leaders = float((gen_range.end() - gen_range.begin()));
+	value -= num_leaders;
+
 	return value;
 }
 
@@ -166,8 +171,12 @@ void prune_alliances(sys::state& state) {
 		if(!n.get_is_player_controlled() && !n.get_ai_is_threatened() && !(n.get_overlord_as_subject().get_ruler())) {
 			prune_targets.clear();
 			for(auto dr : n.get_diplomatic_relation()) {
+				auto other = dr.get_related_nations(0) != n ? dr.get_related_nations(0) : dr.get_related_nations(1);
+				if(military::can_use_cb_against(state, n, other)) {
+					prune_targets.push_back(other);
+					continue;
+				}
 				if(dr.get_are_allied()) {
-					auto other = dr.get_related_nations(0) != n ? dr.get_related_nations(0) : dr.get_related_nations(1);
 					if(other.get_in_sphere_of() != n) {
 						prune_targets.push_back(other);
 					}
@@ -269,6 +278,11 @@ bool ai_will_accept_alliance(sys::state& state, dcon::nation_id target, dcon::na
 	// Otherwise we may consider alliances only iff they are close to our continent or we are adjacent
 	if(!ai_is_close_enough(state, target, from))
 		return false;
+
+	//We'd rather want to kill them rather than ally
+	if(military::can_use_cb_against(state, from, target)) {
+		return false;
+	}
 
 	// And also if they're powerful enough to be considered for an alliance
 	auto target_score = estimate_strength(state, target);
@@ -385,10 +399,10 @@ void initialize_ai_tech_weights(sys::state& state) {
 	for(auto t : state.world.in_technology) {
 		float base = 1000.0f;
 		if(state.culture_definitions.tech_folders[t.get_folder_index()].category == culture::tech_category::army)
-			base *= 1.5f;
+			base *= 3.0f;
 
 		if(t.get_increase_building(economy::province_building_type::naval_base))
-			base *= 1.1f;
+			base *= 9.0f;
 		else if(state.culture_definitions.tech_folders[t.get_folder_index()].category == culture::tech_category::navy)
 			base *= 0.9f;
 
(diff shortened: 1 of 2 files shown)


src/ai/ai.cpp
```diff
@@ -166,8 +171,12 @@ void prune_alliances(sys::state& state) {
 		if(!n.get_is_player_controlled() && !n.get_ai_is_threatened() && !(n.get_overlord_as_subject().get_ruler())) {
 			prune_targets.clear();
 			for(auto dr : n.get_diplomatic_relation()) {
+				auto other = dr.get_related_nations(0) != n ? dr.get_related_nations(0) : dr.get_related_nations(1);
```
schombert, line 174: I don't think there is any point putting this check outside of the are allied check. The prune list is only used to remove existing alliances, and checking for possible cbs against nations you aren't allied to here would be kind of a waste if there is nothing to prune.

src/ai/ai.cpp
```diff
@@ -270,6 +279,11 @@ bool ai_will_accept_alliance(sys::state& state, dcon::nation_id target, dcon::na
 	if(!ai_is_close_enough(state, target, from))
 		return false;
 
+	//We'd rather want to kill them rather than ally
+	if(military::can_use_cb_against(state, from, target)) {
+		return false;
```
schombert, line 284: This will need a corresponding line added to the offer alliance button in the diplomacy window explaining why the ai won't accept

src/ai/ai.cpp
```diff
@@ -385,30 +399,30 @@ void initialize_ai_tech_weights(sys::state& state) {
 	for(auto t : state.world.in_technology) {
 		float base = 1000.0f;
 		if(state.culture_definitions.tech_folders[t.get_folder_index()].category == culture::tech_category::army)
-			base *= 1.5f;
+			base *= 3.0f;
 
 		if(t.get_increase_building(economy::province_building_type::naval_base))
-			base *= 1.1f;
+			base *= 9.0f;
```
schombert, line 405: really, *nine* times more weight to increasing naval base level? Nothing should have more weight than the weight given to techs that increase the research point modifier. I should also point out that only the relative values matter. Increasing all the weights won't really matter; only the relative weightings matter in most cases

src/ai/ai.cpp
```diff
@@ -532,6 +583,9 @@ void perform_influence_actions(sys::state& state) {
 				command::execute_remove_from_sphere(state, gprl.get_great_power(), gprl.get_influence_target(), gprl.get_influence_target().get_in_sphere_of());
 			} else if(state.defines.addtosphere_influence_cost <= gprl.get_influence() && !current_sphere && clevel == nations::influence::level_friendly) {
 				command::execute_add_to_sphere(state, gprl.get_great_power(), gprl.get_influence_target());
+			//De-sphere countries we have wargoals against
+			} else if(military::can_use_cb_against(state, gprl.get_great_power(), gprl.get_influence_target()) && state.defines.removefromsphere_influence_cost <= gprl.get_influence() && current_sphere && clevel == nations::influence::level_friendly) {
+				command::execute_remove_from_sphere(state, gprl.get_great_power(), gprl.get_influence_target(), gprl.get_influence_target().get_in_sphere_of());
```
schombert, line 588: This needs to check for whether the nation has the infamy room. We wouldn't want a nation pushing themselves over the infamy cap for this

src/ai/ai.cpp
```diff
@@ -1833,11 +1887,23 @@ bool will_accept_crisis_peace_offer(sys::state& state, dcon::nation_id to, dcon:
 
 void update_war_intervention(sys::state& state) {
 	for(auto& gp : state.great_nations) {
-		if(state.world.nation_get_is_player_controlled(gp.nation) == false && state.world.nation_get_is_at_war(gp.nation) == false) {
```
schombert, line 1836: Are you sure removing this is a good idea? I don't think that the AI should generally be jumping into more wars if it is already at war

src/ai/ai.cpp
```diff
@@ -1833,11 +1887,23 @@ bool will_accept_crisis_peace_offer(sys::state& state, dcon::nation_id to, dcon:
...
+		if(state.world.nation_get_is_player_controlled(gp.nation) == false){
 			bool as_attacker = false;
 			dcon::war_id intervention_target;
 			[&]() {
 				for(auto w : state.world.in_war) {
+					//GPs will try to intervene in wars to protect smaller nations in the same cultural union
+					if(command::can_intervene_in_war(state, gp.nation, w, false)) {
+						for(auto par : w.get_war_participant()) {
```
schombert, line 1897: instead of looping over everyone in the war, just check the primary defender -- this will prevent the US jumping into any war that texas gets involved in as a defensive ally, for an example of the weird things that could happen

src/ai/ai.cpp
```diff
@@ -2735,6 +2800,14 @@ bool will_accept_peace_offer(sys::state& state, dcon::nation_id n, dcon::nation_
 		if(war_duration < 365) {
 			return false;
 		}
+		//The country is ruined and it's been a bit into the war, we'll recover later
+		if(state.world.nation_get_war_exhaustion(n) > 80) {
+			return true;
+		}
+		//Stalemated war, peace out
+		if(war_duration > (365 * 6) && (overall_score <= 10 && overall_score >= -10)) {
```
schombert, line 2808: maybe only if the offer is a white peace? Otherwise this will let you take 100% warscore in a stalemated war

src/ai/ai.cpp
```diff
@@ -2735,6 +2800,14 @@ bool will_accept_peace_offer(sys::state& state, dcon::nation_id n, dcon::nation_
 		if(war_duration < 365) {
 			return false;
 		}
+		//The country is ruined and it's been a bit into the war, we'll recover later
+		if(state.world.nation_get_war_exhaustion(n) > 80) {
```
schombert, line 2804: these changes will also need to be reflected in the function that calculates the same thing with a bunch of extra parameters (this is used for evaluating a peace offer possibility without actually making it, mostly for the ui)

src/culture/rebels.cpp
```diff
@@ -389,26 +402,13 @@ bool pop_is_compatible_with_rebel_faction(sys::state& state, dcon::pop_id p, dco
...
-		return true;
-	}
 	return true;
 }
 
 bool pop_is_compatible_with_rebel_type(sys::state& state, dcon::pop_id p, dcon::rebel_type_id t) {
 	auto fac = fatten(state.world, t);
 	auto pop = fatten(state.world, p);
 
-	if(fac.get_ideology() && fac.get_ideology() != pop.get_dominant_ideology())
```
schombert, line 410: why did you remove this check? This prevents, say, pops that are mostly reactionary from joining communist rebels, etc

schombert: oh, I see, it just got moved

schombert:
Something weird seems to be going on here. This PR says that it was merged, but I don't recall doing that (and didn't intended to before the issues raised above were addressed). Nor does it appear to be actually merged (its changes don't appear to be in the source files). Can anyone clue me into what is going on?

ma44:
Mystery solved

PR #777, opened 2023-12-12
Makes war exhaustion and occupation matter more

ma44:
- War exhaustion now gives a hefty militancy debuff
- Occupied provinces will no longer employ anyone for RGO production, leading to mass unemployment
- Nothing in this PR for implementing country wide RGO/factory debuffs or disabling factories when a state is occupied


coderguy57:
A bit weird if all farmers just stop working because the province is occupied. A fully occupied nation would lead to mass starvation

ma44:
> A bit weird if all farmers just stop working because the province is occupied. A fully occupied nation would lead to mass starvation
@coderguy57 

While I agree there can be more "realistic" mechanics attached to this, I'm just copying the behavior that is present within the Vanilla Victoria 2 experience

coderguy57:
Huh, I didnt know that was how it worked in vanilla Vic2. Then I have no objection 👍

PR #784, opened 2023-12-16
war start date display

wxwisiasdf:
as per feature request

### DIFF
--- a/src/gui/topbar_subwindows/gui_diplomacy_window.hpp
+++ b/src/gui/topbar_subwindows/gui_diplomacy_window.hpp
@@ -1452,13 +1452,13 @@ public:
 	void update_tooltip(sys::state& state, int32_t x, int32_t y, text::columnar_layout& contents) noexcept override {
 		dcon::war_id content = retrieve<dcon::war_id>(state, parent);
 		auto fat_id = dcon::fatten(state.world, content);
-
-		auto box = text::open_layout_box(contents, 0);
-		text::substitution_map sub{};
-		text::add_to_substitution_map(sub, text::variable_type::date, fat_id.get_start_date());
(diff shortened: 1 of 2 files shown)


src/gui/topbar_subwindows/gui_diplomacy_window.hpp
```diff
@@ -1452,6 +1452,15 @@ class war_side_strength_text : public button_element_base {
 	void update_tooltip(sys::state& state, int32_t x, int32_t y, text::columnar_layout& contents) noexcept override {
 		dcon::war_id content = retrieve<dcon::war_id>(state, parent);
 
+		sys::year_month_day start_date = o.get_war().get_start_date().to_ymd();
```
schombert, line 1455: you can just write the base text with a single variable to substitute and then pass the date in directly to the substitution map without converting it into anything. It will then be expanded to a formatted date for the variable (just like we do for country ids -> names)

src/gui/topbar_subwindows/gui_diplomacy_window.hpp
```diff
@@ -1451,8 +1451,15 @@ class war_side_strength_text : public button_element_base {
 
 	void update_tooltip(sys::state& state, int32_t x, int32_t y, text::columnar_layout& contents) noexcept override {
 		dcon::war_id content = retrieve<dcon::war_id>(state, parent);
-
 		auto fat_id = dcon::fatten(state.world, content);
+
+		sys::year_month_day start_date = fat_id.get_start_date().to_ymd(state.start_date);
+		auto box = text::open_layout_box(contents, 0);
+		text::substitution_map sub{};
+		text::add_to_substitution_map(sub, text::variable_type::date, start_date);
```
schombert, line 1459: pass get_start_date here; you don't need to convert it to a ymd

PR #789, opened 2023-12-17
AI nationalist enhanced, take 2

ma44:
Reopening this PR for now after the last one got accidentally auto-merged because of a mistake on my part.

Big list of changes
- Rebel AI can now properly form pan nationalist factions, although nationalists still appear to be broken, this should drastically increase the chance of an Italy appearing alongside the occasional Gutter crown for the germans
- Nations will try to unally and desphere countries they have wargoals against; stops certain situations like USA sphering mexico causing them to not be able to manifest a destiny
- Modified some tech weights, effective outcome should be a much higher focus on national focus techs, culture and army
- Redid the sphere AI; now tries to not sphere countries they don't have a neighboring sphere or a coastal state to access. Countries will now hyperfocus on sphering countries within their culture union
- AI will now intervene in wars even if they're already in a war if it means they can protect a non-GP culture union; this could use some military strength checks, but this is mainly to ensure the USA will keep Texas safe from the Mexicans and not be pseudo locked out from manifest destiny

~~Wars that extend for > 6 years and have minor amounts of war score will now be peaced out easily; stops forever stalemate wars particularly with Europeans doing "cut down to size" wars against asian countries.
 Countries with very high war exhaustion will peace out after 1 year in war; countries locked in rebel death spiral will try to escape the situation, especially when the aggressor cannot occupy the lands because of rebels~~

- GPs will now look globally to figure out where they can use their wargoals; helps Prussia decide in attacking Denmark and would allow for other free wargoals to be used like 'pay debts' and 'cut down to size'
- Undid the rebel size reduction, rebels winning a revolution will now negate militancy by a good amount; this should help nations not enter a permanent death spiral because of occupied provinces generating vast amounts of militancy and should result in a mostly calm nation after 1-2 mega rebellions, though it's not a solution to having a poor economy


ma44:
I'll request another review once I finish addressing the comments posted on #774 
@schombert

PR #793, opened 2023-12-18
black and white font fixes for text

wxwisiasdf:
makes mods like CWE behave better with text


src/gui/gui_event.cpp
```diff
@@ -383,7 +383,7 @@ void event_desc_text::on_update(sys::state& state) noexcept {
 
 	auto contents = text::create_endless_layout(delegate->internal_layout,
 			text::layout_parameters{0, 0, static_cast<int16_t>(base_data.size.x), static_cast<int16_t>(base_data.size.y),
-					delegate->base_data.data.text.font_handle, 0, text::alignment::left, text::text_color::black, false});
+					delegate->base_data.data.text.font_handle, 0, text::alignment::left, delegate->black_text ? text::text_color::black : text::text_color::white, false});
```
schombert, line 386: then you need to update the loop below too

PR #794, opened 2023-12-18
Implement multiplayer state transfers

wxwisiasdf:
:D


schombert:
I'll only consider it if you fix the bug I assigned you

PR #796, opened 2023-12-19
Add support for a religion map mode

OptimizedForDensity:
Adds mod support for a religion map mode (`mapmode_24`) that works in the same way as the cultural map mode although it *does not* actually add the button to activate the map mode into PA. The GUI needs modified to add a button for it by adding a block of code like this to `interface/menubar.gui`. Note that this code specifically reuses the naval map mode icon and it is out of place so further UI tweaking would be needed but the feature is there for UI mods.

```
guiButtonType = {
	name = "mapmode_24"
	position = { x = -62 y = -136 }
	quadTextureSprite ="GFX_mapmode_naval"
	Orientation = "LOWER_RIGHT"
	clicksound = click
}
```

Pictures

Screenshot1

Screenshot2

Screenshot3


wxwisiasdf:
what if we just held shift when clicking on the nationality mapmode button instead?

PR #798, opened 2023-12-19
Add 2D railroads to the map

wxwisiasdf:


assets/shaders/railroad_line_f.glsl
```diff
@@ -0,0 +1,17 @@
+in float tex_coord;
```
schombert, line 1: why does the railroad need its own fragment shader?

src/gui/gui_nation_picker.hpp
```diff
@@ -215,6 +215,9 @@ class select_save_game : public button_element_base {
 			} else {
 				state.fill_unsaved_data();
 			}
+			/* Savefiles might load with new railroads, so for responsiveness we
```
schombert, line 218: this is true, but it probably makes more sense to put this at the end of regenerate unsaved data, since you know that has to be called wherever loading is done

PR #799, opened 2023-12-20
Add religious conversions for pops

OptimizedForDensity:
Religious conversions were something from Victoria 2 that were missing so this lets pops convert to the state religion of the nation they in if the following hardcoded conditions are met

- The province the pop is in isn't unowned
- The pop isn't following the state religion
- There's at least 1 pop following the state religion in the province

Conversion scale can be edited in the defines file then other variable modifiers are defined in `common/pop_types.txt` and any edits here will affect the conversion chance. I think vanilla V2 required that there be a pop of the same type as well as the same religion in the province to allow conversion so this is a slight change.

Two new modifiers were also added that I don't think were present in Victoria 2 but they were pretty much low hanging fruit to implement. The new modifiers are `global_conversion_rate` (nation scope) and `conversion_rate` (province scope) which are the conversion analogues to `global_assimilation_rate` and `assimilation_rate`.

In the pop demographics page, predicted conversions/month can be viewed by hovering your mouse over the religion icon. Note that in this image, most of the modifiers are mod-specific so they'll look differently depending on whichever mod you play with.
Screenshot


OptimizedForDensity:
I didn't notice any increase in tick time in my tests but I haven't done any serious profiling so far.

schombert:
I would like to see some numbers comparing its performance in release so that I can make an informed decision. I am happy to add it if the cost turns out to be negligible, which it may very well since I believe that most of our time is in fact spent elsewhere.

schombert:
Let me know if you want to try to get this in by 1.0. If you can get me some numbers before the end of the week, it could probably make it.

OptimizedForDensity:
I did some rudimentary profiling using `std::chrono` where the I timed `single_game_tick()` in `system_state.cpp` as a whole then the two places where the conversion code was added in this function. I let the game run for a year using HPM with the game compiled in the debug configuration the total amount of time spent running `single_game_tick()` was 130 seconds for this year while the total amount of time spent for updating religious conversions then applying them totaled just 1.2 seconds. It seems pretty negligible to me.

schombert:
That does sound reasonable, but you need to test in release, not in debug (release can perform very differently than debug). It is also probably easier to use visual studio's built-in profiler ( https://learn.microsoft.com/en-us/visualstudio/profiling/beginners-guide-to-performance-profiling?view=vs-2022 ) to measure the relative time spent in the demographics update function with and without your pr.

OptimizedForDensity:
Unfortunately I'm using VS Code since I'm on Linux so I don't have access to that profiler, but I did re-run my previous profiling using the release configuration which gave a cumulative tick time for a year of 7.3 seconds while the religious conversion parts took 0.14 seconds. If you'd like more in-depth profiling I'll have to figure something else out.

PR #803, opened 2023-12-24
Option to paint vassals of the same colour

wxwisiasdf:


schombert:
uh oh, I was already implementing a version of this that would put them as a derived color

PR #807, opened 2023-12-28
pad diplomatic messages

wxwisiasdf:


schombert:
you would also need to memset 0 the union member on construction, since not all members are the same size

PR #811, opened 2023-12-28
render impassable borders on higher zoom levels

wxwisiasdf:
makes the map on higher levels look nicer

### DIFF
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -560,6 +560,15 @@ void display_data::render(sys::state& state, glm::vec2 screen_size, glm::vec2 of
 				glDrawArrays(GL_TRIANGLE_STRIP, b.start_index,  b.count);
 			}
 		}
+	} else {
+		glUniform1f(4, 0.0003f); // width
+		glActiveTexture(GL_TEXTURE14);
+		glBindTexture(GL_TEXTURE_2D, textures[texture_imp_border]);
+		for(auto b : borders) {
+			if((state.world.province_adjacency_get_type(b.adj) & (province::border::non_adjacent_bit | province::border::coastal_bit)) == 0) {
+				glDrawArrays(GL_TRIANGLE_STRIP, b.start_index, b.count);
+			}
+		}
 	}
 	// national borders
 	{
--- a/src/map/map.cpp
+++ b/src/map/map.cpp
@@ -529,55 +529,80 @@ void display_data::render(sys::state& state, glm::vec2 screen_size, glm::vec2 of
 
 	//glMultiDrawArrays(GL_TRIANGLE_STRIP, coastal_starts.data(), coastal_counts.data(), GLsizei(coastal_starts.size()));
 	
-	if(zoom > 8) { // Render province borders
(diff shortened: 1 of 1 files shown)


src/map/map.cpp
```diff
@@ -551,7 +551,7 @@ void display_data::render(sys::state& state, glm::vec2 screen_size, glm::vec2 of
 		}
 	}
 	// impassible borders
-	{
+	if(zoom > 5) {
```
schombert, line 554: but then *no* border will render here at all

src/map/map.cpp
```diff
@@ -560,6 +560,15 @@ void display_data::render(sys::state& state, glm::vec2 screen_size, glm::vec2 of
 				glDrawArrays(GL_TRIANGLE_STRIP, b.start_index,  b.count);
 			}
 		}
+	} else {
+		glUniform1f(4, 0.0003f); // width
+		glActiveTexture(GL_TEXTURE14);
+		glBindTexture(GL_TEXTURE_2D, textures[texture_imp_border]);
+		for(auto b : borders) {
+			if((state.world.province_adjacency_get_type(b.adj) & (province::border::non_adjacent_bit | province::border::coastal_bit)) == 0) {
```
schombert, line 568: I think you should test the result that this produces

src/map/map.cpp
```diff
@@ -529,46 +529,80 @@ void display_data::render(sys::state& state, glm::vec2 screen_size, glm::vec2 of
...
 		}
-	}
-	if(zoom > 5) { // Render state borders
-		glUniform1f(4, 0.0002f); // width
-		glActiveTexture(GL_TEXTURE14);
-		glBindTexture(GL_TEXTURE_2D, textures[texture_state_border]);
-		for(auto b : borders) {
-			if((state.world.province_adjacency_get_type(b.adj) & (province::border::non_adjacent_bit | province::border::coastal_bit | province::border::impassible_bit | province::border::national_bit | province::border::state_bit)) == province::border::state_bit) {
-				glDrawArrays(GL_TRIANGLE_STRIP, b.start_index, b.count);
+		{ // Render state borders
```
schombert, line 545: doesn't this change mean that you are always rendering the state borders?

src/map/map.cpp
```diff
@@ -529,46 +529,80 @@ void display_data::render(sys::state& state, glm::vec2 screen_size, glm::vec2 of
...
+			glActiveTexture(GL_TEXTURE14);
+			glBindTexture(GL_TEXTURE_2D, textures[texture_national_border]);
+			for(auto b : borders) {
+				if((state.world.province_adjacency_get_type(b.adj) & (province::border::non_adjacent_bit | province::border::coastal_bit | province::border::national_bit | province::border::impassible_bit)) == province::border::national_bit) {
+					glDrawArrays(GL_TRIANGLE_STRIP, b.start_index, b.count);
+				}
+			}
+		}
+	} else {
+		if(zoom > 8) { // Render province borders
```
schombert, line 577: else ... if zoom is < 5 ... if zoom is > 8 ?

PR #812, opened 2023-12-28
1.reduce curveness of the railroad so it doesnt go into the ocean as …

wxwisiasdf:
…much.2.avoid impassables or straits


src/map/map.cpp
```diff
@@ -1073,7 +1073,7 @@ void create_railroad_connection(sys::state& state, std::vector<glm::vec2>& railr
 	//
 	auto mid = ((m1 + m2) / glm::vec2(2.f, 2.f));
 	const auto rp = rng::get_random(state, p1.index(), p2.index() ^ p1.index());
-	const float rf = float(rng::reduce(uint32_t(rp), 8192)) / (8192.f * 0.1f);
+	const float rf = float(rng::reduce(uint32_t(rp), 8192)) / (8192.f * 0.055f);
```
schombert, line 1076: this will increase how far the curve can bend out, not decrease it

PR #814, opened 2023-12-28
add message when scenario checksums dont match - that is true from wh…

wxwisiasdf:
…at ive been testing


schombert:
or that they haven't clicked make scenario, which is something that you should expect to happen a lot

wxwisiasdf:
there

PR #823, opened 2023-12-31
Fix compiler error

OptimizedForDensity:
Complains about `std::max(int, uint32_t)`.


OptimizedForDensity:
@schombert for future reference, I was able to compile the new code in this file on Linux using the `windows.h` file from Wine and as far as I can tell the new bitmap code works as intended on Linux. I wasn't able to get the game fully compiled using Wine since the launcher still has too much Windows-specific code but it seems possible to use some Windows library functions on non-Windows systems without needing to maintain separate code.

PR #832, opened 2024-01-01
change uses of system_clock to steady_clock

yrrrmmbl:
using std::chrono::system_clock for timing stuff could lead to funnies

### DIFF
--- a/src/map/map_state.cpp
+++ b/src/map/map_state.cpp
@@ -299,9 +299,9 @@ void update_text_lines(sys::state& state, display_data& map_data) {
 }
 
 void map_state::update(sys::state& state) {
-	std::chrono::time_point<std::chrono::system_clock> now = std::chrono::system_clock::now();
+	std::chrono::time_point<std::chrono::steady_clock> now = std::chrono::steady_clock::now();
 	// Set the last_update_time if it hasn't been set yet
-	if(last_update_time == std::chrono::time_point<std::chrono::system_clock>{})
+	if(last_update_time == std::chrono::time_point<std::chrono::steady_clock>{})
 		last_update_time = now;
 
 	update_unit_arrows(state, map_data);
--- a/src/map/map_state.hpp
+++ b/src/map/map_state.hpp
@@ -59,7 +59,7 @@ public:
 	bool is_dragging = false;
 
 	// Last update time, used for smooth map movement
-	std::chrono::time_point<std::chrono::system_clock> last_update_time{};
+	std::chrono::time_point<std::chrono::steady_clock> last_update_time{};
 
 	// Time in seconds, send to the map shader for animations
 	float time_counter = 0;


schombert:
Could it? Would you mind explaining what you are worried about?

yrrrmmbl:
https://en.cppreference.com/w/cpp/chrono/system_clock

> Class std::chrono::system_clock represents the system-wide real time wall clock.
> The system_clock's time value can be internally adjusted at any time by the operating system, for example due to NTP synchronization or the user changing the system's clock. Daylight Saving Time and time zone changes, however, do not affect it since it is based on the [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) time-zone.

system_clock isnt guaranted to have monotonic ticks nor to progress steadily. theoretically it could supply any possible timepoint.
system_clock is meant to get the real time. for example for a calendar or showing what time it is or something. for timing stuff we should use steady_clock as it is guaranteed to progress steadily and have monotonic ticks

schombert:
But that is exactly why I used system_clock. System clock is trying to match the "real time" as much as possible, which means that the game ticks it measures out will come at regular real time intervals from the user's point of view. My concern about steady clock is that, as far as I can tell, the standard doesn't guarantee a resolution for it, and we need something with a resolution of about 1ms to make the faster ticks regular. In particular I am worried about things like this: https://stackoverflow.com/questions/29551126/does-steady-clock-have-only-10ms-resolution-on-windows-cygwin

yrrrmmbl:
system_clock is not monotonic. i dont know about usage on cygwin but both msvc and clangs libc++ have high_resolution_clock, which is guaranteed to be the highest resolution clock, aliased to steady_clock. the problem with high_resolution_clock is that it is not universally guaranteed to be monotonic. this can however be checked with a static_assert(std::chrono::high_resolution_clock::is_steady). i habitually use steady_clock and have never had problems with resolution. we can use high_resolution_clock safely if we check for steadiness.

https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

yrrrmmbl:
:(

yrrrmmbl:
@schombert schould it stay as it was or change clocks to high_resolution_clock s although they are not steady?

schombert:
you could even go with steady clock if you know that its resolution will work for us. I basically just want you to double check that we won't actually have resolution problems, at which point I am happy to go with whatever you think is best

schombert:
I have left this open because I don't know where your research landed on the resolution question

yrrrmmbl:
well i would suggest using steady_clock. if we have problems we can change to something different. but as long as we stick to native compiled with msvc or clang we shouldnt have any weird timing resolution problems.

https://youtu.be/P32hvk8b13M?t=2865
this video is from 2016 but it still sounds like good advice.

PR #835, opened 2024-01-01
optimization bringing down 15~ secs per minute to 3~ secs per minute

wxwisiasdf:
omg i can go to 1845 in lightspeed

### DIFF
--- a/src/economy/economy.cpp
+++ b/src/economy/economy.cpp
@@ -1773,29 +1773,17 @@ void update_pop_consumption(sys::state& state, dcon::nation_id n, ve::vectorizab
 		state.world.nation_get_modifier_values(n, sys::national_mod_offsets::rich_luxury_needs) + 1.0f,
 	};
 
-	static auto ln_real_demand_vector = state.world.commodity_make_vectorizable_float_buffer();
-	state.world.execute_serial_over_commodity([&](auto ids) { ln_real_demand_vector.set(ids, ve::fp_vector{}); });
-	static auto en_real_demand_vector = state.world.commodity_make_vectorizable_float_buffer();
-	state.world.execute_serial_over_commodity([&](auto ids) { en_real_demand_vector.set(ids, ve::fp_vector{}); });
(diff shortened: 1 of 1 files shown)


src/economy/economy.cpp
```diff
@@ -1719,78 +1719,85 @@ void update_pop_consumption(sys::state& state, dcon::nation_id n, dcon::province
...
+	static auto lx_real_demand_vector = state.world.commodity_make_vectorizable_float_buffer();
+	state.world.execute_serial_over_commodity([&](auto ids) { lx_real_demand_vector.set(ids, ve::fp_vector{}); });
+	for(uint32_t i = 1; i < total_commodities; ++i) {
+		dcon::commodity_id cid{ dcon::commodity_id::value_base_t(i) };
+		for(const auto t : state.world.in_pop_type) {
+			auto strata = state.world.pop_type_get_strata(t);
+			ln_real_demand_vector.get(cid) += state.world.pop_type_get_life_needs(t, cid) * ln_demand_vector.get(t) * base_demand * ln_mul[strata] * (state.world.nation_get_life_needs_weights(n, cid) + 1.0f);
+			en_real_demand_vector.get(cid) += state.world.pop_type_get_everyday_needs(t, cid) * en_demand_vector.get(t) * base_demand * invention_factor * en_mul[strata] * (state.world.nation_get_everyday_needs_weights(n, cid) + 1.0f) * en_extra_factor;
+			lx_real_demand_vector.get(cid) += state.world.pop_type_get_luxury_needs(t, cid) * lx_demand_vector.get(t) * base_demand * invention_factor * lx_mul[strata] * (state.world.nation_get_luxury_needs_weights(n, cid) + 1.0f) * lx_extra_factor;
```
schombert, line 1788: can you cut out these intermediate real demand vectors now and just add the results directly to nation_get_real_demand ? (also, add all three values together first and them add them to the nations demand or get a reference to the demand float and add them to the reference, to avoid the cost of looking up the value three times)

PR #836, opened 2024-01-02
accepted_cultures becomes a bitfield

wxwisiasdf:
1.leaving this for review mainly
2.i will provide measurements of tests later


src/scripting/triggers.cpp
```diff
@@ -1413,7 +1413,7 @@ TRIGGER_FUNCTION(tf_accepted_culture) {
 	auto is_accepted = ve::apply(
```
schombert, line 1413: here is_accepted can be loaded directly, without apply, by simply doing nation_get_accepted_cultures(to_nation(primary_slot), trigger::payload(tval[1]).cul_id)

PR #841, opened 2024-01-02
Added "nextsong" command

aintDatCap:


schombert:
it can't hurt to merge this, but I will give you a heads up: we used to have some buttons for doing this in the esc menu. The problem was, sound::play_new_track() is random, and people were expecting to be able to cycle through the songs in some ordered way (and I guess to be able to go back to the previous one if they skipped too far ahead? maybe? I don't quite understand everything that was expected). Your build failure by the way is because you are using the os-dependent pointer directly rather than the wrapper function.

aintDatCap:
> it can't hurt to merge this, but I will give you a heads up: we used to have some buttons for doing this in the esc menu. The problem was, sound::play_new_track() is random, and people were expecting to be able to cycle through the songs in some ordered way (and I guess to be able to go back to the previous one if they skipped too far ahead? maybe? I don't quite understand everything that was expected). Your build failure by the way is because you are using the os-dependent pointer directly rather than the wrapper function.

So you're suggesting that we should add something like a jukebox like in hoi4?

schombert:
I think that is what people really want when they ask for track switching. But that is a much larger task and will require some extensions to the very simple sound system wrapper.

PR #846, opened 2024-01-02
do not redundantly regenerate demographics for adding wargoals?

wxwisiasdf:


schombert:
This is unfortunately necessary for the game to function. The on add effect of wargoals reduces jingoism support. That effect doesn't show up in national demographics until the update is done. And that propagation is in turn necessary to make the "you have to have this much jingoism support to add more wargoals" check work. If you want to come up with another workaround, that would be great, but until then, it has to stay

wxwisiasdf:
just update jingoism then?

schombert:
the issue is that it needs to propagate up to the national demographics. And yeah, we could write a function to do just that, but ... lazyness

PR #851, opened 2024-01-03
headless mode

wxwisiasdf:
reasons why have headless mode:
a) test in vtune performance improvments without the noise from opengl drivers
b) the public server idea

### DIFF
--- a/src/entry_point_win.cpp
+++ b/src/entry_point_win.cpp
@@ -123,6 +123,7 @@ int WINAPI wWinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPWSTR
 		int num_params = 0;
 		auto parsed_cmd = CommandLineToArgvW(GetCommandLineW(), &num_params);
 
+		bool headless = false;
 		if(num_params < 2) {
 #ifdef NDEBUG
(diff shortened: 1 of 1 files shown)


src/entry_point_win.cpp
```diff
@@ -211,14 +214,21 @@ int WINAPI wWinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPWSTR
...
-
-		// entire game runs during this line
-		window::create_window(game_state,
-				window::creation_parameters{1024, 780, window::window_state::maximized, game_state.user_settings.prefer_fullscreen});
+		if(headless) {
+			game_state.actual_game_speed = 6;
+			game_state.ui_pause.store(false, std::memory_order::release);
+			game_state.mode = sys::game_mode_type::in_game;
+			game_state.local_player_nation = dcon::nation_id{};
+			game_state.game_loop();
```
schombert, line 222: but how do you quit?

wxwisiasdf: VTune can stop/kill the process, which is sufficient to quit
also the user can manually Ctrl-C the process or kill it
requiring input to quit may not be a good idea since we dont provide a console per se, and this is more meant to be a "alice runs as a children of another process who controls when it starts/stops/etc"

PR #852, opened 2024-01-03
multiplayer in spanish

elmatero08:
Spanish translation of multiplayer instructions


schombert:
ok, it seems to no longer be interfering with the english version, but the new spanish file isn't included

PR #854, opened 2024-01-03
fix + qol improvement: changed instant research cheat implementation

aintDatCap:


src/gamestate/system_state.cpp
```diff
@@ -3898,6 +3898,15 @@ void state::single_game_tick() {
 			ai::refresh_home_ports(*this);
 			break;
 		case 1:
+			// Instant research cheat
+			if(this->user_settings.instant_research_enabled) {
+				auto n = this->local_player_nation;
```
schombert, line 3903: I believe that this would instantly de-sync multiplayer, as each instance of the game has a different local player set and so would end up with different gamestates after this step. Now, I don't know if leaf designed the cheats to work in multiplayer. If they are blocked from working in multiplayer, than this is fine, but if they can be used in multiplayer, then some other approach will be needed.

src/gui/gui_console.cpp
```diff
@@ -1191,7 +1191,7 @@ void ui::console_edit::edit_box_enter(sys::state& state, std::string_view s) noe
 		command::c_complete_constructions(state, state.local_player_nation);
 		break;
 	case command_info::type::instant_research:
-		command::c_change_research_points(state, state.local_player_nation, float(640000000.f));
+		state.user_settings.instant_research_enabled = !state.user_settings.instant_research_enabled;
```
schombert, line 1194: This would also be a problem for multiplayer: any user-initiated change to the game state that can affect how the game progresses must be wrapped in a command so that it can be sent to the server which then synchronizes it with any other commands to produce a single authoritative sequence of the order in which things happen and what happened

src/gamestate/commands.cpp
```diff
@@ -5177,6 +5178,8 @@ void execute_command(sys::state& state, payload& c) {
 	case command_type::c_complete_constructions:
 		execute_c_complete_constructions(state, c.source);
 		break;
+	case command_type::c_instant_research:
+		execute_c_instant_research(state, c.source);
```
schombert, line 5182: needs a break I think so that the next time we add a command someone doesn't introduce a bug

PR #857, opened 2024-01-03
added ticking casualties to battles

Wonobory:
added a text that shows the casualties taken of each side in the last tick, very useful for knowing if a battle is going well and how expensive in terms of manpower is being.

PR


schombert:
you need to take out any windows-specific stuff you added for debugging

src/local_user_settings.hpp.bak
```diff
@@ -0,0 +1 @@
+#pragma once
```
schombert, line 1: you seem to have added this file by accident

src/military/military.cpp
```diff
@@ -5181,7 +5189,7 @@ void update_land_battles(sys::state& state) {
 					case unit_type::special:
 						state.world.land_battle_get_defender_support_lost(b) += str_damage;
 						break;
-					default:
```
schombert, line 5184: why did you remove the default case? yes, it shouldn't occur in a sane world, but if something really wacky has happened (like ship types appearing here), that is what it exists for. I also makes the static analysis tools happy

src/military/military.cpp
```diff
@@ -5202,6 +5210,8 @@ void update_land_battles(sys::state& state) {
 				str_damage = std::min(str_damage, cstr);
 				state.world.regiment_get_pending_damage(att_front[i]) += str_damage;
 				cstr -= str_damage;
+				attacker_casualties += str_damage;
```
schombert, line 5213: you seem to be missing the equivalent addition for the backline inflicted defender casualties (there should be 2 of each, and you have only one place defender casualties are increases)

src/military/military.cpp
Wonobory: i think i removed it by accident

src/military/military.cpp
Wonobory: isn't already in line 5176?

schombert: I see two attacker casualties += but only one defender casualties +=; there should be two of each (but I was just guessing about which was was dropped since I am just looking at the changes without much context here)

Wonobory: well, idk if this is what you meant but i noticed that the front line damage dealt was flipped (so the attacker fronline damage dealt was added to attacker_casualties instead of defender_casualties)

PR #860, opened 2024-01-03
Add ability to name wars based on belligerents

OptimizedForDensity:
A casus belli gives a war its name, usually something like `WAR_CONQUEST_ANY_NAME` which looks up a localization string `NORMAL_WAR_CONQUEST_ANY_NAME`, but Victoria 2 also had a feature where you can specify a localization string `NORMAL_WAR_CONQUEST_ANY_NAME_ABC_DEF` which, if there was a war where the primary attacker had the tag `ABC` and the primary defender had the tag `DEF`, would give that specific war the name specified in the localization file for that combination of tags, otherwise it would default to the usual war name if there wasn't any. Mods like HPM used this for wars like the Oriental Crisis and European intervention against Egypt for example.

This feature has been implemented, although the code isn't the prettiest since it needs to look up the raw localization tag instead of using the `dcon::text_sequence_id` which is stored for the war, however this approach doesn't require storing any additional data. Iterating over the entire text sequence map to look up a key given a value isn't efficient, however the war name is only generated for certain UI functions so performance shouldn't be an issue. I'm open to feedback if there's a nicer way to do this.

Here's a snapshot from a test game with this PR.

Screenshot

### DIFF
--- a/src/gui/topbar_subwindows/gui_diplomacy_window.hpp
+++ b/src/gui/topbar_subwindows/gui_diplomacy_window.hpp
@@ -14,7 +14,7 @@
 #include "gui_crisis_window.hpp"
 
 namespace military {
-void populate_war_text_subsitutions(sys::state&, dcon::war_id, text::substitution_map&);
+std::string get_war_name(sys::state&, dcon::war_id);
 }
 
 namespace ui {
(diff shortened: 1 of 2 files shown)


schombert:
let me think for a new moments on whether there is a faster way to get back to the original key

PR #862, opened 2024-01-04
micro optimize nation_accepts_culture

wxwisiasdf:


src/nations/nations.cpp
```diff
@@ -74,15 +74,11 @@ dcon::nation_id get_nth_great_power(sys::state const& state, uint16_t n) {
...
-				if(n)
-					return state.world.nation_get_accepted_cultures(n, c);
-				else
-					return false;
-			},
-			ids, cul_ids);
-	return (state.world.nation_get_primary_culture(ids) == cul_ids) || is_accepted;
+	return ve::apply(
+		[&state](dcon::nation_id n, dcon::culture_id c) {
+			return state.world.nation_get_accepted_cultures(n, c)
```
schombert, line 79: it is faster to do the || outside the apply because then it does that operation and the primary culture lookup as a single instruction each. By doing it this way, you are doing the || and the lookup many more times

PR #863, opened 2024-01-04
fix conquer all/tag console command

wxwisiasdf:


schombert:
has it been tested?

src/gui/gui_console.cpp
```diff
@@ -1207,18 +1207,17 @@ void ui::console_edit::edit_box_enter(sys::state& state, std::string_view s) noe
 	case command_info::type::conquer_tag:
 	{
 		auto tag = std::get<std::string>(pstate.arg_slots[0]);
-		auto nid = smart_get_national_identity_from_tag(state, parent, tag);
 		if(tag == "ALL" || tag == "all") {
-			auto n = state.world.national_identity_get_nation_from_identity_holder(nid);
 			for(const auto po : state.world.in_province_ownership) {
 				if(po.get_nation() != state.local_player_nation)
-					command::c_change_owner(state, state.local_player_nation, po.get_province(), po.get_nation());
+					command::c_change_owner(state, state.local_player_nation, po.get_province(), state.local_player_nation);
```
schombert, line 1213: won't this also break multiplayer, given that local nation is different on each machine? Or are these commands all blocked in mp? We raised that question in bug reports

wxwisiasdf: all console commands using command:: are blocked in mp

PR #865, opened 2024-01-05
add minimum and maximum window size to windows

rarafael:


src/window/window_win.cpp
```diff
@@ -297,6 +297,12 @@ LRESULT CALLBACK WndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam)
 		// this is the message that tells us there is a DirectShow event
 		sound::update_music_track(*state);
 		break;
+	case WM_GETMINMAXINFO:
+		LPMINMAXINFO info = static_cast<LPMINMAXINFO>(lParam);
+		info->ptMinTrackSize.x = 640;
+		info->ptMinTrackSize.y = 400;
+		info->ptMaxTrackSize.x = 2400;
```
schombert, line 304: I don't think that you should touch ptMaxTrackSize -- my monitor is larger than this, for example, and I wouldn't like to be prevented from making my window as large as I want

rarafael: hm, I just assigned what were the max sizes on the glfw version, I guess removing it does not harm

PR #867, opened 2024-01-05
Fx g cbugs

wxwisiasdf:


src/gui/gui_province_window.hpp
```diff
@@ -1194,10 +1194,11 @@ class province_rebel_percent_text : public simple_text_element_base {
 		// substitute.
 		auto box = text::open_layout_box(contents, 0);
 		text::localised_single_sub_box(state, contents, box, std::string_view("avg_mil_on_map"), text::variable_type::value, text::fp_one_place{ province::revolt_risk(state, content) });
```
schombert, line 1196: suggest replacing this with add_line and removing the layout box completely

PR #877, opened 2024-01-05
as per f.r: 3 map options for vassal colouring

wxwisiasdf:


src/gui/gui_main_menu.cpp
```diff
@@ -297,6 +297,43 @@ void gamma_display::on_update(sys::state& state) noexcept {
...
+		send(state, parent, notify_setting_update{});
+	}
+}
+void vassal_color_left::on_update(sys::state& state) noexcept {
+	disabled = (state.user_settings.vassal_color == sys::map_vassal_color_mode(0));
+}
+void vassal_color_right::button_action(sys::state& state) noexcept {
+	auto index = uint8_t(state.user_settings.vassal_color);
+	if(state.user_settings.vassal_color < sys::map_vassal_color_mode(2)) {
+		state.user_settings.vassal_color = sys::map_vassal_color_mode(index + 1);
```
schombert, line 313: this won't work, visually, because it doesn't actually force the colors to be regenerated

src/map/modes/political.hpp
```diff
@@ -10,9 +10,27 @@ std::vector<uint32_t> political_map_from(sys::state& state) {
...
+				auto ovr = state.world.nation_get_overlord_as_subject(id);
+				if(ovr) {
+					auto ruler = state.world.overlord_get_ruler(ovr);
+					color = state.world.nation_get_color(id);
+				}
+				break;
+			case sys::map_vassal_color_mode::none:
+				auto ovr = state.world.nation_get_overlord_as_subject(id);
+				if(ovr) {
+					color = state.world.national_identity_get_color(state.world.nation_get_identity_from_identity_holder(id));
```
schombert, line 32: or rather, this won't work, because determining the color is more complicated than that (it depends partly on the nation's current government type). Furthermore, a common asked-for extension is for effects that can change a nation's color. Which, this would prevent from working.

schombert: I suggest moving the vassal coloring logic out of the update political function and move it here, possibly with caching if it is too expensive. Then it could be turned on and off trivially

PR #882, opened 2024-01-05
Fix pop data regeneration

OptimizedForDensity:
`extra_size = sz - csz` so `if(index >= csz)`, `index += extra_size * some positive integer` will always be `>= sz` so the function returns instead of updating certain pop data like was mentioned in the comments here 001716c (although I think the cause was in a different commit).

### DIFF
--- a/src/economy/demographics.cpp
+++ b/src/economy/demographics.cpp
@@ -63,7 +63,7 @@ uint32_t size(sys::state const& state) {
 				 uint32_t(2) * state.world.pop_type_size() + state.world.culture_size() + state.world.religion_size();
 }
 uint32_t common_size(sys::state const& state) {
-	return count_special_keys + uint32_t(2) * state.world.pop_type_size();
+	return count_special_keys;
 }
 
 template<typename F>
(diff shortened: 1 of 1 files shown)


src/economy/demographics.cpp
```diff
@@ -63,7 +63,7 @@ uint32_t size(sys::state const& state) {
 				 uint32_t(2) * state.world.pop_type_size() + state.world.culture_size() + state.world.religion_size();
 }
 uint32_t common_size(sys::state const& state) {
-	return count_special_keys + uint32_t(2) * state.world.pop_type_size();
+	return count_special_keys;
```
schombert, line 66: the pop size bit was intentional, because I want to force the per-pop-type population numbers and the per-pop-type employment numbers to be updated daily (those numbers are used by the economy update in a couple places, and so I considered them more crucial to keep up to date correctly)

src/economy/demographics.cpp
```diff
@@ -140,7 +140,7 @@ void regenerate_from_pop_data(sys::state& state) {
 		auto index = base_index;
 		if constexpr(!full) {
 			if(index >= csz) {
-				index += extra_size * (state.current_date.value % extra_demo_grouping);
```
schombert, line 143: but, oh shit, yeah that was a big bug

src/economy/demographics.cpp
OptimizedForDensity: I see now, reverted.

PR #883, opened 2024-01-05
stack cycling into boats

wxwisiasdf:


src/gamestate/commands.cpp
```diff
@@ -3156,8 +3156,16 @@ void execute_move_army(sys::state& state, dcon::nation_id source, dcon::army_id
...
-	if(battle && !province::has_naval_access_to_province(state, source, dest)) {
-		return;
+	if(dest.index() < state.province_definitions.first_sea_province.index()) {
+		/* Case for land destinations */
+		if(battle && !province::has_naval_access_to_province(state, source, dest)) {
+			return;
+		}
+	} else {
+		/* Case for naval destinations, we check the land province adjacent henceforth */
+		if(battle && !province::has_naval_access_to_province(state, source, state.world.army_get_location_from_army_location(a))) {
```
schombert, line 3166: This ... won't work for several reasons. Naval access == when you could dock in a province. So, if you are fighting in enemy territory, you wouldn't have naval access, and so you wouldn't be able to retreat onto boats. Secondly, the retreat order may be more than one province, so you could be testing some far away province for access to determine whether the army can move to some distant sea province. And maybe you should check whether there are boats with capacity to hold the army floating there?

wxwisiasdf: wouldnt that be handled by military::emark tho through?

schombert: the last bit? sure. but if they can't actually get on the boats it would probably be best to treat it like trying to retreat to an illegal province: the retreat isn't issued at all and the units stay in combat. But in any case, point three is mainly a qol issue. The other two are more critical in terms of making it work correctly.

PR #888, opened 2024-01-05
provid tooltip and parsing fix

yrrrmmbl:


schombert:
this should work

schombert:
I accept that it doesn't, but it should

src/gui/gui_console.cpp
```diff
@@ -313,7 +317,7 @@ parser_state parse_command(sys::state& state, std::string_view text) {
 	parser_state pstate{};
 	pstate.cmd = possible_commands[0];
 	for(auto const& cmd : possible_commands)
-		if(s.starts_with(cmd.name)) {
+		if(s.compare(0, cmd.name.size(), cmd.name) == 0) {
```
schombert, line 320: won't this have the same problem? It will compare the first cmd.name.size() characters of cmd to s. In other words, it will return true if cmd is a prefix of s, which is what starts with was doing, no?

yrrrmmbl: yes

yrrrmmbl: already realized

yrrrmmbl:
not sure if my fix is too ugly. but dont know what else to do

schombert:
there is a function in parsers called `remove_surrounding_whitespace` that you can use so `if(remove_surrounding_whitespace(s) == cmd.name)`

yrrrmmbl:
that removes surrounding whitespace but i remove all whitespace and everything after it

schombert:
ah, so s includes the parameters too, I see

PR #894, opened 2024-01-06
Colonial power tooltip detail added

iwanim:
Colonial power functions broken into parts for intuitive call


schombert:
I don't know what the status is of this, but I will point out that the encoding of the csv file seems to have been accidentally altered

iwanim:
Yeah that's why I closed it. I think I'm missing characters on my computer and it replaces them because it does it both in my browser and in Visual Studio. I will make a new one, sec

PR #895, opened 2024-01-06
Colonial power tooltip detail added

iwanim:
The contents of max_colonial_points() and free_colonial_points() functions have been separated into functions in the same namespace so they can be called in a way that looks intuitive for the newly added details in gui_topbar, instead of basically repeating the contents of nations.cpp there

The csv file needs to be updated with some text keys because as you have seen my computer is messing it up. It will work fine without them so I can add them later if you don't want to
explain_colonial_points;Secondary and Great Powers accumulate colonial points.
colonial_points_from_technology;From technologies:
colonial_points_from_naval_bases;From naval bases:
colonial_points_from_ships;From ships:
used_colonial_maintenance;Maintaining existing colonies:


schombert:
do these csv entries need a space after the colon?

iwanim:
No it's fine the spaces are added with add_space_to_layout_box()

PR #899, opened 2024-01-06
RGO payment tweaks

OptimizedForDensity:
According to [the wiki page for RGOs](https://vic2.paradoxwikis.com/Resource_gathering_operation), the payment for RGO work is allocated like this: RGO owners get $\min\left(0.5 \times \text{RGO income}, 2 \times \text{RGO income} \times \frac{\text{number of owners}}{\text{number of workers (paid or unpaid)}}\right)$ then paid workers get whatever is left over. Additionally, RGO workers don't get minimum wage.

This PR changes the payment formula to match this one (with the additional change that unpaid pops aren't counted in the formula since that's how PA already calculates the number of workers), meaning that RGO owners get $\min\left(0.5 \times \text{RGO income}, 2 \times \text{RGO income} \times \frac{\text{number of owners}}{\text{number of paid workers}}\right)$ and paid workers get whatever is left over. It also accounts for an edge case where all the RGO workers are unpaid, meaning that owners will get all the money instead of some getting lost.

### DIFF
--- a/src/economy/economy.cpp
+++ b/src/economy/economy.cpp
@@ -10,7 +10,7 @@ bool can_take_loans(sys::state& state, dcon::nation_id n) {
 		return false;
 
 	/*
-	A country cannot borrow if it is less than define:BANKRUPTCY_EXTERNAL_LOAN_YEARS since their last bankruptcy. 
+	A country cannot borrow if it is less than define:BANKRUPTCY_EXTERNAL_LOAN_YEARS since their last bankruptcy.
 	*/
 	auto last_br = state.world.nation_get_bankrupt_until(n);
 	if(last_br && state.current_date < last_br)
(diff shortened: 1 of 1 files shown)


schombert:
I was aware that our payment formulas weren't exactly the same (the ones for factories are also different). I am willing to bring them more in line with the V2 formulas, if you think that is better, *but* I would very much like to keep the minimum wage for non-slave RGO workers. RGO workers tend to be a huge chunk of the wage-earners, even in late game, and I don't think that they should be exempt from the minimum wage laws.

OptimizedForDensity:
I wasn't aware that that formula was deliberately changed so I'll close this PR.

schombert:
Ok, but I am open to changes; I just want to maintain a minimum wage

PR #901, opened 2024-01-06
Fix compiler warnings

OptimizedForDensity:
First commit is just whitespace fixes that VS Code automatically made, the second one is addressing different compiler warnings that GCC throws. Most of them were complaining about `memset` or `memcpy` on structs with non-trivial constructors and most of those were replaced by a call to the struct's constructor.

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -1010,7 +1010,7 @@ void update_ai_econ_construction(sys::state& state) {
 								// TODO: try to delete a factory here
 							}
 					} // END for(auto si : ordered_states) {
-				} // END if((rules & issue_rule::build_factory) == 0) 
+				} // END if((rules & issue_rule::build_factory) == 0)
 			} // END if(!desired_types.empty()) {
 		} // END  if((rules & issue_rule::expand_factory) != 0 || (rules & issue_rule::build_factory) != 0)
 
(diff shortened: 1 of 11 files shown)


src/ai/ai.cpp
```diff
@@ -1638,8 +1638,7 @@ void update_crisis_leaders(sys::state& state) {
 								for(auto s : potential_states) {
 									if(military::cb_instance_conditions_satisfied(state, par.id, target, offer_cb, state.world.state_instance_get_definition(s), dcon::national_identity_id{}, dcon::nation_id{})) {
 
-										diplomatic_message::message m;
-										memset(&m, 0, sizeof(diplomatic_message::message));
```
schombert, line 1642: Are you sure that these memsets are safe to remove? Sometimes they are used to ensure that padding and the unused parts of unions are still set to 0 so that, if they get saved to a file via memcopy, a checksum generated from that will always produce the same result (which is necessary to avoid false positive out of sync detections in multiplayer)

src/filesystem/simple_fs_nix.cpp
```diff
@@ -192,7 +192,7 @@ void restore_state(file_system& fs, native_string_view data) {
 namespace impl {
 bool contains_non_ascii(native_char const* str) {
 	for(auto c = str; *c != 0; ++c) {
-		if(int32_t(*c) > 127 || int32_t(*c) < 0)
```
schombert, line 195: this was a protection against the possibility that `char` could be unsigned in some compilers

src/gamestate/cheats.cpp
```diff
@@ -10,8 +10,7 @@ namespace command {
 void add_to_command_queue(sys::state& state, payload& p);
 
 void c_switch_nation(sys::state& state, dcon::nation_id source, dcon::national_identity_id t) {
-	payload p;
-	memset(&p, 0, sizeof(payload));
```
schombert, line 14: see above

src/ai/ai.cpp
OptimizedForDensity: Hmm I haven't been able to get local multiplayer on two clients working to test out desyncs so I'm can't guarantee it. I'll close this PR for now since it's just a file cleanup.

PR #904, opened 2024-01-07
Rebel hunter balance estimations (this makes rebels more tolerable)

wxwisiasdf:


schombert:
this very much looks like your rebel hunters will commit suicide, especially early game

wxwisiasdf:
> this very much looks like your rebel hunters will commit suicide, especially early game

but it wolrks

PR #907, opened 2024-01-07
Finished GLSL conversion

aintDatCap:


schombert:
There is another PR in the works that adds an additional projection function, which means adding an additional projection function into the shaders. Do you still want this to be merged now, or would you rather include the additional projection function in this PR?

aintDatCap:
You can merge this pr right now

PR #908, opened 2024-01-07
Implementation of perspective projection.

ineveraskedforthis:
Implement perspective projection function in vertex shaders.
Adjust according functions.


assets/shaders/glsl/textured_line_b_v.glsl
```diff
@@ -42,7 +42,7 @@ vec4 globe_coords(vec2 world_pos) {
 	new_world_pos.xyz += 0.5; 	// Move the globe to the center
 
 	return vec4(
-		(2. * new_world_pos.x - 1.f)  * zoom,
```
schombert, line 45: Does putting the aspect ratio adjustment here work properly? It was moved out originally to fix the joins where the lines met doing weird things.

ineveraskedforthis: I haven't noticed. How did weirdness looked like?

schombert: You might have to zoom in pretty closely to see it. The problem was that when lines made sharper turns you would sometimes see the corners "stick out" too much. It was caused by the lines needing to also project the neighboring points in order to know the angle between them, which in turn determines where the points for the join need to go. Scaling for the aspect ratio threw off those calculations.

ineveraskedforthis: This shader behavior should be fixed in the next commit.

assets/shaders/glsl/text_line_v.glsl
```diff
@@ -44,7 +44,7 @@ vec4 globe_coords(vec2 world_pos) {
 	new_world_pos.xyz += 0.5; 	// Move the globe to the center
 
 	return vec4(
-		(2. * new_world_pos.x - 1.f) * zoom,
+		(2. * new_world_pos.x - 1.f) * zoom / aspect_ratio,
```
schombert, line 47: aspect ratio has appeared here

assets/shaders/glsl/text_line_v.glsl
```diff
@@ -54,9 +54,42 @@ vec4 flat_coords(vec2 world_pos) {
 	world_pos += vec2(-offset.x, offset.y);
 	world_pos.x = mod(world_pos.x, 1.0f);
 	return vec4(
-		(2. * world_pos.x - 1.f) * zoom * map_size.x / map_size.y,
+		(2. * world_pos.x - 1.f) * zoom / aspect_ratio * map_size.x / map_size.y,
```
schombert, line 57: and here

PR #914, opened 2024-01-08
increase netowkr time

wxwisiasdf:


schombert:
If a timeout of a minute isn't enough, I doubt that one of 10 minutes will be better, and at some point you are just keeping dead connections open for no reason ...

wxwisiasdf:
> If a timeout of a minute isn't enough, I doubt that one of 10 minutes will be better, and at some point you are just keeping dead connections open for no reason ...

you dont know how latin american internet exists

schombert:
If you want to make things more robust, I think that you will need to go in the other direction. You will need to be sending your own connection testing packets regularly and doing more work to try to drop connections as soon as there is a problem and to re-establish connections seamlessly from the perspective of the players

PR #920, opened 2024-01-09
fix macro builder reporting less units than what can be built

wxwisiasdf:


schombert:
does this account for the fact that in some provinces there may be enough free soldiers from a single pop to build more than one regiment out of?

schombert:
You can use these functions:
```
int32_t regiments_created_from_province(sys::state& state, dcon::province_id p); // does not include mobilized regiments
int32_t regiments_max_possible_from_province(sys::state& state, dcon::province_id p);
int32_t main_culture_regiments_created_from_province(sys::state& state, dcon::province_id p);
int32_t main_culture_regiments_max_possible_from_province(sys::state& state, dcon::province_id p);
```
to make your life easier

PR #923, opened 2024-01-11
Add support for COUNTRY_ADJ and RELIGION tokens in rebel names

OptimizedForDensity:
Re-adding support for $COUNTRY_ADJ$ as a rebel name token since that was something that Victoria 2 had then also added in support for a $RELIGION$ token since that type of rebel group is supported (although I'm not sure what the V2 token was since it was never really used).

### DIFF
--- a/src/text/text.hpp
+++ b/src/text/text.hpp
@@ -289,6 +289,7 @@ enum class variable_type : uint16_t {
 	region,
 	rel,
 	relation,
+	religion,
 	req,
 	reqlevel,
 	required,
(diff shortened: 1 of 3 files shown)


src/text/text.hpp
```diff
@@ -289,6 +289,7 @@ enum class variable_type : uint16_t {
 	region,
 	rel,
 	relation,
+	religion,
```
schombert, line 292: You will also need to add religion to the place where the text is parsed into variables for it to be recognized (text.cpp line 209, where they are grouped by length)

PR #931, opened 2024-01-11
implemented "addpop" cheat

aintDatCap:


src/nations/nations.cpp
```diff
@@ -798,6 +798,17 @@ int32_t max_colonial_points(sys::state& state, dcon::nation_id n) {
 	}
 }
 
+uint32_t get_total_pop_in_owned_provinces(sys::state& state, dcon::nation_id n) {
```
schombert, line 801: this can be accomplished by looking up the `demographics::total` value in the nation's demographics

PR #933, opened 2024-01-12
initialize data in commands

wxwisiasdf:


wxwisiasdf:
this is so visual studio stops complaining i am aware we already do a memset yes

schombert:
These structs only exist as members of a union, and obviously they can't actually be initialized in this way, since that would require contradictory things ...

PR #935, opened 2024-01-12
add missing csv key for ortho perspective

wxwisiasdf:


assets/alice.csv
```diff
@@ -9,6 +9,7 @@ window_mode_label;Fullscreen;;;Pantalla completa;;;;;;;;;x
 map_projection;Map projection;;;Proyeccion del mapa;;;;;;;;;x
 map_projection_globe;Orthographic;;;Ortografica;;;;;;;;;x
 map_projection_flat;Equirectangular;;;Equirectangular;;;;;;;;;x
+map_projection_globe_perspective;Ortho-perspective
```
schombert, line 12: but, it isn't orthographic; that is what makes it different from the projection we already had

wxwisiasdf: what do we name it then

PR #939, opened 2024-01-12
consistency for UI sacale fix when classic fonts not used causing wei…

wxwisiasdf:
…rd padding

### DIFF
--- a/src/gui/gui_main_menu.cpp
+++ b/src/gui/gui_main_menu.cpp
@@ -58,9 +58,7 @@ void ui_scale_right::on_update(sys::state& state) noexcept {
 	disabled = (scale_index >= uint32_t(sys::ui_scales_count - 1));
 }
 void ui_scale_display::on_update(sys::state& state) noexcept {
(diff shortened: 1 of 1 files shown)


src/gui/gui_main_menu.cpp
```diff
@@ -58,9 +58,7 @@ void ui_scale_right::on_update(sys::state& state) noexcept {
 	disabled = (scale_index >= uint32_t(sys::ui_scales_count - 1));
 }
 void ui_scale_display::on_update(sys::state& state) noexcept {
-	std::string temp(16, '\0');
-	std::snprintf(temp.data(), temp.size(), "%.2fx", state.user_settings.ui_scale);
-	set_text(state, temp);
+	set_text(state, text::format_float(state, state.user_settings.ui_scale));
```
schombert, line 61: `+ "x"`

wxwisiasdf: meant to remove the x btw

PR #941, opened 2024-01-12
use alice menubar instead of vanilla one

wxwisiasdf:


wxwisiasdf:
a fucking oversight WHOOPS

PR #944, opened 2024-01-13
Compile parsers separately

OptimizedForDensity:
I've been playing around with pulling a lot of the `#include "[file].cpp"` out of `main.cpp` and compiling them separately and I've gotten it to the point where changing a line in the AI or economy code means the game recompiles in 10 seconds vs the original 3 minutes to recompile the entire game with `-O0`. This PR just separates the parsers from the main compilation which saves a bit of time for repeated compilations where the parser code doesn't change. If this is something you're interested in, I can open a few more PRs but I personally find it makes a huge difference for tweaking parts of the code. `system_state.hpp` gets moved to the precompiled header since ClangBuildAnalyzer says that its inclusion takes nearly a second per file and separate translation units means more inclusions so it adds up.


schombert:
I am interested in what you have done generally, but I would like it as a separate CMake target, leaving the unity build mostly as it is for the purposes of making a release. I am happy to merge this, however, because parsing isn't even part of the main executable, and so taking out of the unity build shouldn't impact anything.

schombert:
I should also add: don't forget to make sure that the launcher also compiles. That isn't covered by the CI.

OptimizedForDensity:
Not sure if there's a better way to do it with CMake but there's an additional AliceIncremental target that can support incremental building now. The original `main.cpp` file has an `#ifndef` macro that conditionally includes the incremental build files back in for the unity build. According to [this](https://github.com/OptimizedForDensity/Project-Alice/actions/runs/7514205445/job/20456815369), the launcher successfully compiles on Windows too.

schombert:
What I was thinking for the incremental target was removing `"src/main.cpp"` from it completely and instead giving it the list of cpp files, like a normal build would have. The ifdefs in main that select for the appropriate OS backend could be replaced by an if in cmake for that target, adding only the OS appropriate files to the target. That doesn't mean that we can't merge this as a first step however (although I am going to wait until after the weekend release, just in case)

PR #946, opened 2024-01-13
Added a controls option to change left mouse clicks so one must click, hold, and release over a button to activate it.

Masterchef118:


src/gamestate/system_state.cpp
```diff
@@ -2011,6 +2031,7 @@ void state::save_user_settings() const {
 	US_SAVE(autosaves);
 	US_SAVE(bind_tooltip_mouse);
 	US_SAVE(use_classic_fonts);
+	US_SAVE(left_mouse_click_hold_and_release);
```
schombert, line 2034: New settings need to be placed at the end of the order so that loading the settings from an older file won't give them random values

PR #947, opened 2024-01-13
[DISCUSS BEFORE MERGING] no avx2

aintDatCap:


schombert:
So, obviously we can't merge this as-is because I want to keep the main build the high performance build. So, do you want to introduce some new CMake target or something like that?

src/launcher/launcher_main.cpp
```diff
@@ -445,7 +448,7 @@ void mouse_click() {
 		return;
 	case ui_obj_play_game:
 		if(file_is_ready.load(std::memory_order::memory_order_acquire) && !selected_scenario_file.empty()) {
-			if(IsProcessorFeaturePresent(PF_AVX512F_INSTRUCTIONS_AVAILABLE)) {
```
schombert, line 448: This shouldn't be necessary, since if the CPU doesn't support avx, it certainly doesn't support avx512, and so this should return false.

aintDatCap:
> So, obviously we can't merge this as-is because I want to keep the main build the high performance build. So, do you want to introduce some new CMake target or something like that?

That would certainly be the best option

schombert:
closed as redundant now

PR #958, opened 2024-01-14
descriptive stuff in factory building (and show what the green coloured factories mean lol)

wxwisiasdf:


wxwisiasdf:
oh yeah it also fixes some NF tooltips too :D

PR #964, opened 2024-01-15
Rework loan system

OptimizedForDensity:
Fleshes out the loan system with mechanics for tracking how much one country owes another at any given time which helps with handling bankruptcy events. This is still a WIP but I wanted to see if there was anything I needed to change or if there were any suggestions, particularly for handling things like what happens when a country loans out a pop's savings then that pop goes to withdraw its savings (basically how to handle the reserve requirement of a national bank) and how to track how much of each pop's money is currently being loaned out for the purpose of interest payments. Victoria 2 had a bug where interest payments on national bank debt never made its way to the pops' pockets which meant that that money was vanishing from the simulation so I want to make sure that bug is fixed as well.

Currently implemented is the national bank system, "shadowy financiers" or private investors which Victoria 2 had to make sure there was money available for loans if no other nations had money available to loan, functions to take/repay loans given an amount of money to cover/repay, and bankruptcy turning loan interest off but there's still some interface work to do, more extensive testing, and some sort of AI loan logic would be nice.

I'll see about getting any sort of low-hanging fruit for increased modability into this PR without bloating the scope, I think an easy one would be having a separate interest rate (and modifiers) for national vs foreign debt.

### DIFF
--- a/src/economy/economy.cpp
+++ b/src/economy/economy.cpp
@@ -10,7 +10,7 @@ bool can_take_loans(sys::state& state, dcon::nation_id n) {
 		return false;
 
 	/*
-	A country cannot borrow if it is less than define:BANKRUPTCY_EXTERNAL_LOAN_YEARS since their last bankruptcy. 
+	A country cannot borrow if it is less than define:BANKRUPTCY_EXTERNAL_LOAN_YEARS since their last bankruptcy.
 	*/
 	auto last_br = state.world.nation_get_bankrupt_until(n);
 	if(last_br && state.current_date < last_br)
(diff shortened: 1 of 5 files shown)


schombert:
There will be some non-trivial work in getting an improved loan system. Specifically, you will have to introduce the concept of pop saving over and above pop money, as currently pops don't save. There will obviously be a temptation here to add a separate saving value and have some of the pop's money go into that "account" etc, etc. And maybe at the end of the day that is what you will go with. But I think that you should try to avoid that if at all possible. Any additional per-pop value that has to be updated each day is potentially a serious performance stumbling block because of the sheer number of pops and cache effects (probably, although, as with all things performance, only a measurement can say for sure). So you might want to look into "simulating" that bank value without storing it. For example, you could use some combination of the pop's most recent income and maybe needs satisfaction values to estimate what a steady-state bank account would probably look like for that pop. Or you could go for a more cynically realistic route and pay out interest to capital rather than private accounts -- paying only aristocrats and capitalists based on the profitability of the rgos and factories (representing how corporate accounts have a large role in the financial sector).

As a side note, some of the developers on discord got a bit excited about the "faster build target" we talked about. Is that still in the works?

PR #966, opened 2024-01-15
incrementalise the incremental build target and move template definition to header

yrrrmmbl:


schombert:
As the incremental target gets faster to build, you may also want to add some text to the guide for contributors explaining the difference between it and the release build and directing developers to use the incremental build for most work

PR #972, opened 2024-01-16
Showbattlesofothernations

wxwisiasdf:


schombert:
problems are: (a) who is red and who is blue, and (b) what happens when there is more than one battle happening in the province (c) how do we illustrate when the player's units are in the province but not part of the battle

PR #975, opened 2024-01-17
Map mode tooltips

CyberTropic:
Adds missing tooltips for the new mapmodes
 
Fort - Shows fort level and building options
Ideology - Shows ideologies by pop, same as pie chart
Income - Shows total income of all pops in state
Consciousness - Shows consciousness and what's affecting it, same as revolt risk with militancy
Militancy - N/A, it's duplicate of revolt risk mode
Literacy - Literacy of pops
Employment - Shows percentage employed. This is out of employable pops not total pops, which excludes soldiers etc. Changed the map mode to use employable as well
Factories - Shows number of factories in state as well, as well as factory types and their levels. Changed the map mode to handle factories in split states
Issues - Shows top 5 issues from all pops, same as pie chart
Growth - Shows monthly growth, map mode is two gradients with yellow fixed at 0, positive growth goes to green and negative growth goes to red

### DIFF
--- a/src/gui/map_tooltip.cpp
+++ b/src/gui/map_tooltip.cpp
@@ -937,14 +937,14 @@ void factory_map_tt_box(sys::state& state, text::columnar_layout& contents, dcon
 		}
 		std::sort(factories.begin(), factories.end(), [&](auto a, auto b) {return a.get_level() > b.get_level(); });
 
-		text::add_to_layout_box(state, contents, box, text::pretty_integer(int64_t(factories.size())), text::text_color::yellow);
+		text::add_to_layout_box(state, contents, box, text::format_wholenum(factories.size()), text::text_color::yellow);
 		for(size_t i = 0; i < factories.size(); i++) {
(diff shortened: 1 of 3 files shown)


src/gui/map_tooltip.cpp
```diff
@@ -792,82 +792,300 @@ void religion_map_tt_box(sys::state& state, text::columnar_layout& contents, dco
...
+		text::close_layout_box(contents, box);
+	}
+}
+
+void con_map_tt_box(sys::state& state, text::columnar_layout& contents, dcon::province_id prov) {
+	auto fat = dcon::fatten(state.world, prov);
+	country_name_box(state, contents, prov);
+
+	if(prov.value < state.province_definitions.first_sea_province.value) {
+		float average_con = (state.world.province_get_demographics(prov, demographics::consciousness) / state.world.province_get_demographics(prov, demographics::total));
```
schombert, line 877: you may want to add a check for 0 total population here. This can happen in some mods (for example, those with Antarctica), as well as from seeing incomplete data mid update

PR #976, opened 2024-01-17
Made AI avoid units when pathing (also applies to player units), of c…

wxwisiasdf:
…ourse, it avoids all units EXCEPT the one they are going for (if any), for example pathing into an unit in province Zacatecas will make the AI avoid ALL other units EXCEPT the one in Zacatecas, for obvious reasons

AI will now consider generales and army composition to determine attack and defense of each unit

TESTED

With TNO
With the og savefile where the AI suicides repeatedly
With vanilla
With TGC
and with HFM

AI now doesnt die as much and is much more conservative on when doing attacks, and when they do them they do them good

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -4209,36 +4209,110 @@ void gather_to_battle(sys::state& state, dcon::nation_id n, dcon::province_id p)
 	}
 }
 
-float estimate_army_strength(sys::state& state, dcon::army_id a) {
+float estimate_balanced_composition_factor(sys::state& state, dcon::army_id a) {
 	auto regs = state.world.army_get_army_membership(a);
 	if(regs.begin() == regs.end())
 		return 0.0f;
-	auto last_reg = regs.end() - 1;
-	float scale = state.world.army_get_controller_from_army_control(a) ? 1.f : 0.5f;
-	return float(regs.end() - regs.begin()) * (*last_reg).get_regiment().get_org() * scale;
+	// account composition
+	// Ideal composition: 4/1/4 (1 cavalry for each 4 infantry and 1 infantry for each arty)
+	float total_str = 0.f;
+	float str_art = 0.f;
+	float str_inf = 0.f;
+	float str_cav = 0.f;
+	for(const auto reg : regs) {
+		float str = reg.get_regiment().get_strength() * reg.get_regiment().get_org();
+		auto utid = reg.get_regiment().get_type();
+		if(utid) {
+			switch(state.military_definitions.unit_base_definitions[utid].type) {
+			case military::unit_type::infantry:
+				str_inf += str;
+				break;
+			case military::unit_type::cavalry:
+				str_cav += str;
+				break;
+			case military::unit_type::support:
+			case military::unit_type::special:
+				str_art += str;
+				break;
+			}
+		}
+		total_str += str;
+	}
+	if(total_str == 0.f)
+		return 0.f;
+	//provide continous function for each military unit composition
+	//such that 4x times the infantry (we min with arty for equality reasons) and 1/4th of cavalry
+	float scale = 1.f - math::sin(std::abs(std::min(str_art / total_str, str_inf / total_str) - (4.f * str_cav / total_str)));
+	return total_str * scale;
 }
 
-float conservative_estimate_army_strength(sys::state& state, dcon::army_id a) {
-	auto regs = state.world.army_get_army_membership(a);
-	if(regs.begin() == regs.end())
-		return 0.0f;
+float estimate_army_defensive_strength(sys::state& state, dcon::army_id a) {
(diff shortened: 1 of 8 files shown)


src/ai/ai.cpp
```diff
@@ -4209,49 +4209,121 @@ void gather_to_battle(sys::state& state, dcon::nation_id n, dcon::province_id p)
 	}
 }
 
-float estimate_army_strength(sys::state& state, dcon::army_id a) {
+float estimate_balanced_composition_factor(sys::state& state, dcon::army_id a) {
```
schombert, line 4212: Did you take a performance measurement for this? You have turned what was meant to be a cheap function call into a quite complex one?

src/ai/ai.cpp
```diff
@@ -4366,7 +4438,7 @@ void assign_targets(sys::state& state, dcon::nation_id n) {
 		if(!potential_targets[i].location)
 			continue; // target has been removed as too close by some earlier iteration
 
-		auto target_attack_force = estimate_attack_force(state, potential_targets[i].location, n);
```
schombert, line 4369: Here and in the location below, for example, this function may be called repeatedly on basically the same provinces. Is it necessary to cache the result now?

src/provinces/province.cpp
```diff
@@ -1714,8 +1714,9 @@ std::vector<dcon::province_id> make_land_path(sys::state& state, dcon::province_
 
 				if(other_prov.id.index() < state.province_definitions.first_sea_province.index()) { // is land
 					if(has_access_to_province(state, nation_as, other_prov)) {
+						float danger_factor = has_safe_access_to_province(state, nation_as, other_prov) ? 1.f : 4.f;
```
schombert, line 1717: this won't make it avoid enemy armies, just provinces that the enemy controls. Once inside enemy territory then (or outside it) it will be just as blind to enemy armies as before

src/ai/ai.cpp
wxwisiasdf: cached

src/provinces/province.cpp
wxwisiasdf: made it be wary of any armies not theirs

src/provinces/province.cpp
```diff
@@ -1714,7 +1714,9 @@ std::vector<dcon::province_id> make_land_path(sys::state& state, dcon::province_
 
 				if(other_prov.id.index() < state.province_definitions.first_sea_province.index()) { // is land
 					if(has_access_to_province(state, nation_as, other_prov)) {
-						float danger_factor = has_safe_access_to_province(state, nation_as, other_prov) ? 1.f : 4.f;
+						/* This will work fine for most instances, except, possibly, for allied nations or enemy ones */
+						auto armies = state.world.province_get_army_location(other_prov);
+						float danger_factor = (armies.begin() == armies.end() || (*armies.begin()).get_army().get_controller_from_army_control() == nation_as) ? 1.f : 4.f;
```
schombert, line 1719: this will result in trying to avoid allied and neutral armies as well

wxwisiasdf: attrition

src/culture/rebels.cpp
```diff
@@ -831,8 +831,8 @@ void rebel_hunting_check(sys::state& state) {
 				std::sort(rebel_hunters.begin(), rebel_hunters.end(), [&](dcon::army_id a, dcon::army_id b) {
 					auto pa = state.world.army_get_location_from_army_location(a);
 					auto pb = state.world.army_get_location_from_army_location(b);
-					auto as = 0.001f * std::max<float>(ai::estimate_army_strength(state, a), 1.f);
-					auto bs = 0.001f * std::max<float>(ai::estimate_army_strength(state, b), 1.f);
+					auto as = 0.001f * std::max<float>(ai::estimate_army_defensive_strength(state, a), 1.f);
```
schombert, line 834: now that these calls have become expensive, you probably want to cache this too

src/ai/ai.cpp
```diff
@@ -4392,7 +4475,7 @@ void assign_targets(sys::state& state, dcon::nation_id n) {
 					continue;
 				}
 
-				a_force_str += conservative_estimate_army_strength(state, ar.get_army());
+				a_force_str += estimate_army_offensive_strength(state, ar.get_army());
```
schombert, line 4478: un-cached

schombert:
In general, in the places where you might want to cache army strength, instead of a vector of armies / provinces, you could move to a vector of structs containing the army/province and the cached strength value (you could even store that value as 0 until the first access if you think you might not need it, but I think in the places where it is used it will always be used)

src/ai/ai.cpp
wxwisiasdf: were calculating how much attack force we have, which needs to account for leaders attack too

PR #981, opened 2024-01-17
Update build documentation to include information about the incremental build target

OptimizedForDensity:


schombert:
You could go with stronger language if you want. I would say that the incremental build is recommended for most developers. CI should catch any problems with the unity build.

OptimizedForDensity:
Good point about the CI.

PR #983, opened 2024-01-18
AI prioritize uncivilized targets for CBs 

wxwisiasdf:


src/ai/ai.cpp
```diff
@@ -4440,14 +4453,17 @@ void assign_targets(sys::state& state, dcon::nation_id n) {
 
 	// organize attack stacks
 	bool is_at_war = state.world.nation_get_is_at_war(n);
-	int32_t max_attacks_to_make = is_at_war ? (ready_count + 3) / 4 : ready_count; // not at war -- allow all stacks to attack rebels
+	int32_t max_attacks_to_make = is_at_war ? (ready_count + 1) / 3 : ready_count; // not at war -- allow all stacks to attack rebels
 	auto const psize = potential_targets.size();
 
+	for(uint32_t i = 0; i < psize; ++i) {
+		potential_targets[i].strength_estimate = estimate_enemy_defensive_force(state, potential_targets[i].location, n) + 0.00001f;
```
schombert, line 4460: why are you doing it all up here? The reason that it was delayed to below was that it is possible that not all of them will actually be looked at, so it is more efficient not to compute the strength unless we actually need that number

src/ai/ai.cpp
```diff
@@ -2048,15 +2048,28 @@ void update_cb_fabrication(sys::state& state) {
...
 						possible_targets.push_back(i.id);
+						if(!i.get_is_civilized())
+							possible_unciv_targets.push_back(i.id);
+					}
+				}
+				// prioritize uncivilized
+				if(n.get_is_civilized() && !possible_unciv_targets.empty()) {
+					auto t = possible_unciv_targets[rng::reduce(uint32_t(rng::get_random(state, uint32_t(n.id.index())) >> 2), uint32_t(possible_unciv_targets.size()))];
+					if(auto cb = pick_fabrication_type(state, n, t); cb) {
+						n.set_constructing_cb_target(t);
```
schombert, line 2066: ... and then this would be just overwritten by the code below. Even if it did work and you add an else or something, this would make nations target uncivs exclusively. And that is far from desired.

PR #994, opened 2024-01-20
Multiplayer complaints and extra ruleset for partial retreats

wxwisiasdf:


src/gamestate/commands.cpp
```diff
@@ -3141,12 +3141,82 @@ std::vector<dcon::province_id> can_move_army(sys::state& state, dcon::nation_id
...
+			if(state.world.army_get_black_flag(a)) {
+				return province::make_unowned_land_path(state, last_province, dest);
+			} else if(province::has_access_to_province(state, source, dest) && b_12 && b_13) {
+				return province::make_land_path(state, last_province, dest, source, a);
+			} else if(b_10) {
+				return province::make_unowned_land_path(state, last_province, dest);
+			} else {
+				return std::vector<dcon::province_id>{};
+			}
 		} else {
```
schombert, line 3212: The way you have re ordered the conditionals, I do not believe this is correct. I believe that you would get here if the army is in battle and you are not in multiplayer, which would allow partial retreats outside of multiplayer

wxwisiasdf: ``		if(state.network_mode != sys::network_mode_type::single_player
		&& state.world.army_get_battle_from_army_battle_participation(a)) {``

wxwisiasdf: ok fixes

PR #996, opened 2024-01-22
cascade fallbacking on launcher

wxwisiasdf:


schombert:
three problems. First the previous version where we actually try to launch the program is useful because if the particular exe doesn't exist (for example, when giving people demo releases) it will fall back to alice.exe eventually rather than just silently failing to launch and doing nothing. Second, I am not compiling that many executables every time I make a release. Third I don't want mulitplayer to fall back that far.

wxwisiasdf:
> three problems. First the previous version where we actually try to launch the program is useful because if the particular exe doesn't exist (for example, when giving people demo releases) it will fall back to alice.exe eventually rather than just silently failing to launch and doing nothing. Second, I am not compiling that many executables every time I make a release. Third I don't want mulitplayer to fall back that far.

1. What? this new code errors when the user user user alice, going from AVX512, AVX2, AVX, SSE4.2, SSE3 and then to the default Alice
2. It's fine
3. It needs to because most MPs and MP hosts still runs on Core 2 duos

schombert:
2. is not "fine" as for 3., get them to come talk to me and tell me their CPU specs. I am highly reluctant to somewhat cripple everyone's multiplayer experience for the sake of a few people. The core 2 duo is 17 years old at this point.

PR #1000, opened 2024-01-22
literally 1984. Display mobilization milscore as red

wxwisiasdf:


schombert:
change icon (occurs in 4 places in the ui) not text color

PR #1002, opened 2024-01-22
Update (and reorganize a bit) the documentation

wxwisiasdf:


wxwisiasdf:
Changes:

- Added XAC and XSM notes and general overview
- Merged both guides into one (multiplayer guide is now below user guide, for CONVENIENCE)
- did the same for spanish
- some lone german ocumentation was just standing on the english folder unattended, so i put it on the de-de folder like it should

PR #1008, opened 2024-01-23
Economy tweaks

ineveraskedforthis:
Removes most mentions of base prices from economy simulation, implements prices graph page in a ledger and changes most of constants to achieve more adaptive and smooth economy. Review of performance effects is required.


src/ai/ai.cpp
```diff
@@ -857,20 +859,36 @@ void get_desired_factory_types(sys::state& state, dcon::nation_id nid, std::vect
...
 		} // END if building unlocked
 	}
+	*/
+
 	if(desired_types.empty()) { // second pass: try to make money
 		for(auto type : state.world.in_factory_type) {
 			if(n.get_active_building(type) || type.get_is_available_from_start()) {
 				auto& inputs = type.get_inputs();
 				bool lacking_input = false;
+				bool lacking_output = n.get_demand_satisfaction(type.get_output()) < 1.02f;
```
schombert, line 869: unless you have changed something, demand satisfaction should always be in the range 0 - 1 (representing percent of the demand that could be purchased)

src/military/military.cpp
```diff
@@ -1021,12 +1021,18 @@ void regenerate_land_unit_average(sys::state& state) {
...
-		float total = 0;
-		float count = 0;
-
 		auto lo_mod = state.world.nation_get_modifier_values(n, sys::national_mod_offsets::land_attack_modifier);
 		auto ld_mod = state.world.nation_get_modifier_values(n, sys::national_mod_offsets::land_defense_modifier);
 
+		auto irregular = state.military_definitions.irregular;
+		auto& irreg_stats = state.world.nation_get_unit_stats(n, irregular);
+
+		// handle irregular infantry as it seems to be unhandled by code below...
```
schombert, line 1030: why do you believe that the loop below doesn't cover irregulars? An irregular is one of the unit types, and so it should appear as part of the iteration

src/economy/economy.hpp
```diff
@@ -114,7 +114,28 @@ constexpr inline dcon::commodity_id money(0);
 // to feed everyone, and *every* the rgos was grain (and everyone farmed), the scaling factor would have to be about 60'000
 // if all rgos were equally common (there are about 20 of them), the scaling factor would have to be about
 // 1'200'000. Assuming that grain is slightly more prevalent, we arrive at the factor below as a nice round number
-constexpr inline float needs_scaling_factor = 1'000'000.0f * 2.0f;
+//constexpr inline float needs_scaling_factor = 1'000'000.0f * 2.0f;
+constexpr inline float needs_scaling_factor = 100'000.0f * 1.0f;
+
+inline constexpr float production_scale_delta = 0.0001f;
+inline constexpr uint32_t price_history_lenght = 256;
```
schombert, line 121: spelling of length

src/military/military.cpp
ineveraskedforthis: For some reason, all unciv nations had 0 military score with undefined average stats before this change. Was this behavior unique to my copy?

src/gui/gui_ledger_window.hpp
```diff
@@ -1918,28 +1920,115 @@ class ledger_commodity_plupp : public tinted_image_element_base {
...
 protected:
 	std::string_view get_row_element_name() override {
 		return "ledger_legend_entry";
 	}
-
 public:
 	void on_update(sys::state& state) noexcept override {
 		row_contents.clear();
-		state.world.for_each_commodity([&](dcon::commodity_id id) { row_contents.push_back(id); });
+		state.world.for_each_commodity([&](dcon::commodity_id id) {
```
schombert, line 2029: I know that you didn't write this loop. But, as a comment on the original loop, since the commodities themselves don't change over the course of the game, the list could be generated just once when the element is created. Doesn't really matter I suppose

src/military/military.cpp
schombert: Ah,  yes, there is a bug in the body of the loop. It should check if the unit has been activated *or if it is active from the start of the game* (irregulars being one of the units covered by the second part of that check)

src/ai/ai.cpp
```diff
@@ -890,66 +908,74 @@ void update_ai_econ_construction(sys::state& state) {
...
+				if(apop != bpop)
+					return apop > bpop;
+				else
+					return a.index() < b.index();
+			});
 
-						if(pw_employed >= pw_num && pw_num > 0.0f)
-							continue; // no spare workers
+			// try to upgrade factories first:
+			if((rules & issue_rule::build_factory) == 0 && (rules & issue_rule::expand_factory) != 0) { // can't build -- by elimination, can upgrade
```
schombert, line 928: This condition will only pass if the nation does not have the ability to build new factories. IIRC your rational for moving this loop up was to prioritize upgrading, even for nations that could build, so you may want to remove the first part of the condition.

src/economy/economy.cpp
```diff
@@ -11,6 +11,39 @@
 
 namespace economy {
 
+enum class economy_reason {
+	pop, factory, rgo, artisan, construction, nation, stockpile, overseas_penalty
+};
+
+std::string reason_to_string(economy_reason reason) {
```
schombert, line 18: since these are all static constants, it may make more sense to return a `std::string_view`

src/economy/economy.cpp
```diff
@@ -124,116 +157,125 @@ void initialize_artisan_distribution(sys::state& state) {
...
+			if(state.world.commodity_get_is_luxury_need(c) && (state.world.commodity_get_is_available_from_start(c) || (kf && state.world.nation_get_active_building(n, kf)))) {
+				auto& w = state.world.nation_get_luxury_needs_weights(n, c);
+				w = 1.f;
+			}
+		});
+	}
+}
+
+float need_weight(sys::state& state, dcon::nation_id n, dcon::commodity_id c) {
+	return 1 / sqrt(state.world.commodity_get_current_price(c));
```
schombert, line 194: this may potentially be an issue down the line. Unless you know otherwise, my understanding is that the result of the sqrt function is not defined down to the last bit of precision between different compilers. To keep deterministic lockstep working, you probably need to use `math::sqrt` which, with the implementation we provide, should always give the same results, no matter where it is compiled.

src/economy/economy.cpp
```diff
@@ -341,6 +383,9 @@ void randomly_assign_artisan_production(sys::state& state, dcon::province_id p,
 float base_artisan_profit(sys::state& state, dcon::nation_id n, dcon::commodity_id c) {
 	auto& inputs = state.world.commodity_get_artisan_inputs(c);
 
+	auto fat_id = dcon::fatten(state.world, c);
+	auto name = text::get_name_as_string(state, fat_id);
```
schombert, line 387: this looks like a left over from a debugging aid and should be removed

src/economy/economy.cpp
```diff
@@ -396,6 +443,7 @@ void adjust_artisan_balance(sys::state& state, dcon::nation_id n) {
 
 				total_ex_weights += w;
 				temp_weights[cid.index()] = profit * profit;
+				profits[cid.index()] = profit;
```
schombert, line 446: you don't need to store both profits and temp weights. Just store the profit and then square it as needed when the temp weight was previously called for

src/economy/economy.cpp
```diff
@@ -639,14 +695,16 @@ void update_factory_triggered_modifiers(sys::state& state) {
...
 	auto c = state.world.province_get_rgo(p);
 	if(!c)
 		return 0.01f;
 
-	//Occupied provinces don't employ RGOs
+	float occupation_effect = 1.0f;
+	//Occupied provinces employ less RGOs
 	if(state.world.province_get_nation_from_province_ownership(p) != state.world.province_get_nation_from_province_control(p))
-		return 0.0f;
+		occupation_effect = 0.95f;
```
schombert, line 707: with the occupation effect being this close to 1.0, is it worth having it at all?

src/economy/economy.cpp
```diff
@@ -1236,6 +1362,41 @@ void populate_navy_consumption(sys::state& state) {
...
+inline constexpr float diff_factory = day_1_build_time_modifier_factory - day_inf_build_time_modifier_factory;
+inline constexpr float shift_factory = -diff_factory / day_1_derivative_factory;
+inline constexpr float slope_factory = diff_factory * shift_factory;
+
+// also we want to speed up factories construction right at the start
+// as it's the most vulnerable time for them
+// and we need to establish *some* industrial base for world to develop
+// their build time should also become faster with time to delay growth bottleneck
+float global_factory_construction_time_modifier(sys::state& state) {
+	float t = sqrt((static_cast<float>(state.current_date.value) * 0.01f + 2.f));
```
schombert, line 1396: see earlier note about sqrt

src/gamestate/system_state.cpp
```diff
@@ -3484,6 +3484,34 @@ void state::load_scenario_data(parsers::error_handler& err) {
...
+					world.commodity_get_rgo_amount(c) * modifer);
+			});
+
+			world.for_each_commodity([&](dcon::commodity_id c) {
+				world.commodity_set_current_price(c, world.commodity_get_cost(c));
+			});
+		}
+
+		//final adjustment of the economy
+		for(int i = 1; i < 10000; i++) {
```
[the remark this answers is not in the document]
OptimizedForDensity: If this is to run a simulation to tweak parameters, maybe it would be possible to have this as a separate debug feature that would run the simulation then output any of the parameters it's tweaking to a file. That way mod makers could do a big simulation run to set up the economy then provide those parameters to users so they don't have to run a simulation every time they build a scenario (might also help with reproducibility since everyone would have the same set of parameters in case simulations end up with slightly different results if there's any randomness in the economy and seeds).

OptimizedForDensity:
I played around with the PR a bit since I myself have been tweaking the economy, good work. A few things I noticed:
- Countries/pops collect money each economy tick so you might want to reset it at some point(s) if you're trying to get stable prices/production. You'll probably also want to reset pop/country money after you're done setting up the economy because the UK will start out with billions of pounds after 35 years of simulated tax collecting.
- Have you played around with using consumption rather than real demand to drive prices? Consumption is actualized demand which makes prices less likely to enter a negative feedback loop of high demand -> high prices -> reduced consumption but equal demand -> higher prices -> reduce consumption but equal demand -> etc.

ineveraskedforthis:
True, I should reset treasuries after "adjustments", thanks. Demand is already based on amount of wealth pops are "ready" to spend, so it hopefully shouldn't have a problem you are talking about. Consumption has an annoying feature of being always lower than production and it "forgets" income above costs of bought goods so no, I haven't played with it because properly calculated demand felt as a better candidate. I guess one could increase price when consumption is at `(1 - 1 / "market weight") * production` and play with ways to calculate "market weight".

PR #1010, opened 2024-01-23
Added option for enabling mouse edge scrolling

aintDatCap:
I'm not a GUI guru, someone has to fix that label


src/gui/gui_main_menu.cpp
```diff
@@ -141,7 +141,12 @@ void map_zoom_mode_display::on_update(sys::state& state) noexcept {
 		break;
 	}
 }
-
+void map_mouse_edge_scrolling::button_action(sys::state& state) noexcept {
+	state.user_settings.mouse_edge_scrolling = !state.user_settings.mouse_edge_scrolling;
```
schombert, line 145: just needs a `send(state, parent, notify_setting_update{});`

PR #1011, opened 2024-01-23
Added how many pops promote/demote each day

Wonobory:
as the title says

image
image


src/economy/demographics.cpp
```diff
@@ -1482,6 +1482,25 @@ float get_estimated_type_change(sys::state& state, dcon::pop_id ids) {
 			: (std::ceil(demotion_chance * state.defines.promotion_scale * current_size)));
 }
 
+void save_yesterday_pop_size(sys::state& state, dcon::nation_id nation)
+{
+	for(auto prov : state.world.nation_get_province_ownership(nation)) {
+		for(auto pop : prov.get_province().get_pop_location()) {
+			state.world.pop_set_yesterday_size(pop.get_pop(), state.world.pop_get_size(pop.get_pop()));
```
schombert, line 1489: there is more than one pop in a province, so what this will do is essentially save the size of one of the pops living in the province, chosen at more or less random

schombert:
Writing a value every day for all the pops belonging to the player's nation is too high of a price to pay to support this tooltip.

Wonobory:
would it be better if i try to make that it only saves yesterdays pop size only when the population window is open?

schombert:
Hmm. At least as I tend to do things, I usually have the game paused when I am going through the various windows. So for my usage habits, the tooltip wouldn't work at all that way because it wouldn't have a month to gather data over (this is also a problem with saving just the last day -- lots of pop changes run once a month, so in many days the change could be zero even if the pop is going to grow by thousands on its monthly growth tick, for example).

schombert:
closed for rennovations

PR #1021, opened 2024-01-25
create full crash dumps

wxwisiasdf:


schombert:
I don't see the point of this; I don't actually rebuild dbg_alice, and even if I did, you don't have the pdbs for release. Nevertheless, here you go.

PR #1023, opened 2024-01-25
Fix provid off by 1

CyberTropic:


schombert:
The displayed province id isn't meant to match the province id from the files; it is meant to match our internal province id so that people can use it in other cheat commands. Our province ids tend to match the file ids of provinces +1 for the early provinces, but in fact we do reorder the provinces so that all the land provinces come first, so that relationship tends to stop holding at some point as soon as the game files list a sea province prior to a land province.

wxwisiasdf:
literally 1984

PR #1024, opened 2024-01-25
fix negative sum or 0 sum causing ub

wxwisiasdf:


src/military/military.cpp
```diff
@@ -3934,7 +3934,7 @@ sys::date arrival_time_to(sys::state& state, dcon::army_id a, dcon::province_id
 	float distance = province::distance(state, adj);
 	float sum_mods = state.world.province_get_modifier_values(p, sys::provincial_mod_offsets::movement_cost) +
 									 state.world.province_get_modifier_values(p, sys::provincial_mod_offsets::movement_cost);
-	float effective_distance = distance * (sum_mods + 1.0f);
+	float effective_distance = std::max(0.001f, distance * (sum_mods + 1.0f));
 
 	float effective_speed = effective_army_speed(state, a);
```
schombert, line 3939: you should probably also make sure that effective army speed is > 0, if that isn't done already. Also you should make these changes to the navy version of the function, below

wxwisiasdf: done, and yes the effective_army_speed is > 0

PR #1025, opened 2024-01-25
Make AI be more prone to target countries with active rebellions

wxwisiasdf:


schombert:
I don't agree that this is a good idea. Rebels have very little effective strength. And if they are an actual problem, they will be killing the military units of the nation, thus signaling that it is an easy target by lowering its strength

PR #1039, opened 2024-01-26
Use blocking sockets

wxwisiasdf:
did you test it? yes we're literally testing it rn and its going fine


schombert:
this doesn't actually change the sockets to be blocking

schombert:
Also, you reported some new oos errors, so ...

schombert:
You need to remove this
```
ioctlsocket(socket_fd, FIONBIO, &mode);
```
which sets your socket to non blocking (or rather explicitly set it to 0 I guess, for "please block")

PR #1040, opened 2024-01-26
Use a random username for launcher

wxwisiasdf:


schombert:
if you aren't going to use a human-sensible default, than I am going to also require you to save and restore changes to the user name people make in the launcher

PR #1041, opened 2024-01-26
Use a random username for launcher

wxwisiasdf:


src/common_types/container_types.hpp
```diff
@@ -288,7 +288,7 @@ struct checksum_key {
 static_assert(sizeof(checksum_key) == sizeof(checksum_key::key));
 
 struct player_name {
-	char data[8];
+	char data[8] = { 0, 0, 0, 0, 0, 0, 0, 0 };
```
schombert, line 291: can't even fit "schombert" in that ... plus it will need even more room when it supports unicode

src/launcher/launcher_main.cpp
```diff
@@ -378,6 +378,17 @@ native_string produce_mod_path() {
 	return simple_fs::extract_state(dummy);
 }
 
+void save_playername() {
+	auto path = launcher::produce_mod_path();
+	auto game_state = std::make_unique<sys::state>();
```
schombert, line 383: No, this cannot be done, this is several hundred megabytes of stuff you are spinning up just to save a string to a file...

src/launcher/launcher_main.cpp
```diff
@@ -1799,8 +1813,15 @@ int WINAPI wWinMain(
 	char username[256 + 1];
 	DWORD username_len = 256 + 1;
 	GetComputerNameA(username, &username_len);
-	launcher::player_name = std::string(reinterpret_cast<const char*>(&username[0]));
-	//
+
+	// Load from user settings
+	{
+		auto path = launcher::produce_mod_path();
+		auto game_state = std::make_unique<sys::state>();
```
schombert, line 1820: see above

src/launcher/launcher_main.cpp
```diff
@@ -499,6 +508,7 @@ void mouse_click() {
 		return;
 	case ui_obj_play_game:
 		if(file_is_ready.load(std::memory_order::memory_order_acquire) && !selected_scenario_file.empty()) {
+			save_playername();
```
schombert, line 511: shouldn't you save it when it is changed, not every time the game is started?

wxwisiasdf: its easier this way - people often change their name before joining a game

schombert: It also means you are going to be repeatedly saving it when nothing has changed. It also means that you will be silently discarding changes unless someone launches a game after making them.

wxwisiasdf: ok fine

PR #1044, opened 2024-01-27
Stop crashed players from being controlled by AI

wxwisiasdf:


schombert:
so then how can players leave a game and turn over control to the ai?

wxwisiasdf:
> so then how can players leave a game and turn over control to the ai?

host kicks "dead" nation

schombert:
what about your automated server idea? No host there to do that manually. If a player crashes, they won't send an "i'm leaving" command to the host. If they drop without leaving, the host can close their connection without sending the notify player has left command if they want to keep the slot player controlled until someone else can fill it.

wxwisiasdf:
> what about your automated server idea? No host there to do that manually. If a player crashes, they won't send an "i'm leaving" command to the host. If they drop without leaving, the host can close their connection without sending the notify player has left command if they want to keep the slot player controlled until someone else can fill it.

they can simply be kicked, or the command can set them to "AI" by default if on headless mode

schombert:
I think having to kick any player that leaves is kind of unnecessary busy work. How about adding a bool to say whether the player should be set to AI on leaving in the command?

PR #1048, opened 2024-01-27
Display "READY" when READY in multiplayer

wxwisiasdf:


src/gui/gui_nation_picker.hpp
```diff
@@ -657,13 +657,15 @@ class multiplayer_status_text : public simple_text_element_base {
...
 						auto completed = c.total_sent_bytes - c.save_stream_offset;
 						auto total = c.save_stream_size;
 						float progress = (float(total) / float(completed));
-						text::substitution_map sub{};
-						text::add_to_substitution_map(sub, text::variable_type::value, text::fp_percentage_one_place{ progress });
-						set_text(state, text::produce_simple_string(state, text::resolve_string_substitution(state, "alice_status_stream", sub)));
+						if(progress < 1.f) {
+							text::substitution_map sub{};
+							text::add_to_substitution_map(sub, text::variable_type::value, text::fp_percentage_one_place{ progress });
+							set_text(state, text::produce_simple_string(state, text::resolve_string_substitution(state, "alice_status_stream", sub)));
```
schombert, line 667: Just a heads up, but you won't be able to call set text every frame like this once we move to unicode

wxwisiasdf: fine

PR #1051, opened 2024-01-27
Added instant_industry cheat and modified industry upgrade buttons

aintDatCap:


schombert:
It looks like this PR has accidentally changed the encoding of the alice.csv file

PR #1053, opened 2024-01-27
add extra controls to production window

wxwisiasdf:


src/gui/topbar_subwindows/gui_production_window.hpp
```diff
@@ -121,7 +121,7 @@ class factory_priority_button : public button_element_base {
 	}
 };
 
-class factory_upgrade_button : public shift_button_element_base {
+class factory_upgrade_button : public button_element_base {
```
schombert, line 124: if you change this to be just button element base, the button shift action below will never be called

wxwisiasdf: see the on_lbutton and on_rbutton

PR #1054, opened 2024-01-27
Hide country_flags by default (unless user wants to be spoilered, via…

wxwisiasdf:
… "show spoilers option)

Modders want it, so we must give it to them - and as a compromise, have it as a user toggleable on the misc menu (if the user so wishes)


schombert:
should default to showing them. And the description in the ui for this option should be just "show country flags in effects" so that we explain what it actually does

wxwisiasdf:
> should default to showing them. And the description in the ui for this option should be just "show country flags in effects" so that we explain what it actually does

that defeats the whole point :(

schombert:
that is the only way that I will merge this PR

PR #1068, opened 2024-01-28
set player on AI (with make_ai parameter for command too)

wxwisiasdf:
added make_ai parameter


schombert:
do clients have no way of notifying the host that they plan to leave and thus being disconnected "gracefully"?

wxwisiasdf:
> do clients have no way of notifying the host that they plan to leave and thus being disconnected "gracefully"?

theycanjustsend a command

wxwisiasdf:
command::notify_player_leaves(state, client.playing_as, true);

schombert:
Yes, but what I meant was, does that command every get sent if a client closes the game normally? (i.e. via closing the window or an esc > exit command)

PR #1069, opened 2024-01-28
fi mobilization progress bar

wxwisiasdf:


src/gui/topbar_subwindows/gui_military_window.hpp
```diff
@@ -38,45 +38,40 @@ class military_mob_button : public button_element_base {
...
 	void on_update(sys::state& state) noexcept override {
-		if(state.world.nation_get_is_mobilized(state.local_player_nation) == false) {
-			progress = 0.0f;
-			return;
+		auto n = state.local_player_nation;
+		int32_t used = 0;
+		int32_t total = 0;
+		for(const auto po : state.world.nation_get_province_ownership(n)) {
+			used += military::regiments_created_from_province(state, po.get_province());
+			total += military::regiments_max_possible_from_province(state, po.get_province());
```
schombert, line 46: Both of these functions are for proper soldiers, and won't reflect mobilization at all. To test your fix, hit the mobilize button. The progress bar should start at 0% and should advance over time to 100%, and the point that it hits 100% should be the point where new units stop appearing.

wxwisiasdf: fxed

src/gui/topbar_subwindows/gui_military_window.hpp
```diff
@@ -42,8 +42,8 @@ class military_mob_progress_bar : public progress_bar {
 		int32_t used = 0;
 		int32_t total = 0;
 		for(const auto po : state.world.nation_get_province_ownership(n)) {
-			used += military::regiments_created_from_province(state, po.get_province());
-			total += military::regiments_max_possible_from_province(state, po.get_province());
+			used += military::mobilized_regiments_created_from_province(state, po.get_province());
+			total += military::mobilized_regiments_possible_from_province(state, po.get_province());
```
schombert, line 46: I am afraid that this won't work either. There can be more mobilized regiments possible from the provinces collectively than the nation is allowed to raise, and thus the mobilization progress bar will never hit 100% with this change.

PR #1078, opened 2024-01-29
AI do aggressive foreign investment

wxwisiasdf:
Instead of hoarding money, it will use it to inflate its own industrial score
It doesn't care what it's building, it just needs to throw money to get free industrial points.
It's a big gamey, but it benefits the target because they get industrialized fast (assuming they are able to afford the factories)


wxwisiasdf:
With new economy, now AI helps buff and stabilize prices accordingly

PR #1089, opened 2024-01-30
Implemented innovate cheat

aintDatCap:


aintDatCap:
is there a way to avoid that compilation-breaking warning?

src/gui/gui_console.cpp
```diff
@@ -1571,6 +1575,27 @@ void ui::console_edit::edit_box_enter(sys::state& state, std::string_view s) noe
...
 		break;
 	}
+	case command_info::type::innovate:
+	{
+		auto searched_name = std::get<std::string>(pstate.arg_slots[0]);
+		std::replace(searched_name.begin(), searched_name.end(), '_', ' ');
+		bool found = false;
+		for(auto invention: state.world.in_invention) {
+			auto innovation_name = text::produce_simple_string(state, invention.get_name());
+			std::transform(innovation_name.begin(), innovation_name.end(), innovation_name.begin(),[](unsigned char c) { return std::tolower(c); });
```
schombert, line 1585: re: your comment -- you probably just need a cast to char here, since tolower returns an int (and also, be prepared for it to break in interesting ways post transition to unicode, since tolower is an ascii function)

PR #1095, opened 2024-02-04
that would be nice

wxwisiasdf:


src/gui/topbar_subwindows/gui_production_window.hpp
```diff
@@ -1019,15 +1019,18 @@ class production_factory_info_bounds_window : public window_element_base {
...
 			auto ptr = make_element_by_type<production_factory_info>(state,
 					state.ui_state.defs_by_name.find("factory_info")->second.definition);
 			ptr->index = factory_index;
-			ptr->base_data.position.x = factory_index * ptr->base_data.size.x;
+			ptr->base_data.position.x = (factory_index % 8) * ptr->base_data.size.x;
+			ptr->base_data.position.y += (factory_index / 8) * (ptr->base_data.size.y - 26);
 			infos.push_back(ptr.get());
 			add_child_to_front(std::move(ptr));
 		}
+		base_data.size.y += state.ui_defs.gui[state.ui_state.defs_by_name.find("factory_info")->second.definition].size.y * (uint8_t(state.defines.factories_per_state) / 8);
```
schombert, line 1033: this will always increase the size, even when the number of factories per state is left at the default value

schombert:
Also, things are going to round down. 10 factories will end up with the same space as 8. Test both the default 8 factories and 10 factories to make sure that everything is working as expected.

PR #1110, opened 2024-02-05
Stop factory_type from being uint16_t

wxwisiasdf:


schombert:
peter needed more for his mod; check with him to see if 125 is enough

PR #1115, opened 2024-02-09
fix cb window being buggy overall

Nivaturimika:


assets/alice.csv
```diff
@@ -888,6 +888,7 @@ fab_explain_3;You are not currently fabricating a casus belli
 fab_explain_4;You are an independent nation or the target is your overlord
 fab_explain_5;The target is not in your sphere of influence
 fab_explain_6;You are not currently at war with the target
+fab_explain_7;You have any fabricable casus belli
```
schombert, line 891: Suggest wording this as "You can fabricate a casus belli"

PR #1131, opened 2024-02-24
Dynamic localisation for AHMP

Nivaturimika:
in vanilla yoiu can rename states to do this
however thats a terrible trick
i see this more useful for whatever thing we wish to perform :3

### DIFF
--- a/ParserGenerator/parser_generator.cpp
+++ b/ParserGenerator/parser_generator.cpp
@@ -117,7 +117,7 @@ size_t get_column(std::string_view const s, std::string_view::iterator const it)
 }
 
 bool is_ident(char c) {
-	return std::isalnum(c) || c == '#' || c == '@' || c == '_';
+	return std::isalnum(c) || c == '#' || c == '@' || c == '_' || c == '[' || c == ']';
 }
 
 void tokenize_line(std::string_view const line) {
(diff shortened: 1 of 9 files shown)


Makkkkus:
This way of adding scripted localisation seems a bit hack-y. [On the forum](https://forum.paradoxplaza.com/forum/threads/grand-jomini-modding-information-manuscript.1170261/) Paradox has a neat article about how different features work, including dynamic description. The main benefit is a cleaner syntax and theoretically endless descriptions.

Nivaturimika:
Doing that would require an overhaul of the existing effect/trigger system -- which will come later on but by then OpenVic would've released v0.1, so not very near at all.

This is the best that we can do with the current infrastructure.

Nivaturimika:
> I see; although I am worried about backwards-compatability if were to change it in the future.

That's a valid concern, however we'll eventually figure things out - afterall AHMP is the only mod at the moment making use of the Alice engine at it's fullest

Makkkkus:
That's reassuring. I think in this case it would be good to add some context around that in the extensions.md file.

schombert:
This seems like a poor way of doing things. First, present a clear use case. Your description in the modding extensions doc seems like it would be better satisfied by just having two events -- one for each case. Once we have some real examples of where an extension is necessary, we can think about the best way to make them work (keeping in mind the document I posted in translations, since they will need to work in that system too). If you really want arbitrary text generation, then you would want to embed a proper scripting language to run when the text is needed. If you want something less, then maybe something less than that can be added

schombert:
I feel like, given how janky this proposed system is, the redundant events are better

Nivaturimika:
> I feel like, given how janky this proposed system is, the redundant events are better

It doesn't affect performance nor hinders any other mechanic EXCEPT just UI flavour... how is this janky?

Nivaturimika:
Have a better system in mind then?

Nivaturimika:
And no I don't mean the new UI, it would be very hard to adapt the existing text to the new format already as-is, unless you provide tools for doing so

PR #1143, opened 2024-03-02
add reform desc

Nivaturimika:
A small addition for AHMP flavour

### DIFF
--- a/src/gui/topbar_subwindows/politics_subwindows/gui_reforms_window.hpp
+++ b/src/gui/topbar_subwindows/politics_subwindows/gui_reforms_window.hpp
@@ -101,13 +101,15 @@ void reform_description(sys::state& state, text::columnar_layout& contents, dcon
 	auto reform = fatten(state.world, ref);
 
 	text::add_line(state, contents, reform.get_name());
-	text::substitution_map sub{};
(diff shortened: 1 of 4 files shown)


src/gui/topbar_subwindows/politics_subwindows/gui_reforms_window.hpp
```diff
@@ -100,6 +100,15 @@ void reform_rules_description(sys::state& state, text::columnar_layout& contents
 void reform_description(sys::state& state, text::columnar_layout& contents, dcon::issue_option_id ref) {
 	auto reform = fatten(state.world, ref);
 
+	text::add_line(state, contents, reform.get_name());
```
schombert, line 103: you need to handle the case where no name has been set (here and elsewhere) or you may mess up text for the base game

schombert: In particular, you are always forcing a ..._desc key to be created, so that will show up for most mods

Nivaturimika: basegame does define the name for it tho

PR #1157, opened 2024-03-24
Fix events not properly doing ai_chance for empty effects on options

Nivaturimika:


schombert:
a) you would need to adjust the corresponding loops below each of these. b) the issue was not with events that had a null-effect option

schombert:
I dont remember if effect::execute checks if the effect is a valid handle. If it does not, you will have to add the condition back in around the execution

PR #1169, opened 2024-03-25
merge rally point checkboxes will display why no merge rally points s…

Nivaturimika:
…o people stop complaining


schombert:
encoding of the csv file got messed up

schombert:
not by you

schombert:
When I added the merge, github fkkd the encoding

Nivaturimika:
what do we do

schombert:
roll back my merge and then merge locally so that github doesn't have a chance to mess the text up

Nivaturimika:
Did this fix it?

schombert:
I don't see any changes to the csv file now, so the new text about the buttons being disabled won't be included I think

Nivaturimika:
> I don't see any changes to the csv file now, so the new text about the buttons being disabled won't be included I think

fiuxes

PR #1177, opened 2024-03-25
explain why AI will not accept a demand ol white peace

Nivaturimika:


src/gui/topbar_subwindows/diplomacy_subwindows/gui_peace_window.hpp
```diff
@@ -282,6 +282,7 @@ class diplomacy_peace_send : public button_element_base {
 			text::add_line(state, contents, "ai_will_accept_po");
 		} else {
 			text::add_line(state, contents, "ai_will_not_accept_po");
+			text::add_line(state, contents, "alice_ai_behaviour_on_demand_po");
```
schombert, line 285: This will display the message unconditionally, regardless of whether it is relevant. Plus, the AI will absolutely accept a white peace demand in the right circumstances

Nivaturimika: > This will display the message unconditionally, regardless of whether it is relevant. Plus, the AI will absolutely accept a white peace demand in the right circumstances

Only if the AI is saying no

PR #1187, opened 2024-04-03
Toggle daily oos checks on console (default is OFF)

Nivaturimika:


schombert:
First, this should be a transient property that you can turn on, not a setting that gets saved permanently. Secondly, will this work if the server and clients don't have the option set the same way?

Nivaturimika:
> First, this should be a transient property that you can turn on, not a setting that gets saved permanently. Secondly, will this work if the server and clients don't have the option set the same way?

a) i disagree
b) yes it will

schombert:
I'm not going to merge it in if the setting is part of the user settings file. You know from experience how that file can end up corrupted / giving people random default settings. And giving people the daily oos check randomly would degrade their mp experience and they wouldn't know why and would then complain to everyone about how mp is slow/bad

Nivaturimika:
fair, closing and making new pr making it transient [l8tr]

PR #1194, opened 2024-04-04
AI improve behaviour on PO of crisis

Nivaturimika:
- AI will now join crisis with colonial concessions and such
- Will also send peace offers accordingly at max crisis temperature, because:
- a) Going to war will make them lose more than they win (given the economy)
- b) If the other side accepts, then it's best for the AI
- c) If the other side declines, we go to war faster, indeed, but if we are superior in strength and we demanded terms, we will likely win
- d) And if not, then it's probably because the other side wants to go to war anyways so yeah

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -1630,13 +1630,13 @@ void update_crisis_leaders(sys::state& state) {
 
 		bool defender_victory = str_est.attacker < str_est.defender * 0.66f;
 		if(state.world.nation_get_is_player_controlled(state.primary_crisis_attacker) == false) {
-			assert(command::can_start_crisis_peace_offer(state, state.primary_crisis_attacker, true));
-			command::execute_start_crisis_peace_offer(state, state.primary_crisis_attacker, true);
+			assert(command::can_start_crisis_peace_offer(state, state.primary_crisis_attacker, defender_victory));
+			command::execute_start_crisis_peace_offer(state, state.primary_crisis_attacker, defender_victory);
 			auto pending = state.world.nation_get_peace_offer_from_pending_peace_offer(state.primary_crisis_attacker);
 			for(auto& par : state.crisis_participants) {
 				if(!par.id)
 					break;
-				bool side_ood = !defender_victory ? !par.supports_attacker : par.supports_attacker;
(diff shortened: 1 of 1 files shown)


src/ai/ai.cpp
```diff
@@ -1536,9 +1536,8 @@ bool will_join_crisis_with_offer(sys::state& state, dcon::nation_id n, sys::cris
 	if(offer.target == state.world.nation_get_ai_rival(n))
 		return true;
 	auto offer_bits = state.world.cb_type_get_type_bits(offer.wargoal_type);
-	if((offer_bits & (military::cb_flag::po_demand_state | military::cb_flag::po_annex)) != 0)
+	if((offer_bits & (military::cb_flag::po_colony | military::cb_flag::po_demand_state | military::cb_flag::po_annex)) != 0)
```
schombert, line 1539: po_colony isn't place in the sun; that is also po_demand_state. po_colony is the special crisis cb that removes competitors from a potential colony IIRC. thus it can never appear in a third party offer in the first place

schombert:
Your changes look like they will prevent the AI from trying to WP a crisis

schombert:
Also, make sure that the logic for accepting crisis offers is also updated to match. Otherwise the AI will just be sending offers that won't be accepted

Nivaturimika:
> Your changes look like they will prevent the AI from trying to WP a crisis

it will try to WP if the other side hasn't added any wargoals, and we are losing

> Also, make sure that the logic for accepting crisis offers is also updated to match. Otherwise the AI will just be sending offers that won't be accepted

it uses the same "win losing" logic to determine if accepting or not

schombert:
Yes, but your logic means that the only outcomes for the crisis phase is that side A gets everything or side B gets everything. It isn't possible, if both A and B have something on the table (e.g. a colonial crisis or where the defender has made a wg offer) for the crisis leaders to back down into a white peace rather than one side surrendering.

Nivaturimika:
> Yes, but your logic means that the only outcomes for the crisis phase is that side A gets everything or side B gets everything. It isn't possible, if both A and B have something on the table (e.g. a colonial crisis or where the defender has made a wg offer) for the crisis leaders to back down into a white peace rather than one side surrendering.

thats the idea

PR #1196, opened 2024-04-04
fix duplicate code for taxes in nations

Nivaturimika:


src/nations/nations.cpp
```diff
@@ -1066,7 +1066,7 @@ float tariff_efficiency(sys::state& state, dcon::nation_id n) {
 
 float tax_efficiency(sys::state& state, dcon::nation_id n) {
 	auto eff_mod = state.world.nation_get_modifier_values(n, sys::national_mod_offsets::tax_efficiency);
-	return eff_mod + state.defines.base_country_tax_efficiency;
+	return std::clamp(state.defines.base_country_tax_efficiency + eff_mod, 0.01f, 1.f);
```
schombert, line 1069: you have changed a lower limit of 0.1 to 0.01. 10% is as low as tax efficiency is supposed to get

Nivaturimika: fixed

PR #1204, opened 2024-04-04
display accurate % of mobilization

Nivaturimika:


src/gui/topbar_subwindows/gui_military_window.hpp
```diff
@@ -44,38 +44,60 @@ class military_mob_button : public right_click_button_element_base {
...
+		}
+		auto schedule_array = state.world.nation_get_mobilization_schedule(n);
+		int32_t rem_provinces = int32_t(schedule_array.size());
+
+		if(total_provinces == 0) {
+			progress = 1.f;
 			return;
 		}
-		progress = std::min(1.0f, mob_rem / mob_size);
+		progress = float(total_provinces - rem_provinces) / float(total_provinces);
```
schombert, line 70: this will show the percentage of provinces remaining, not the percentage of units remaining

Nivaturimika: It's still more accurate than before

Nivaturimika: Also the mobilization logic makes it so the mobilized regiments spawn immediately when the province is "no longer remaining"

PR #1211, opened 2024-04-05
Show wargoal button in red if it could exceed infamy limit

Nivaturimika:


src/gui/topbar_subwindows/diplomacy_subwindows/gui_pick_wargoal_window.hpp
```diff
@@ -32,6 +32,11 @@ class wargoal_type_item_button : public button_element_base {
 
 	void on_update(sys::state& state) noexcept override {
 		dcon::cb_type_id content = retrieve<dcon::cb_type_id>(state, parent);
+		color = sys::pack_color(255, 255, 255);
+		if(state.world.nation_get_infamy(state.local_player_nation) + military::cb_infamy(state, content) > state.defines.badboy_limit) {
```
schombert, line 36: you need to use the cb_addition_infamy_cost function to account for the possibility of great war infamy discounts

Nivaturimika: Done, albeit only IFF there is a war

PR #1212, opened 2024-04-05
map zoom speed setting

Nivaturimika:


src/map/map_state.cpp
```diff
@@ -370,7 +370,7 @@ void map_state::update(sys::state& state) {
 
 	auto zoom_diff = (zoom_change * seconds_since_last_update) / (1 / zoom);
 	zoom += zoom_diff;
-	zoom_change *= std::exp(-seconds_since_last_update * 20);
+	zoom_change *= std::exp(-seconds_since_last_update * state.user_settings.zoom_speed);
```
schombert, line 373: test for isfinite and then clamp between two reasonable values in case of a junk settings value

Nivaturimika: clamped on load, does this do kosher?

src/gamestate/system_state.cpp
```diff
@@ -2274,6 +2276,7 @@ void state::load_user_settings() {
 		user_settings.gaussianblur_level = std::clamp(user_settings.gaussianblur_level, 1.0f, 1.25f);
 		user_settings.gaussianblur_level = std::clamp(user_settings.gaussianblur_level, 1.0f, 1.5f);
 		user_settings.gamma = std::clamp(user_settings.gamma, 0.5f, 2.5f);
+		user_settings.zoom_speed = std::clamp(user_settings.zoom_speed, 15.f, 25.f);
```
schombert, line 2279: clamp will not necessarily work if the value is a NaN

Nivaturimika: my bad, and while we are at it, did it for the rest of float settings too

PR #1215, opened 2024-04-05
Cursor for whena  move is valid, for when a move is on enemy territor…

Nivaturimika:
…y, when the game is busy doing some shit, etc

### DIFF
--- a/src/window/window_win.cpp
+++ b/src/window/window_win.cpp
@@ -450,8 +450,11 @@ void change_cursor(sys::state const& state, cursor_type type) {
 		if(auto f = simple_fs::peek_file(cursors_dir, fname); f) {
 			auto path = simple_fs::get_full_name(*f);
 			state.win_ptr->cursors[uint8_t(type)] = LoadCursorFromFileW(path.c_str()); //.cur or .ani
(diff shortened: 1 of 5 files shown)


src/window/window_win.cpp
```diff
@@ -427,6 +417,47 @@ void create_window(sys::state& game_state, creation_parameters const& params) {
...
+			break;
+		default:
+			fname = NATIVE("normal.cur");
+			break;
+		}
+		if(auto f = simple_fs::peek_file(cursors_dir, fname); f) {
+			auto path = simple_fs::get_full_name(*f);
+			state.win_ptr->cursors[uint8_t(type)] = LoadCursorFromFileW(path.c_str()); //.cur or .ani
+		} else {
+			return; //can't load
```
schombert, line 454: there is an issue here: if the cursor file is missing, then your code will hit the file system each frame looking for it, which will be extremely slow. You need to instead store a default cursor or some other indicator not to go looking for it.

Nivaturimika: Ok schowomby

Nivaturimika: fixed

PR #1216, opened 2024-04-05
Deafen on focus lost

Nivaturimika:


schombert:
the word you are looking for is "mute" (stop making sound). Deafen would mean that the program stops responding to audio input

Nivaturimika:
> the word you are looking for is "mute" (stop making sound). Deafen would mean that the program stops responding to audio input

Expix told me to, so i complied, because expix is jesus christ v2.0

PR #1222, opened 2024-04-06
Topbar notification windows

Nivaturimika:


schombert:
I would like to have the messages go directly to pop-ups, as they currently do, as an option at least. Personally I don't want to have to do a bunch of extra work to resolve them

Nivaturimika:
I will make an option that will be ON by default if thats kosher

src/ai/ai.cpp
```diff
@@ -160,8 +163,11 @@ void form_alliances(sys::state& state) {
 					else
 						return a.index() > b.index();
 				});
-				nations::make_alliance(state, n, alliance_targets[0]);
-
+				if(state.world.nation_get_is_player_controlled(alliance_targets[0])) {
+					command::execute_ask_for_alliance(state, n, alliance_targets[0]);
```
schombert, line 167: the alliance target list should never have human-controlled nations in it

Nivaturimika: AI can now ask player for alliance, which will make life infinitely easier for most players
and given that its not as spammy as popups, it will be better for people using the topbar thing above

PR #1226, opened 2024-04-06
Console open sounds

Nivaturimika:


src/gui/gui_console.cpp
```diff
@@ -1654,8 +1654,21 @@ void ui::console_text::render(sys::state& state, int32_t x, int32_t y) noexcept
...
+	ui::console_window::show_toggle(state);
+}
+void ui::console_edit::edit_box_backtick(sys::state& state) noexcept {
+	ui::console_window::show_toggle(state);
+}
+
 void ui::console_window::show_toggle(sys::state& state) {
 	assert(state.ui_state.console_window);
+	if(state.ui_state.console_window->is_visible()) { //close
+		sound::play_interface_sound(state, sound::get_console_close_sound(state), state.user_settings.master_volume * state.user_settings.interface_volume);
```
schombert, line 1667: problem is that when this is called from a button command, the button will already be playing the default button sound. This will then interrupt that sound, which may sound weird. You would need to create a new button class without the sound effect / with this sound effect instead to avoid that.

Nivaturimika: it is never called from a button command, its either tilde or edit_backtick/edit_esc

PR #1227, opened 2024-04-06
Properly say disabled/enabled on console to prevent confusion

Nivaturimika:


src/gui/gui_console.cpp
```diff
@@ -1486,9 +1486,17 @@ void ui::console_edit::edit_box_enter(sys::state& state, std::string_view s) noe
...
 		break;
-	case command_info::type::instant_research:
+	case command_info::type::instant_research: {
+		auto has_us = false;
+		for(const auto n : state.cheat_data.instant_research_nations)
+			if(n == state.local_player_nation) {
+				has_us = true;
+				break;
+			}
+		log_to_console(state, parent, !has_us ? "\xA7GEnabled" : "\xA7RDisabled");
```
schombert, line 1496: shouldn't this come from the localization csv file?

Nivaturimika: logging to console works a bit different and cant make use of localization csv

Nivaturimika: atleast not w/o discarding colour info

schombert: yes you can; you can grab the csv entry and convert it to a string and then pass the string to the log to console function

Nivaturimika: the idea is to have COLOR

Nivaturimika: Now that I think of it, it would be more efficient to make use of the checkmark/cross symbols we already have

schombert: what is wrong with "\xA7G" + to_string(csv_key) ?

Nivaturimika: > what is wrong with "\xA7G" + to_string(csv_key) ?

a checkmark will do then

PR #1230, opened 2024-04-06
Automatically discard messages that can no longer be accepted

Nivaturimika:


src/gui/gui_diplomacy_request_topbar.hpp
```diff
@@ -121,8 +121,10 @@ class diplomatic_message_topbar_listbox : public overlapping_listbox_element_bas
 public:
 	std::vector< diplomatic_message::message> messages;
 	void on_update(sys::state& state) noexcept override {
-		auto it = std::remove_if(messages.begin(), messages.end(),
-				[&](auto& m) { return m.when + diplomatic_message::expiration_in_days <= state.current_date; });
+		auto it = std::remove_if(messages.begin(), messages.end(), [&](auto& m) {
+			return m.when + diplomatic_message::expiration_in_days <= state.current_date
+				|| !diplomatic_message::can_accept(state, m);
+		});
 		auto r = std::distance(it, messages.end());
```
schombert, line 128: why is r calculated?

src/gui/gui_diplomacy_request_window.hpp
```diff
@@ -337,8 +337,10 @@ class diplomacy_request_window : public window_element_base {
 	}
 
 	void on_update(sys::state& state) noexcept override {
-		auto it = std::remove_if(messages.begin(), messages.end(),
-				[&](auto& m) { return m.when + diplomatic_message::expiration_in_days <= state.current_date; });
+		auto it = std::remove_if(messages.begin(), messages.end(), [&](auto& m) {
+			return m.when + diplomatic_message::expiration_in_days <= state.current_date
+				|| !diplomatic_message::can_accept(state, m);
+		});
 		auto r = std::distance(it, messages.end());
```
schombert, line 344: and here

src/gui/gui_diplomacy_request_topbar.hpp
Nivaturimika: wtf is r calculated

schombert: why is `std::distance(it, messages.end());` called since it doesn't appear that anything is done with the result

Nivaturimika: fixed

PR #1233, opened 2024-04-06
Stylize budget tab to be like an accounting report

Nivaturimika:
- use () for negative numbers
- colour every number, in dark tones of course :)

### DIFF
--- a/src/gui/topbar_subwindows/gui_budget_window.hpp
+++ b/src/gui/topbar_subwindows/gui_budget_window.hpp
@@ -1096,7 +1096,7 @@ public:
 
 		if(total < 0.0f) {
 			color = text::text_color::dark_red;
-			set_text(state, std::string("(") + text::format_money(-total) + std::string(")"));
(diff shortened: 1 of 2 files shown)


schombert:
(...) for negative numbers is something that only appears in certain locales. It isn't a thing for US English

Nivaturimika:
> (...) for negative numbers is something that only appears in certain locales. It isn't a thing for US English

this isn't because of language, this is for the accounting LARP, trust me

schombert:
no, see for example https://learn.microsoft.com/en-us/windows/win32/intl/locale-ineg-constants

Nivaturimika:
> no, see for example https://learn.microsoft.com/en-us/windows/win32/intl/locale-ineg-constants

this is the same reason like why vanilla uses diacritics... on the english language
it's for immersion and gameplay stuff, not for usefulness
just trust me it's a good addition that some players will appreciate

schombert:
poll the discord

PR #1237, opened 2024-04-07
fix war reparations income not registering

Nivaturimika:


src/economy/economy.cpp
```diff
@@ -3981,17 +3981,15 @@ float estimate_war_subsidies_income(sys::state& state, dcon::nation_id n) {
...
-			if(uni.get_reparations()) {
-				auto source = uni.get_source();
-				auto const tax_eff = nations::tax_efficiency(state, n);
-				auto total_tax_base = state.world.nation_get_total_rich_income(source) +
-					state.world.nation_get_total_middle_income(source) +
-					state.world.nation_get_total_poor_income(source);
-				auto payout = total_tax_base * tax_eff * state.defines.reparations_tax_hit;
-				total += payout;
-			}
+	for(auto uni : state.world.nation_get_unilateral_relationship_as_target(n)) {
```
schombert, line 3984: you should double check the loop that makes the payments too

Nivaturimika: the loop that makes the payments is OK, it was just the estimation FOR INCOME that was wrong
assumedly by copy pasting code without changing a bit of the behaviour first

PR #1250, opened 2024-04-07
Mark if we are interested in alliance offers from a given nation

Nivaturimika:


src/ai/ai.cpp
```diff
@@ -160,11 +165,11 @@ void form_alliances(sys::state& state) {
 					else
 						return a.index() > b.index();
 				});
-				nations::make_alliance(state, n, alliance_targets[0]);
-				// Call our new allies into wars.... they may not accept but they may just may join!
-				//for(auto wp : state.world.nation_get_war_participant(n))
-				//	if(!military::are_allied_in_war(state, n, alliance_targets[0]) && will_join_war(state, alliance_targets[0], wp.get_war(), wp.get_is_attacker()))
-				//		command::execute_call_to_arms(state, n, alliance_targets[0], wp.get_war());
+				if(state.world.nation_get_is_player_controlled(alliance_targets[0])) {
+					command::execute_ask_for_alliance(state, n, alliance_targets[0]);
```
schombert, line 169: asking is redundant here. Since we can only get here if the player has said that they are interested in the alliance, you should instead just make the alliance and send a notification that it has been made, rather than asking.

PR #1263, opened 2024-04-13
Reset event window position

Nivaturimika:
when a new event spawns, resetting the window position is useful as to not confuse the user why some windows appear moved


schombert:
add a check to make sure that the window isn't already open. You wouldn't want to jump a window back to the middle that someone has intentionally moved out of the way

Nivaturimika:
added

PR #1264, opened 2024-04-13
stop same ideology rebels to spawn against ruler

Nivaturimika:


schombert:
Hmm, not sure if this is a good change. Some of the rebels also want to change the government type. They still may want to overthrow a monarchy for a democracy, for example, even if the government has a liberal appointed party. So you should also check if the rebels winning will or won't do anything

PR #1271, opened 2024-04-13
limit ip address field to something more reasonable

Nivaturimika:
image
avoid this


schombert:
a valid ipv6 address can be up to 45 characters

PR #1272, opened 2024-04-13
Prevent crashes due to recursion stackoverflows

Nivaturimika:


schombert:
no effect can call itself directly. If this is caused by effects triggering events, and this is a real concern, I think it would be better (and more performant) to put those events into a delay buffer to be triggered in their own stack frame (i.e. like the events that get delayed by some days, but to fire in the same day)

Nivaturimika:
> no effect can call itself directly. If this is caused by effects triggering events, and this is a real concern, I think it would be better (and more performant) to put those events into a delay buffer to be triggered in their own stack frame (i.e. like the events that get delayed by some days, but to fire in the same day)

wouldnt that mess up order of execution?

schombert:
it might, but thinking more about your solution here, I don't believe it would even solve the problem as-is, because the trip through the event -> ai logic to pick a choice would reset the stack depth counter in any case

PR #1273, opened 2024-04-13
Fix recursion issues on events that trigger other events, by delaying…

Nivaturimika:
… them

### DIFF
--- a/src/scripting/effects.cpp
+++ b/src/scripting/effects.cpp
@@ -10,8 +10,14 @@
 
 namespace effect {
 
+struct effect_context {
+	std::vector<event::pending_human_n_event> future_n_event;
+	std::vector<event::pending_human_p_event> future_p_event;
+	bool else_if = false;
+};
+
 #define EFFECT_PARAMTERS                                                                                                         \
-	uint16_t const *tval, sys::state &ws, int32_t primary_slot, int32_t this_slot, int32_t from_slot, uint32_t r_hi, uint32_t r_lo, bool& els
+	uint16_t const *tval, sys::state &ws, int32_t primary_slot, int32_t this_slot, int32_t from_slot, uint32_t r_hi, uint32_t r_lo, effect_context& els
 
 uint32_t internal_execute_effect(EFFECT_PARAMTERS);
 
@@ -33,16 +39,16 @@ uint32_t es_generic_scope(EFFECT_PARAMTERS) {
 
 uint32_t es_if_scope(EFFECT_PARAMTERS) {
 	uint32_t ret = 0;
-	els = false;
+	els.else_if = false;
 	if((tval[0] & effect::scope_has_limit) != 0) {
 		auto limit = trigger::payload(tval[2]).tr_id;
 		if(trigger::evaluate(ws, limit, primary_slot, this_slot, from_slot)) {
 			ret = apply_subeffects(tval, ws, primary_slot, this_slot, from_slot, r_hi, r_lo, els);
-			els = true;
+			els.else_if = true;
 		}
 	} else {
 		ret = apply_subeffects(tval, ws, primary_slot, this_slot, from_slot, r_hi, r_lo, els);
-		els = true;
+		els.else_if = true;
 	}
 	return ret;
 }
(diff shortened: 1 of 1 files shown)


schombert:
We do already have a delayed event mechanism for events that are delayed at least 1 day. Is it possible to make immediate events use that same mechanism (i.e. give them a delay of 0 days)? This might make draining the queue more complicated of course ...

src/scripting/effects.cpp
```diff
@@ -5477,14 +5499,26 @@ uint32_t internal_execute_effect(EFFECT_PARAMTERS) {
...
 void execute(sys::state& state, uint16_t const* data, int32_t primary, int32_t this_slot, int32_t from_slot, uint32_t r_lo,
 		uint32_t r_hi) {
-	bool els = false;
+	effect_context els;
 	internal_execute_effect(data, state, primary, this_slot, from_slot, r_lo, r_hi, els);
+	for(const auto& p : els.future_n_event) {
+		event::trigger_national_event(state, p.e, p.n, p.r_lo, p.r_hi, p.from_slot, p.ft);
+	}
+	for(const auto& p : els.future_p_event) {
+		event::trigger_provincial_event(state, p.e, p.p, p.r_lo, p.r_hi, p.from_slot, p.ft);
```
schombert, line 5520: so, this won't work. Suppose that one of the stored provincial events fires a national event (I have seem mods do this). That event will now never fire since it is added to the future_n_event list which will no longer be checked at this point.

Nivaturimika:
> We do already have a delayed event mechanism for events that are delayed at least 1 day. Is it possible to make immediate events use that same mechanism (i.e. give them a delay of 0 days)? This might make draining the queue more complicated of course ...

we do, but the idea is to allow events to be fired on the same day instead of using a mechanism MEANT for delayaed events
FOR EXAMPLE
Events that are used as a "FAQ", they require to be immediately fired because the user needs to read stuff, and waiting 1 day for it is going to break a LOT OF MODS

src/scripting/effects.cpp
schombert: I also don't like how we are creating new vectors (and hence new allocations) constantly. We probably can't just make the variable static because of the possibility of recursion, but I wish something could be done here.

schombert:
> > We do already have a delayed event mechanism for events that are delayed at least 1 day. Is it possible to make immediate events use that same mechanism (i.e. give them a delay of 0 days)? This might make draining the queue more complicated of course ...
> 
> we do, but the idea is to allow events to be fired on the same day instead of using a mechanism MEANT for delayaed events FOR EXAMPLE Events that are used as a "FAQ", they require to be immediately fired because the user needs to read stuff, and waiting 1 day for it is going to break a LOT OF MODS

I mean that we could accommodate a delay of 0 days, so that the event would fire on that same say

src/scripting/effects.cpp
Nivaturimika: uhm, excuse me? it does work

Nivaturimika: check the logic above for pushing events to the queue

schombert: I did check the logic

schombert: Province event in future_p_event adds an event to the future_n_event vector. That vector is never checked again. Hence the event is never fired

Nivaturimika: i dont get what you mean then, the events do fire because their everything is pushed to the local effect context

schombert: oh, so each gets its own local effect context ... then this doesn't fix the recursion problem. event fires -> local context -> fires event -> sub local context -> etc

Nivaturimika: > Province event in future_p_event adds an event to the future_n_event vector. That vector is never checked again. Hence the event is never fired

Ok that is wrong
a) the effect within the event will have ITS OWN effect context
b) so this will occur:

event 100 gets fired (province)
ok good, now we execute the effect of event 100
we execute whatever, immediate or option 0 etc, MAKING OUR OWN LOCAL EFFECT CONTEXT
this event fires event 500 (national)
this EFFECT LOCAL CONTEXT is independent from the topmost one, it will execute event 500

Nivaturimika: effect_context is LOCAL to each execution unit - basically i made it so events can recursively call other events, but with a specific delay

schombert: Yeah, I see what you have done now. But the problem is that it doesn't fix the recursion isue

PR #1277, opened 2024-04-14
AI will no longer cheat upgrades

Nivaturimika:
AI having an advantage over players is unacceptable
Additionally, we make AI choose the BEST available units
Not objectively the best, but the ones with most attack
So guards and such can be built if needed
And airplanes and armored units too! which makes stomping on the AI 100 easy because THEY NEVER BUILD THEM HOLY SHIT

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -8,6 +8,7 @@
 #include "politics.hpp"
 #include "prng.hpp"
 #include "province_templates.hpp"
+#include "nations_templates.hpp"
 #include "triggers.hpp"
 
 namespace ai {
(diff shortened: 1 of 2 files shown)


schombert:
I am not convinced your picks for "best" are actually best. Also, if you want the AI to upgrade, it has to be able to slowly replace units (i.e intentionally disband). Otherwise out of date units will just clog up its available slots. But it can't disband all at once or it will leave itself defenseless when an upgrade is available.

Nivaturimika:
> I am not convinced your picks for "best" are actually best. Also, if you want the AI to upgrade, it has to be able to slowly replace units (i.e intentionally disband). Otherwise out of date units will just clog up its available slots. But it can't disband all at once or it will leave itself defenseless when an upgrade is available.

I made some changes
a) Restored free upgrades (one day we should add a mechanism to upgrade units for the player too?)
b) Best pickings now account attack, defense, siege, and speed
c) Limited upgrades to units according to regiment source (i.e you cant upgrade a ukrainian to a guard as russia - that would INDEED be cheating)

PR #1286, opened 2024-04-14
Money formatting fixups

Nivaturimika:


src/text/text.cpp
```diff
@@ -773,22 +773,25 @@ std::string prettify_currency(float num) {
...
-				snprintf(buffer, sizeof(buffer), sufx_one[i], reduced);
+				snprintf(buffer + 1, sizeof(buffer) - 1, sufx_one[i], reduced);
 			} else {
-				snprintf(buffer, sizeof(buffer), sufx_zero[i], reduced);
+				snprintf(buffer + 1, sizeof(buffer) - 1, sufx_zero[i], reduced);
+			}
+			if(num >= 0) {
+				return std::string(buffer + 1);
 			}
+			buffer[0] = '-'; //negative case
```
schombert, line 790: this is going to overwrite the first digit with the negative sign

src/text/text.cpp
```diff
@@ -773,22 +773,25 @@ std::string prettify_currency(float num) {
...
 	};
 	char buffer[200] = { 0 };
-	double dval = double(num);
+	double dval = std::abs(double(num));
 	for(size_t i = std::extent_v<decltype(mag)>; i-- > 0;) {
-		if(std::abs(dval) >= mag[i]) {
-			auto reduced = num / mag[i];
+		if(dval >= mag[i]) {
+			auto reduced = dval / mag[i];
 			if(reduced < 10.0) {
```
schombert, line 780: this is ugly. Instead of writing to one after the start of the buffer, just test abs(reduced)

PR #1287, opened 2024-04-14
Add merge buttons back to non-owhner and other's province views

Nivaturimika:


schombert:
I don't think rally points on provinces you don't own will work ... (they will show up, but units won't path to them)

Nivaturimika:
> I don't think rally points on provinces you don't own will work ... (they will show up, but units won't path to them)

they are disabled like in vanilla

PR #1289, opened 2024-04-14
Away from cursor panning option

Nivaturimika:


schombert:
away from cursor is an option that no sane person would want, as far as I can tell

Nivaturimika:
> away from cursor is an option that no sane person would want, as far as I can tell

oh gee you never know!

Nivaturimika:
also this pr fixes to_cursor

schombert:
its fine to fix it, but don't add a setting that someone will accidentally set and then complain to us about

PR #1303, opened 2024-04-16
Fix high spending and admin causing users accostumed to the normal ec…

Nivaturimika:
…onomy become confused


src/parsing/parsers_declarations.cpp
```diff
@@ -597,21 +597,21 @@ void poptype_file::equivalent(association_type, std::string_view value, error_ha
 void poptype_file::life_needs(commodity_array const& value, error_handler& err, int32_t line, poptype_context& context) {
 	context.outer_context.state.world.for_each_commodity([&](dcon::commodity_id cid) {
 		if(cid.index() < value.data.ssize())
-			context.outer_context.state.world.pop_type_set_life_needs(context.id, cid, value.data[cid]);
```
schombert, line 600: doing that here will affect how it shows up in the ingame ui describing needs, which is not desirable

PR #1349, opened 2024-04-21
Fix OOB parsing on TNO

Nivaturimika:


src/gamestate/system_state.cpp
```diff
@@ -2773,9 +2773,8 @@ void state::load_scenario_data(parsers::error_handler& err) {
 					break;
 				}
 			}
-			if(last - start_of_name >= 6 && file_name.ends_with(NATIVE("_oob.txt"))) {
```
schombert, line 2776: IIRC this won't work because the base game has a junk file in the directory that doesn't belong to any country

schombert: just tell the tno mod to name its oob files correctly

Nivaturimika: which junk file

schombert: recreate the vanilla scenario and see if it throws out an error. There is a reason I added the _oob filter

PR #1355, opened 2024-04-22
Province names

Nivaturimika:


schombert:
why not a proper saved user setting?

Nivaturimika:
> why not a proper saved user setting?

just let it be transient for now

PR #1375, opened 2024-04-27
Improvements to borders

Nivaturimika:


assets/shaders/glsl/text_line_f.glsl
```diff
@@ -38,5 +38,6 @@ void main() {
 		float t = max(0.0f, color_in.r * 16.0f - 7.0f);
 		frag_color = vec4(outer_color, t * t * t);
 	}
+	frag_color.a *= 0.8f;
```
schombert, line 41: you should just make the source dds more transparent. By doing it this way you are removing the possibility of someone choosing in a mod not to have transparent borders.

Nivaturimika: this is for text labels

Nivaturimika: the reason is to make them **slightly** more transparent, but just very very slightly
it is mainly a small detail that people like for some reason
but i didnt make them transparent that it would be hard to see, just transparent so that the people whi liked that effect wont be mad

PR #1380, opened 2024-04-28
new unit icons

Nivaturimika:


schombert:
why are these being added to the main assets folder?

schombert:
and why are you removing the bronze effect from the outlines?

Nivaturimika:
> and why are you removing the bronze effect from the outlines?

because it looks like garbage

PR #1394, opened 2024-04-30
fix GFM renaming decision

Nivaturimika:


schombert:
I don't see a fix to the bug where it was producing the incorrect size, only a change to the tests. How was the bug fixed?

tests/triggers_tests.cpp
```diff
@@ -513,33 +513,39 @@ TEST_CASE("effect scope absorbsion", "[effect_tests]") {
...
 		t.push_back(uint16_t(effect::clr_global_flag));
 		t.push_back(uint16_t(4));
 		t.push_back(uint16_t(effect::clr_global_flag));
 		t.push_back(uint16_t(7));
 		const auto new_size = parsers::simplify_effect(t.data());
-		REQUIRE(3 == new_size);
-		REQUIRE(t[0] == uint16_t(effect::fop_clr_global_flag_2));
-		REQUIRE(t[1] == uint16_t(4));
-		REQUIRE(t[2] == uint16_t(7));
+		REQUIRE(5 == new_size);
```
schombert, line 523: you should test that the scope contains the correct body size here as well

src/parsing/effect_parsing.cpp
```diff
@@ -339,6 +340,9 @@ void ef_scope_random_owned(token_generator& gen, error_handler& err, effect_buil
 
 		context.compiled_effect[payload_size_offset] = uint16_t(context.compiled_effect.size() - payload_size_offset);
 		context.limit_position = old_limit_offset;
+
+		auto const new_scope_size = simplify_effect(context.compiled_effect.data() + scope_offset);
```
schombert, line 344: why are you running simplify effect on its own here? doesn't it get run on all effects as a last step anyways? if so, this should be unnecessary if it is working correctly

src/parsing/effect_parsing.cpp
```diff
@@ -358,6 +362,9 @@ void ef_scope_random_owned(token_generator& gen, error_handler& err, effect_buil
 	context.compiled_effect[payload_size_offset] = uint16_t(context.compiled_effect.size() - payload_size_offset);
 	context.limit_position = old_limit_offset;
 	context.main_slot = old_main;
+
+	auto const new_scope_size = simplify_effect(context.compiled_effect.data() + scope_offset);
```
schombert, line 366: see above

src/parsing/effect_parsing.cpp
```diff
@@ -1164,12 +1171,42 @@ int32_t simplify_effect(uint16_t* source) {
...
-
-				if(new_size != old_size) { // has been simplified, assumes that new size always <= old size
+				// [0      0     ] | clr_global_flag	[1      2     ] | clr_global_flag
+				// [0      1     ] | <flag>				[1      3     ] | <flag>
+				// --------------- | n = 1				...............
+				// [n      2n    ]
+				// [n      2n + 1]
+				int32_t old_size = 1 + effect::get_generic_effect_payload_size(sub_units_start);
+				int32_t new_size = simplify_effect(sub_units_start);
+				if((sub_units_start[0] & effect::code_mask) == effect::clr_global_flag) {
```
schombert, line 1182: what happened to the clr_country_flag simplification?

src/parsing/effect_parsing.cpp
```diff
@@ -1164,12 +1171,42 @@ int32_t simplify_effect(uint16_t* source) {
...
+						effect::fop_clr_global_flag_2,
+						effect::fop_clr_global_flag_3,
+						effect::fop_clr_global_flag_4,
+						effect::fop_clr_global_flag_5,
+						effect::fop_clr_global_flag_6,
+						effect::fop_clr_global_flag_7,
+						effect::fop_clr_global_flag_8
+					};
+					sub_units_start[0] = fop_table[repeats];
+					new_size = 1 + effect::data_sizes[fop_table[repeats]];
```
schombert, line 1201: There is a bug here. If there are other, non clear global flag effects after the sequence of global flag effects, this new size will chop them all off (you should add such an effect to one of your tests to make sure that it doesn't happen)

tests/triggers_tests.cpp
```diff
@@ -510,6 +510,83 @@ TEST_CASE("effect scope absorbsion", "[effect_tests]") {
...
+	}
+	{
+		std::vector<uint16_t> t;
+		t.push_back(uint16_t(effect::generic_scope));
+		t.push_back(uint16_t(7));
+		t.push_back(uint16_t(effect::clr_global_flag));
+		t.push_back(uint16_t(4));
+		t.push_back(uint16_t(effect::clr_global_flag));
+		t.push_back(uint16_t(7));
+		t.push_back(uint16_t(effect::clr_global_flag));
```
schombert, line 536: add a non-clr global flag effect before and after this run of three to make sure that the simplification isn't interfering with other effects in the same scope

src/parsing/effect_parsing.cpp
Nivaturimika: optimizes random_owned so the overall size is smaller, because scopes store sizes in uint16_t, and at the time we finish the "main" scope, we already surpassed that limit - this is a workaround to make random_owned smaller

schombert: right, but as your tests revealed, simplify *breaks* if you run it on a sequence of effects that are not in a scope, and that isn't that exactly what you are doing here?

PR #1400, opened 2024-05-02
Fix event notification edge case

Nivaturimika:
If the event had an ai_chance, it would not spawn a notification about the national event

### DIFF
--- a/src/scripting/events.cpp
+++ b/src/scripting/events.cpp
@@ -100,38 +100,6 @@ void trigger_national_event(sys::state& state, dcon::national_event_id e, dcon::
 	if(auto immediate = state.world.national_event_get_immediate_effect(e); immediate) {
 		effect::execute(state, immediate, primary_slot, trigger::to_generic(n), from_slot, r_lo, r_hi);
 	}
-	if(state.world.nation_get_is_player_controlled(n)) {
-		pending_human_n_event new_event{r_lo, r_hi + 1, primary_slot, from_slot, e, n, state.current_date, pt, ft};
-		state.pending_n_event.push_back(new_event);
-		if(n == state.local_player_nation)
-			state.new_n_event.push(new_event);
-	} else {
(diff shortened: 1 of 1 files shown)


src/scripting/events.cpp
```diff
@@ -115,24 +115,21 @@ void trigger_national_event(sys::state& state, dcon::national_event_id e, dcon::
...
-						return;
+					if(rvalue < 0.0f && opt[i].effect) {
+						effect::execute(state, opt[i].effect, primary_slot, trigger::to_generic(n), from_slot, r_lo, r_hi + 1);
+						break;
 					}
 				}
 			}
-		}
-
-		if(opt[0].effect) {
```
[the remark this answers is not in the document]
schombert: also, it is outside the if to make potential rounding errors result in option 0 being chosen

src/scripting/events.cpp
```diff
@@ -115,24 +115,21 @@ void trigger_national_event(sys::state& state, dcon::national_event_id e, dcon::
...
-			effect::execute(state, opt[0].effect, primary_slot, trigger::to_generic(n), from_slot, r_lo, r_hi + 1);
+		} else {
```
schombert, line 129: still have an extra else here

src/scripting/events.cpp
```diff
@@ -212,24 +209,21 @@ void trigger_national_event(sys::state& state, dcon::free_national_event_id e, d
...
+						effect::execute(state, opt[i].effect, trigger::to_generic(n), trigger::to_generic(n), 0, r_lo, r_hi + 1);
+						break;
 					}
 				}
 			}
-		}
-
-		if(opt[0].effect) {
-			effect::execute(state, opt[0].effect, trigger::to_generic(n), trigger::to_generic(n), 0, r_lo, r_hi + 1);
+		} else {
```
schombert, line 223: and here

PR #1402, opened 2024-05-02
Make inventions active at start date

Nivaturimika:
like vanilla

### DIFF
--- a/src/gamestate/system_state.cpp
+++ b/src/gamestate/system_state.cpp
@@ -2487,7 +2487,6 @@ void state::load_scenario_data(parsers::error_handler& err) {
 	});
 
 	// load province history files
-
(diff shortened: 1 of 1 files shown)


src/gamestate/system_state.cpp
```diff
@@ -3004,8 +2997,12 @@ void state::load_scenario_data(parsers::error_handler& err) {
 	}
 	for(auto t : world.in_invention) {
 		for(auto n : world.in_nation) {
-			if(n.get_active_inventions(t))
+			if(trigger::evaluate(*this, t.get_limit(), trigger::to_generic(n), trigger::to_generic(n), -1)) {
```
schombert, line 3000: this would produce the incorrect results if the chance of discovering the invention is still 0 or less, which may be used to keep inventions locked in some cases

PR #1404, opened 2024-05-02
Event window for ELECTIONS

Nivaturimika:


src/gamestate/dcon_generated.txt
```diff
@@ -4616,6 +4621,11 @@ object {
 		name{ auto_choice }
 		type{ uint8_t }
 	}
+	property{
```
schombert, line 4624: election windows shouldn't occur free

src/gamestate/dcon_generated.txt
```diff
@@ -4516,6 +4516,11 @@ object {
...
```
schombert, line 4519: We don't need to store this for every event. Instead you can just store it for the set of election events (i.e. store a struct in the vector of election events instead with the issue id in there as well)

PR #1412, opened 2024-05-02
Immediate provincial support

Nivaturimika:


src/parsing/nations_parsing.cpp
```diff
@@ -947,6 +947,7 @@ void scan_province_event(token_generator& gen, error_handler& err, scenario_buil
 		auto new_id = context.state.world.create_free_provincial_event();
 		auto fid = fatten(context.state.world, new_id);
 		fid.set_description(event_result.desc_);
+		fid.set_immediate_effect(event_result.immediate_);
```
schombert, line 950: you need to make sure that the effects are created to take a province-province-approriate-from-slot. The existing logic wasn't expecting them to be added.

Nivaturimika: see lines above

src/gui/gui_event.cpp
```diff
@@ -518,6 +518,26 @@ void event_desc_text::on_update(sys::state& state) noexcept {
...
+			for(auto& l : delegate->internal_layout.contents) {
+				if(l.color == (delegate->black_text ? text::text_color::white : text::text_color::black)) //Invert colours
+					l.color = delegate->black_text ? text::text_color::black : text::text_color::white;
+			}
+		}
+	} else if(std::holds_alternative<event::pending_human_f_p_event>(content)) {
+		auto phe = std::get<event::pending_human_f_p_event>(content);
+		auto imm = state.world.free_provincial_event_get_immediate_effect(phe.e);
+		if(imm) {
+			effect_description(state, contents, imm, trigger::to_generic(phe.p), trigger::to_generic(phe.n), -1, phe.r_lo, phe.r_hi);
```
schombert, line 535: why is the this slot different from the main slot here?

Nivaturimika: because i am an idiot

PR #1435, opened 2024-05-06
Fix dismantlement event not triggering

Nivaturimika:


src/scripting/events.cpp
```diff
@@ -115,7 +115,7 @@ void trigger_national_event(sys::state& state, dcon::national_event_id e, dcon::
 			n, dcon::nation_id{}, dcon::nation_id{},
 			sys::message_base_type::major_event
 		});
-	} else if(n == state.local_player_nation) {
```
schombert, line 118: this and the one following was on purpose to prevent slowing down the update with notification sync. It isn't trivial

src/scripting/events.cpp
```diff
@@ -282,29 +282,27 @@ void trigger_provincial_event(sys::state& state, dcon::provincial_event_id e, dc
 		effect::execute(state, immediate, trigger::to_generic(p), trigger::to_generic(p), from_slot, r_lo, r_hi);
 	}
 	auto owner = state.world.province_get_nation_from_province_ownership(p);
-	if(owner == state.local_player_nation) {
```
schombert, line 285: and here

src/scripting/events.cpp
```diff
@@ -353,29 +351,27 @@ void trigger_provincial_event(sys::state& state, dcon::free_provincial_event_id
 		effect::execute(state, immediate, trigger::to_generic(p), trigger::to_generic(p), 0, r_lo, r_hi);
 	}
 	auto owner = state.world.province_get_nation_from_province_ownership(p);
-	if(owner == state.local_player_nation) {
```
schombert, line 356: and here

PR #1458, opened 2024-05-10
state_instance will always have uniform ownership over provinces

Nivaturimika:


schombert:
I can't tell what the actual changes here are among all the formatting changes. Can you explain them / point them out to me?

Nivaturimika:
image

src/scripting/effects.cpp
```diff
@@ -4065,130 +3917,98 @@ uint32_t ef_war_no_ally_from_province(EFFECT_PARAMTERS) {
...
 uint32_t ef_province_event_this_pop(EFFECT_PARAMTERS) {
 	auto postpone = int32_t(tval[2]);
-	if(postpone > 0) {
-		auto future_date = ws.current_date + postpone;
-		ws.future_p_event.push_back(event::pending_human_p_event {r_lo + 1, r_hi, this_slot, trigger::payload(tval[1]).pev_id, trigger::to_prov(primary_slot), future_date, event::slot_type::pop});
-	} else {
-		event::trigger_provincial_event(ws, trigger::payload(tval[1]).pev_id, trigger::to_prov(primary_slot), r_lo + 1, r_hi, this_slot,
-				event::slot_type::pop);
-	}
+	assert(postpone > 0);
```
schombert, line 4004: why the assert here? I believe that sending an event with a delay of 0 is legal, just redundant

src/scripting/effects.cpp
```diff
@@ -4065,130 +3917,98 @@ uint32_t ef_war_no_ally_from_province(EFFECT_PARAMTERS) {
...
 uint32_t ef_country_event_this_nation(EFFECT_PARAMTERS) {
 	auto postpone = int32_t(tval[2]);
-	if(postpone > 0) {
-		auto future_date = ws.current_date + postpone;
-		ws.future_n_event.push_back(event::pending_human_n_event {r_lo + 1, r_hi, primary_slot, this_slot, trigger::payload(tval[1]).nev_id, trigger::to_nation(primary_slot), future_date, event::slot_type::nation, event::slot_type::nation});
-	} else {
-		event::trigger_national_event(ws, trigger::payload(tval[1]).nev_id, trigger::to_nation(primary_slot), r_lo + 1, r_hi, this_slot,
-				event::slot_type::nation);
-	}
+	assert(postpone > 0);
```
schombert, line 3920: oh, there are a lot of these added asserts ... see comment on the last one

src/scripting/effects.cpp
```diff
@@ -1056,25 +945,21 @@ uint32_t es_x_substate_scope(EFFECT_PARAMTERS) {
 		}
 		return 0;
 	} else {
-		std::vector<dcon::nation_id> nlist;
-
+		uint32_t i = 0;
```
schombert, line 948: I don't understand this change. This case is supposed to handle picking a random substate, and it appears to no longer do that

src/scripting/effects.cpp
Nivaturimika: Exactly, we want to avoid such redundancies

Nivaturimika: This assert is mainly "the behaviour will be ok, and redundant, but the generation of this bytecode is concerning"

src/scripting/effects.cpp
Nivaturimika: random handled above

src/scripting/effects.cpp
schombert: you can emit a warning on parsing if you want, but an assert is not desirable, as it will cause a crash in debugging

src/scripting/effects.cpp
schombert: ok

src/scripting/effects.cpp
Nivaturimika: thats the idea mister

schombert: ok, so remove the asserts please, because I run in debug mode to, well, debug things, and those asserts would interfere with that

PR #1470, opened 2024-05-11
Make economy "easier" to manage by default

Nivaturimika:


schombert:
Tweaking the economy is reasonable. Making everyone need only 1 / 1000 of their normal needs is not, and will just trivialize things

PR #1478, opened 2024-05-12
Stop abortions, do std::exit instead (AFTER emitting the error message box)

Nivaturimika:


Nivaturimika:
Reason:
std::terminate calls std::abort
this generates a crash dump
this is bad because users confuse this with a crash
it is not a crash
we just failed to load something but we did not crash

PR #1490, opened 2024-05-12
Delete ships if landlocked

Nivaturimika:


src/nations/nations.cpp
```diff
@@ -215,6 +215,15 @@ bool are_allied(sys::state& state, dcon::nation_id a, dcon::nation_id b) {
 	return state.world.diplomatic_relation_get_are_allied(rel);
 }
 
+bool is_landlocked(sys::state& state, dcon::nation_id n) {
+	for(const auto pc : state.world.nation_get_province_ownership_as_nation(n)) {
```
schombert, line 219: you don't need this loop; you can just check if the nations total_ports is > 0

PR #1498, opened 2024-05-13
fix wrs, fix parsing bugs, fix outliner showing irrelevant countries

Nivaturimika:


src/gui/gui_outliner_window.hpp
```diff
@@ -696,8 +696,21 @@ class outliner_listbox : public listbox2_base<outliner_data> {
 		if(get_filter(state, outliner_filter::gp_influence)) {
 			row_contents.push_back(outliner_filter::gp_influence);
 			auto old_size = row_contents.size();
-			state.world.nation_for_each_gp_relationship_as_great_power(state.local_player_nation,
-					[&](dcon::gp_relationship_id grid) { row_contents.push_back(grid); });
+			state.world.nation_for_each_gp_relationship_as_great_power(state.local_player_nation, [&](dcon::gp_relationship_id grid) {
+				switch(state.world.gp_relationship_get_status(grid) & nations::influence::level_mask) {
+				case nations::influence::level_neutral:
```
schombert, line 701: this is not going to make people happy; people want the nations that they are influencing to show up there, even if they are currently at neutral or less. I also note that fixing whatever was going wrong with influence in gfm should also help

src/military/military.cpp
```diff
@@ -509,6 +509,18 @@ bool are_in_common_war(sys::state const& state, dcon::nation_id a, dcon::nation_
 	return false;
 }
 
+void remove_from_common_allied_wars(sys::state& state, dcon::nation_id a, dcon::nation_id b) {
+	for(auto wa : state.world.nation_get_war_participant(a)) {
+		auto is_attacker = wa.get_is_attacker();
+		for(auto o : wa.get_war().get_war_participant()) {
+			if(o.get_nation() == b) {
+				military::remove_from_war(state, wa.get_war(), a, false);
```
schombert, line 517: this is not going to work properly; you are changing the contexts of the war participation set while you are iterating over it. you can think of this as UB. IIRC the actual effect will be that you will miss some wars

PR #1504, opened 2024-05-14
ticking warscore tooltip

Nivaturimika:


src/gui/topbar_subwindows/gui_diplomacy_window.hpp
```diff
@@ -52,8 +52,7 @@ class cb_wargoal_button : public tinted_button_element_base {
 
 	void button_action(sys::state& state) noexcept override {
 		const dcon::cb_type_id content = retrieve<dcon::cb_type_id>(state, parent);
-		Cyto::Any newpayload = element_selection_wrapper<dcon::cb_type_id>{ content };
```
schombert, line 55: you have changed the underlying call from get to send. is that intentional?

Nivaturimika: yeh?

schombert: so the button either wasn't working before or it won't be working now?

Nivaturimika: It's the same, send uses impl get internally, this is just for ease of reading

PR #1533, opened 2024-05-17
[WIP] multilanguage

Nivaturimika:
do not merge, leaving this here for ref


schombert:
first comment pre merge: clang build needs to be fixed

src/text/text.cpp
```diff
@@ -741,31 +899,31 @@ std::string prettify_currency(float num) {
...
 	};
 	constexpr static char const* sufx_two[] = {
-		"%.2f \xA4",
-		"%.2fK \xA4",
-		"%.2fM \xA4",
-		"%.2fB \xA4",
-		"%.2fT \xA4",
-		"%.2fP \xA4",
-		"%.2fZ \xA4"
+		"%.2f £",
```
schombert, line 902: you may want to write these string literals as `u8"..."` to force utf8 encoding

schombert:
You have the wrong default font for Chinese. You want https://fonts.google.com/noto/specimen/Noto+Serif+SC

assets/languages.txt
```diff
@@ -0,0 +1,6 @@
+zh-CN;chinese
+ar-AR;arabic
+jp-JP;japan
+ko-SK;korean
+ko-NK;korean
+bg-BG;cyrillic
```
schombert, line 6: cyrillic is not the name of a language. What you have here is Bulgarian. Also, you should write the name of the languages in their native language, so that when someone who barely speaks english is fumbling their way through the setting they can immediately see which is their native language

assets/localisation/alice.csv
```diff
@@ -1,38 +1,60 @@
-alice_save_exit;Save and Exit;;;Salir y Guardar;;;;;;;;;x
-alice_save;Save;;;Guardar;;;;;;;;;x
-alice_exit;Exit;;;Salir;;;;;;;;;x
+language_en-US;English (US)
+language_fr-FR;French (France)
```
schombert, line 2: See previous comment about displaying language names. In any case, displaying the language name should bypass the localization system entirely because it should always display with its native name, not translated.

src/gui/gui_nation_picker.hpp
```diff
@@ -664,15 +664,36 @@ class start_game_button : public button_element_base {
 			}
 			text::close_layout_box(contents, box);
 		}
+
+		auto box = text::open_layout_box(contents, 0);
+		text::add_to_layout_box(state, contents, box, std::string_view("在标准状况下 medžuslovjansky jezyk ?"));
```
schombert, line 669: I am guessing that this text is here for debugging, and you have left it in accidentally?

schombert:
You do need to update the state name function in text.cpp. In fact, look for anywhere that the nation adjective is used. In most of these places we simply stick it in front of things followed by `" "`. All of those will have to change to a two-variable substitution string from the localization file so that translators can control both the order and whether there is a space (both vary from language to language).

src/map/map_state.cpp
```diff
@@ -378,25 +378,31 @@ void update_text_lines(sys::state& state, display_data& map_data) {
...
+									if(degrees < 0.f) {
+										degrees = 360.f + degrees;
+									}
+									assert(degrees >= 0.f && degrees <= 360.f);
+									auto nation_name = text::produce_simple_string(state, n.get_name());
+									if(nation_name.starts_with("The ")) {
+										nation_name.erase(0, 4);
+									}
+									if(degrees >= 0.f && degrees < 90.f) {
+										name = "East " + nation_name;
```
schombert, line 398: strange to put this, untranslatable, addition into the localization update. I would suggest having four directional csv entries and something like a `$adj$ $x$` text entry that you substitute them into, allowing each language to customize how adjectives work.

assets/localisation/alice.csv
Nivaturimika: We are going to eventually show a list of supported languages, hence the multiple definitions

Nivaturimika:
> You do need to update the state name function in text.cpp. In fact, look for anywhere that the nation adjective is used (map labels also come to mind; there are probably more). In most of these places we simply stick it in front of things followed by `" "`. All of those will have to change to a two-variable substitution string from the localization file so that translators can control both the order and whether there is a space (both vary from language to language).

Already did this

PR #1543, opened 2024-05-19
AI "easy" surrender now behind a define, also prevent addition of wgs…

Nivaturimika:
… when it would be redundant

### DIFF
--- a/src/ai/ai.cpp
+++ b/src/ai/ai.cpp
@@ -2596,10 +2596,8 @@ bool has_cores_occupied(sys::state& state, dcon::nation_id n) {
 				return false;
 			}
 		}
-		//no cores or some cores are occupied but some are not
-		auto pc = state.world.nation_get_province_control(n);
-		if(pc.begin() == pc.end())
-			return true;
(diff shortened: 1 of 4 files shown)


src/ai/ai.cpp
```diff
@@ -2589,16 +2589,18 @@ void add_gw_goals(sys::state& state) {
...
+	if(bool(state.defines.alice_surrender_on_cores_lost)) {
+		auto i = state.world.nation_get_identity_from_identity_holder(n);
+		for(auto c : state.world.national_identity_get_core(i)) {
+			if(c.get_province().get_nation_from_province_control() == n) {
+				return false;
+			}
 		}
+		//no cores or some cores are occupied but some are not
+		auto pc = state.world.nation_get_province_control(n);
+		if(pc.begin() == pc.end())
```
schombert, line 2601: the above is redundant. the way you have written this function, it returns true if the nation in question controls no provinces.

Nivaturimika: only for cores

PR #1548, opened 2024-05-19
Add wgs after check

Nivaturimika:


src/ai/ai.cpp
```diff
@@ -2467,6 +2467,7 @@ void add_free_ai_cbs_to_war(sys::state& state, dcon::nation_id n, dcon::war_id w
 		sort_avilable_cbs(potential, state, n, w);
 		for(auto& p : potential) {
 			if(!military::war_goal_would_be_duplicate(state, n, w, p.target, p.cb, p.state_def, p.associated_tag, p.secondary_nation)) {
+				assert(command::can_add_war_goal(state, n, w, p.target, p.cb, p.state_def, p.associated_tag, p.secondary_nation));
```
schombert, line 2470: these asserts won't work because adding the wargoal is meant to bypass some of the checks

src/ai/ai.cpp
```diff
@@ -2570,6 +2571,7 @@ void add_wg_to_great_war(sys::state& state, dcon::nation_id n, dcon::war_id w) {
 	result.clear();
 	place_instance_in_result_war(state, result, n, target, w, cb, target_states);
 	if(result[0].target) {
+		assert(command::can_add_war_goal(state, n, w, target, cb, result[0].state_def, result[0].associated_tag, result[0].secondary_nation));
```
schombert, line 2574: see above

PR #1556, opened 2024-05-21
2 GP limit (instead of 3), Renaming not using terrain names, Fix liberated nations not being on sphere

Nivaturimika:
.


src/military/military.cpp
```diff
@@ -2848,8 +2848,14 @@ void implement_war_goal(sys::state& state, dcon::war_id war, dcon::cb_type_id wa
 				nations::create_nation_based_on_template(state, holder, from);
 			}
 		}
-
-		add_truce(state, holder, target, 365);
+		// add to sphere
+		if(state.world.nation_get_is_great_power(from)) {
```
schombert, line 2852: you should do this only if target existed is false (or, if you don't want that, only if it is not in a sphere already)

PR #1579, opened 2024-05-26
Move post-processing to the finish state of parsing

Nivaturimika:


src/parsing/nations_parsing.cpp
```diff
@@ -38,6 +38,8 @@ void national_identity_file::any_value(std::string_view tag, association_type, s
 	// Recognize the cleanup utility tag
 	if(is_fixed_token_ci(tag.data(), tag.data() + tag.length(), "cln")) {
 		context.state.national_definitions.cleanup_tag = new_ident;
+	} else if(is_fixed_token_ci(tag.data(), tag.data() + tag.length(), "reb")) {
+		context.state.national_definitions.rebel_id = context.state.world.national_identity_get_nation_from_identity_holder(new_ident);
```
schombert, line 42: are the nations actually created at this point (I believe that they are created later ...)?

Nivaturimika: Yes I realized that, changed rebel_id to a national identity anyways

PR #1583, opened 2024-05-27
fix region map mode tooltip

yrrrmmbl:


schombert:
this will still technically give the wrong name; you want text::get_dynamic_state_name

PR #1584, opened 2024-05-31
Fix foreign investment

Nivaturimika:


src/gamestate/commands.cpp
```diff
@@ -599,8 +598,8 @@ void execute_begin_factory_building_construction(sys::state& state, dcon::nation
 			}
 		}
 
-		nations::adjust_foreign_investment(state, source, state.world.state_instance_get_nation_from_state_ownership(location),
-				amount);
+		state.world.nation_get_stockpiles(source, economy::money) -= amount;
```
schombert, line 601: this will result in double paying, since the investing nation already pays for the construction as it is being built instead of the host country

src/gamestate/commands.cpp
```diff
@@ -599,6 +599,7 @@ void execute_begin_factory_building_construction(sys::state& state, dcon::nation
 		}
 
 		state.world.nation_get_stockpiles(source, economy::money) -= amount;
+		state.world.nation_get_stockpiles(state.world.state_instance_get_nation_from_state_ownership(location), economy::money) += amount;
```
schombert, line 602: This is still double paying, because, again, the country invested in doesn't pay for the construction. Also, since construction is new demand, the costs of the construction are likely to increase over time. Paying the initial amount in this way up front will undershoot the costs of construction. (This would be a way to bankrupt foreign countries if they had to pay for construction, as you could start basically infinite constructions at cheap prices and then kill them with the increasing costs)

PR #1594, opened 2024-06-03
Various economy tweaks

ineveraskedforthis:
- Develop subsistence a bit further: helps to avoid tons of rgo workers rebels
- Simply artisans logic while improving stability and increase speed of their demotion to avoid large scale artisan rebellions
- Add new columns to the ledger: GDP(PPP) and GDP(PPP) per 1m people
- Decrease late game economy instability
- Add presimulation period on scenarion creation, which also address a compaint of capitalists starting the same factory at the start of the game
- A few numerical tweaks
- ecodump console command


src/economy/economy.cpp
```diff
@@ -16,11 +16,32 @@ enum class economy_reason {
 	pop, factory, rgo, artisan, construction, nation, stockpile, overseas_penalty
 };
 
+std::vector<std::vector<float>> demand_by_category = {
+	{ }, { }, { }, { }, { }, { }, { }, { }, { }
+};
+
 void register_demand(sys::state& state, dcon::nation_id n, dcon::commodity_id commodity_type, float amount, economy_reason reason) {
 	state.world.nation_get_real_demand(n, commodity_type) += amount;
+	(demand_by_category[(int)reason])[commodity_type.index()] += amount;
```
schombert, line 25: this sort of double indirection is an unnecessary inefficiency. Instead add an
```
	property{
		name{ demand_by_category }
		type{array{issue_id}{int32_t}}
	}
```
property to the commodity and resize it to 8 as part of the economy setup code, which will be marginally faster

src/economy/economy.cpp
```diff
@@ -519,64 +546,120 @@ float artisan_scale_limit(sys::state& state, dcon::nation_id n, dcon::commodity_
...
-					total_possible += 1.0f;
-				}
-			}
+float get_artisans_multiplier(sys::state& state, dcon::nation_id n) {
+	float multiplier = 0.000001f * state.world.nation_get_everyday_needs_costs(n, state.culture_definitions.artisans);
+	return 1.f / (multiplier + 1.f);
+}
+
+void scores_to_distribution(std::vector<float> & scores, std::vector<float> & distribution, float multiplier) {
+	float max_score = std::numeric_limits<float>::min();
```
schombert, line 590: I'll quote from the reference here
> For floating-point types with denormalization, min() returns the minimum positive normalized value. Note that this behavior may be unexpected, especially when compared to the behavior of min() for integral types. To find the value that has no values less than it, use [lowest()](https://en.cppreference.com/w/cpp/types/numeric_limits/lowest).
You probably meant 0? or negative infinity?

src/economy/economy.cpp
```diff
@@ -519,64 +546,120 @@ float artisan_scale_limit(sys::state& state, dcon::nation_id n, dcon::commodity_
...
-			} else {
-				state.world.nation_set_artisan_distribution(n, cid, 0.0f);
-			}
+	for(uint32_t i = 1; i < scores.size(); ++i) {
+		distribution[i] = distribution[i] / (total + 0.001f);
+	}
+}
+
+void fill_artisan_distribution(sys::state& state, dcon::nation_id n, std::vector<float>& distribution) {
+	std::vector<float> scores;
```
schombert, line 612: so, there is a minor inefficiency here. Going to the memory manager to dynamically allocate a buffer is better avoided if you don't need it. In this case, the only reason you seem to be storing them at all is to calculate the maximum. Since the calculations are simple enough, it is almost certainly faster to loop over all the commodities and find the maximum and then calculate each score again as you need it, thereby saving the need to dynamically allocate an array at all.

src/economy/economy.cpp
```diff
@@ -519,64 +546,120 @@ float artisan_scale_limit(sys::state& state, dcon::nation_id n, dcon::commodity_
...
+
+	std::vector<float> current_distribution;
+	std::vector<float> prev_profits;
+	std::vector<float> profits;
+	profits.resize(csize + 1);
+	current_distribution.resize(csize + 1);
+
+	float mult = get_artisans_multiplier(state, n);
+
+	profits[csize] = 10.f / mult; // * state.world.nation_get_everyday_needs_costs(n, state.culture_definitions.artisans) / state.defines.alice_needs_scaling_factor;
```
schombert, line 644: I don't understand the intent of this. Here you initialize just the last value of the array. Did you mean to initialize all of them?

src/economy/economy.cpp
```diff
@@ -519,64 +546,120 @@ float artisan_scale_limit(sys::state& state, dcon::nation_id n, dcon::commodity_
...
+
+
+
+void adjust_artisan_balance(sys::state& state, dcon::nation_id n) {
+	auto const csize = state.world.commodity_size();
+	float distribution_drift_speed = 0.0001f;
+
+	std::vector<float> current_distribution;
+	std::vector<float> prev_profits;
+	std::vector<float> profits;
```
schombert, line 638: my comment about unnecessary buffers applies to these last two here as well (is prev_profits used at all?). In this case, since you aren't calculating anything that depends on the set of values as a whole, you can definitely get rid of these vectors. In fact, you could probably get rid of the current distribution vector as well if you turned the previous functions "inside out" (assuming you calculate the max value once, you can then write a function that takes that max + the given nation and commodity and returns the current distribution for it without having to store anything in memory).

src/economy/economy.cpp
```diff
@@ -1361,28 +1565,48 @@ void update_single_factory_production(sys::state& state, dcon::factory_id f, dco
...
 	auto slaves = state.world.province_get_demographics(p, demographics::to_employment_key(state, state.culture_definitions.slaves));
 
-	return std::tuple<float, float, float>(relevant_paid_population, slaves, relevant_paid_population + slaves);
+	rgo_workers_breakdown result = {
+		.paid_workers = relevant_paid_population,
+		.slaves = slaves,
+		.total = relevant_paid_population + slaves
+	};
+
+	return result;
```
schombert, line 1581: it makes no difference, but you can also write `return rgo_workers_breakdown{ ... }` if you want to save a few keystrokes

src/economy/economy.cpp
```diff
@@ -1361,28 +1565,48 @@ void update_single_factory_production(sys::state& state, dcon::factory_id f, dco
...
 float rgo_overhire_modifier(sys::state& state, dcon::province_id p, dcon::nation_id n) {
-	auto pops_max = rgo_max_employment(state, n, p);
+	float owned_land = (1.f - local_free_land(state, p));
+	auto pops_max = local_available_land(state, p) * owned_land * rgo_max_employment(state, n, p);
 	auto current_employment = pops_max * state.world.province_get_rgo_employment(p);
+
+	float result = math::sqrt(pops_max / (current_employment + math::sqrt(pops_max) + 10.f));
+
+	assert(std::isfinite(result));
+	assert(result >= 0.f);
```
schombert, line 1592: really hope that our sqrt function can't return negative values

src/economy/economy.cpp
```diff
@@ -2059,10 +2327,18 @@ void update_pop_consumption(sys::state& state, dcon::nation_id n, float base_dem
...
+		subsistence -= subsistence_life;
+		float subsistence_everyday = std::clamp(subsistence, 0.f, subsistence_score_everyday);
+		subsistence -= subsistence_everyday;
+		float subsistence_luxury = std::clamp(subsistence, 0.f, subsistence_score_luxury);
+
 		for(auto pl : state.world.province_get_pop_location(p.get_province())) {
 			auto t = pl.get_pop().get_poptype();
 			assert(t);
-			auto total_budget = pl.get_pop().get_savings();
+			auto total_budget = pl.get_pop().get_savings() * 0.8f;
```
schombert, line 2341: is there a reason why you are having pops spend only 80% here? Unless there are other changes, pop income simply overwrites the old values, so the remainder will just be lost. Secondly, even if you have changed that, it still converges to pops spending exactly their income each day, the only difference being a slight lag until it reaches that point.

src/economy/economy.cpp
```diff
@@ -3415,11 +3877,25 @@ void daily_update(sys::state& state) {
 	// make constructions:
 	resolve_constructions(state);
 
+	if(!initiate_buildings) {
+		goto AFTER_INVESTMENTS;
```
schombert, line 3881: just write the if block to surround the entire section, as using a goto will set off many linters

src/gamestate/dcon_generated.txt
```diff
@@ -3555,6 +3560,11 @@ object {
 		type{ array{commodity_id}{float} }
 		tag{ save }
 	}
+	property {
+		name{ intermediate_demand }
+		type{ array{commodity_id}{float} }
+		tag{ save }
```
schombert, line 3566: I don't know your formulas, so I am asking just to double check, but do these intermediate values need to persist between days? If they are reset at the start of each day, then they can be safely omitted from the save file

src/economy/economy.hpp
```diff
@@ -155,15 +167,22 @@ float pop_farmer_min_wage(sys::state& state, dcon::nation_id n, float min_wage_f
...
+
 float rgo_overhire_modifier(sys::state& state, dcon::province_id p, dcon::nation_id n);
 float rgo_desired_worker_norm_profit(sys::state& state, dcon::province_id p, dcon::nation_id n, float min_wage, float total_relevant_population);
 float rgo_expected_worker_norm_profit(sys::state& state, dcon::province_id p, dcon::nation_id n);
 
 
 void update_rgo_employment(sys::state& state);
 void update_factory_employment(sys::state& state);
-void daily_update(sys::state& state);
+void daily_update(sys::state& state, bool initiate_building);
```
schombert, line 185: you could have done this instead: `void daily_update(sys::state& state, bool initiate_building = true);` this would give the new parameter a default value, meaning that in old code, where it was not present, it would be treated as being passed `true`. Of course, this isn't helpful now, since you have already added the new parameter eveywhere

src/economy/economy.cpp
ineveraskedforthis: I meant negative infinity, thanks. But as one of values was guaranteed to be positive, it doesn't influence anything. Will change to lowest.

src/economy/economy.cpp
ineveraskedforthis: No, this is to set the baseline for profit. If all profit (after adjustment) is small compared to this number, softmax will ensure that artisans will not produce goods if they are that unprofitable.

src/economy/economy.cpp
ineveraskedforthis: To smooth out spendings a bit

schombert: Ok, but there are places in the code where I have probably assumed that pop savings == pop income after the end of the day (which was the case before). You may want to search the files for `et_savings` and give every old use of this value an inspection

PR #1600, opened 2024-06-07
Implement `rgo_distribution` for province history.

ineveraskedforthis:
Syntax is 
```
rgo_distribution = {
	entry = {
		trade_good = silk
		max_employment = 100000
	}
	entry = {
		trade_good = opium
		max_employment = 100000
	}
}
```
It overwrites values which would be otherwise generated by a game.

### DIFF
--- a/src/gamestate/dcon_generated.txt
+++ b/src/gamestate/dcon_generated.txt
@@ -1511,6 +1511,11 @@ object {
 		type{ float }
 	}
 	property{
+		name{ rgo_was_set_during_scenario_creation }
+		type{ bitfield }
+		tag{ scenario, save }
+	}
+	property{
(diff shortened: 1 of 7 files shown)


src/gamestate/dcon_generated.txt
```diff
@@ -1510,6 +1510,11 @@ object {
 		name{ subsistence_score }
 		type{ float }
 	}
+	property{
+		name{ rgo_was_set_during_scenario_creation }
+		type{ bitfield }
+		tag{ scenario, save }
```
schombert, line 1516: If this needs to be persistent, it should be in one or the other. But, since the initialize economy function is only run once, during scenario creation, it may not need to be persistent at all.

PR #1612, opened 2024-06-11
Fix console suggestions

Nivaturimika:


assets/localisation/de-DE/map.csv
```diff
@@ -1,12 +1,12 @@
-map_remove_prefix;Der
```
schombert, line 1: won't there be two or three possible predixes because of the various genders?

PR #1615, opened 2024-06-11
Map and text fixes

Nivaturimika:


src/text/text.cpp
```diff
@@ -109,11 +109,11 @@ text_sequence create_text_sequence(sys::state& state, std::string_view content,
 						temp.push_back(char(unicode));
 					} else if(unicode <= 0x7FF) {
 						temp.push_back(char(0xC0 | uint8_t(0x1F & (unicode >> 6))));
-						temp.push_back(char(0x80 | uint8_t(0x3F & unicode)));
```
schombert, line 112: removing this will break the utf8 values

PR #1619, opened 2024-06-12
Restore classic fonts

Nivaturimika:


src/graphics/opengl_wrapper.cpp
```diff
@@ -789,9 +790,120 @@ void render_new_text(sys::state& state, text::stored_glyphs const& txt, color_mo
...
+	// The color variables are unused currently.
+	//
+	// Spacing, kearning, etc. are already applied.
+	// Scaling (unintentionally) is also applied (by whatever part of Alice scales the normal fonts).
+
+	glActiveTexture(GL_TEXTURE0);
+	glBindTexture(GL_TEXTURE_2D, font.ftexid);
+
+	for(uint32_t i = 0; i < txt.glyph_count; ++i) {
+		if(uint8_t(txt.glyph_info[i].codepoint) == '@') {
```
schombert, line 818: codepoint isn't the utf32 value, it is the internal font glyph id, which has not intrinsic relation to the codepoint. Also, you would still need to convert it back to win1252 at some point

Nivaturimika: fixed

PR #1653, opened 2024-06-20
Add page for a gdp history.

ineveraskedforthis:


src/gamestate/system_state.cpp
```diff
@@ -4366,6 +4367,13 @@ void state::single_game_tick() {
 		}
 	}
 
+	if((ymd_date.month % 3) == 0) {
```
schombert, line 4370: on every day of the month?

PR #1655, opened 2024-06-21
fix macro new

stevezxc:


src/gui/gui_minimap.hpp
```diff
@@ -225,6 +225,9 @@ class macro_builder_unit_listbox : public listbox_element_base<macro_builder_uni
 };
 class macro_builder_new_template_button : public button_element_base {
 public:
+	void on_update(sys::state& state) noexcept override {
+		set_button_text(state, text::produce_simple_string(state, "macro_new_template"));
```
schombert, line 229: instead of changing the text in on_update, which will have the effect of changing the text every time the day changes, simply change assets/alice.gui line 4019 to have a default text of macro_new_template instead of new, which will cause that text to be localized properly when it is created. (you will have to recreate the scenario, however, to see this change take effect)

PR #1664, opened 2024-06-23
Classic font

Nivaturimika:


src/text/text.cpp
```diff
@@ -1457,9 +1457,7 @@ void add_to_layout_box(sys::state& state, layout_base& dest, layout_box& box, st
 					++m;
 				}
 				
-				auto cluster_end = m < int32_t(all_glyphs.glyph_info.size())
```
schombert: don't understand why this safety check has been removed

PR #1670, opened 2024-06-27
less willing if more milscore

Nivaturimika:


src/ai/ai.cpp
```diff
@@ -2111,7 +2111,7 @@ dcon::cb_type_id pick_fabrication_type(sys::state& state, dcon::nation_id from,
 
 bool valid_construction_target(sys::state& state, dcon::nation_id from, dcon::nation_id target) {
 	// Copied from commands.cpp:can_fabricate_cb()
-	if(from == target)
+	// if(from == target)
```
schombert, line 2114: uh, this function will now always return false ...

src/ai/ai.cpp
```diff
@@ -3040,48 +3031,11 @@ void make_war_decs(sys::state& state) {
...
-
-			if(real_target == n)
-				continue;
-			if(nations::are_allied(state, n, real_target) || nations::are_allied(state, n, other))
-				continue;
-			if(real_target.get_in_sphere_of() == n)
-				continue;
-			if(state.world.nation_get_in_sphere_of(other) == n)
-				continue;
-			if(military::has_truce_with(state, n, other) || military::has_truce_with(state, n, real_target))
```
schombert, line 3057: why did all these checks get removed?

src/ai/ai.cpp
Nivaturimika: FUCK HOW AM I SO IDIOTIC

src/ai/ai.cpp
Nivaturimika: fixed

PR #1688, opened 2024-07-06
nerf rebels

Nivaturimika:
less organisation, same strength, so more fair and balanced (cause rebels arent professional armies haha)


schombert:
does this actually do anything? if they aren't in combat, their org should recover somewhat quickly

Nivaturimika:
> does this actually do anything? if they aren't in combat, their org should recover somewhat quickly

no they dont recover org because theyre always sieging

Nivaturimika:
USUALLY starting sieging

PR #1702, opened 2024-07-09
define to allow running military tick all ticks

Nivaturimika:


src/gui/gui_element_types.cpp
```diff
@@ -757,30 +757,11 @@ void simple_text_element_base::format_text(sys::state& state) {
...
-		internal_layout.contents.clear();
-		internal_layout.number_of_lines = 0;
-
-		auto al = text::to_text_alignment(base_data.data.button.get_alignment());
-		text::single_line_layout sl{ internal_layout, text::layout_parameters{ 0, 0, static_cast<int16_t>(base_data.size.x), static_cast<int16_t>(base_data.size.y),
-					base_data.data.button.font_handle, 0, al, black_text ? text::text_color::black : text::text_color::white, true, true },
-			state.world.locale_get_native_rtl(state.font_collection.get_current_locale()) ? text::layout_base::rtl_status::rtl : text::layout_base::rtl_status::ltr };
-
-		sl.add_text(state, cached_text);
+		set_text(state, text::produce_simple_string(state, base_data.data.button.txt));
```
schombert, line 760: there is a reason it wasn't like this before. set text uses the existing cached text to bypass an expensive update if the text hasn't changed. However, when changing fonts, for example, even though the text hasn't changed, we still want the button to remake the text because the layout will have changed

Nivaturimika: fiuxdferrddf

PR #1707, opened 2024-07-12
Port to OpenGL 3.3

Nivaturimika:


schombert:
you should probably do all your glGetUniformLocation(open_gl.ui_shader_program, "texture_sampler") calls when you create the program and then store them, so that you don't have to do it again. Since these calls do string comparisons, they aren't entirely cheap, and in some situations we will be rebinding uniforms extremely frequently

PR #1716, opened 2024-07-17
Fix justification when truce

Nivaturimika:


schombert:
you also need to update the explanation for the fabricate button, right?

PR #1729, opened 2024-07-25
fix state selector crash

Nivaturimika:


schombert:
closed with gratitude; I just went with what peter preferred

PR #1745, opened 2024-08-14
Military automation

ineveraskedforthis:
Add basic functionality for military automation with ability to select zones and designate orders for them.


src/gamestate/dcon_generated.txt
```diff
@@ -1678,6 +1678,130 @@ object {
...
+	property {
+		name { ferry_origin }
+		type { province_id }
+	}
+}
+
+
+relationship{
+	name{ automation }
+	tag{ save }
```
schombert, line 1791: should be local_save

src/gamestate/dcon_generated.txt
```diff
@@ -1678,6 +1678,130 @@ object {
...
+		type { sys::army_group_regiment_task }
+		tag { local_save }
+	}
+	property {
+		name { await_command_execution_flag }
+		type { bitfield }
+	}
+	property {
+		name { target }
+		type { province_id }
```
schombert, line 1774: do any of these need to be saved?

src/gamestate/game_scene.cpp
```diff
@@ -464,8 +464,11 @@ void on_lbutton_up(sys::state& state, int32_t x, int32_t y, sys::key_modifiers m
 	if(state.user_settings.left_mouse_click_hold_and_release && state.ui_state.left_mouse_hold_target) {
 		on_lbutton_up_ui_click_hold_and_release(state, x, y, mod);
 	}
-	on_lbutton_up_map(state, x, y, mod);
-	state.current_scene.lbutton_up(state);
+
+	if(!state.ui_state.under_mouse) {
+		on_lbutton_up_map(state, x, y, mod);
```
schombert, line 469: I believe the map lbutton up belongs outside this check because, if you were dragging a selection box for units, you probably wouldn't want the selection process to continue just because you happened to end over a ui element

src/gamestate/game_scene.cpp
```diff
@@ -919,11 +927,9 @@ void clean_up_basic_game_scene(sys::state& state) {
 }
 
 void update_army_group_selection_ui(sys::state& state) {
-	if(state.selected_army_group != nullptr) {
-		state.ui_state.army_group_window_sea->set_visible(state, true);
+	if(state.selected_army_group) {
```
schombert, line 930: small elegance improvement: you can just replace `true` or `false` with this and turn the function into a one liner

src/gamestate/system_state.cpp
```diff
@@ -541,403 +537,12 @@ void state::render() { // called to render the frame may (and should) delay retu
...
-							data[0] = current_regiment;
-							command::mark_regiments_to_split(*this, local_player_nation, data);
-							command::split_army(*this, local_player_nation, army);
-							break;
-						}
-					}
-				}
-			}
-		}
+		world.for_each_automated_army_group([&](dcon::automated_army_group_id item) {
```
schombert, line 540: do we want this to be done every *frame*? Is there a reason that it couldn't be put in the on_update function, which is called every time the game state is updated (so at least once per in-game day)? That should be enough to let it respond to changes

src/gamestate/system_state.cpp
```diff
@@ -4539,385 +4144,962 @@ void state::console_log(std::string_view message) {
...
-	if(index != group->defensive_line.end()) {
-		group->defensive_line.erase(index);
-		return;
+void state::toggle_defensive_position(dcon::automated_army_group_id group, dcon::province_id position) {
+	auto fat_group = fatten(world, group);
+
+	if(fat_group.get_provinces_defend().contains(position)) {
+		fat_group.get_provinces_defend().remove_unique(position);
+	} else {
+		fat_group.get_provinces_defend().add_unique(position);
```
schombert, line 4172: micro optimization: if you know that it isn't in the list, then just pushing it to the back is cheaper

PR #1760, opened 2024-08-22
Ledger SoL, two script effects

SneakBug8:
Imgur

### DIFF
--- a/src/scripting/effects.cpp
+++ b/src/scripting/effects.cpp
@@ -1715,6 +1715,15 @@ uint32_t ef_religion(EFFECT_PARAMTERS) {
 	ws.world.nation_set_religion(trigger::to_nation(primary_slot), trigger::payload(tval[1]).rel_id);
 	return 0;
 }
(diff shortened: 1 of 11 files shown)


src/scripting/effects.cpp
```diff
@@ -2893,6 +2902,12 @@ uint32_t ef_reduce_pop(EFFECT_PARAMTERS) {
 	ws.world.pop_get_size(trigger::to_pop(primary_slot)) *= amount;
 	return 0;
 }
+uint32_t ef_reduce_pop_abs(EFFECT_PARAMTERS) {
+	auto amount = trigger::read_int32_t_from_payload(tval + 1);
+	assert(std::isfinite(amount));
+	ws.world.pop_get_size(trigger::to_pop(primary_slot)) -= amount;
```
schombert, line 2908: this could result in pops with negative size, which could break the game

schombert: also, asserting that an integer is finite doesn't make too much sense, and is responsible for the CI compilation error

PR #1764, opened 2024-08-23
Effects changes, subjects can't become GPs

SneakBug8:
- Allow using `religion` effect in pop context to change pop's religion.
- Introduce `set_culture` effect to change culture of pop to any culture.


schombert:
Your new functions that actually implement the effect / display its tooltip don't appear to be added to the array of functions in either case.

PR #1767, opened 2024-08-24
Fix infinites in the economy

SneakBug8:
Replace some NaNs happening in the economy presim for goods that are not available at game start.


src/economy/economy.cpp
```diff
@@ -887,6 +887,8 @@ float rgo_effective_size(sys::state const& state, dcon::nation_id n, dcon::provi
 	// technology-general-farm-or-mine-size-bonus + provincial-mine-or-farm-size-modifier + 1)
 	auto rgo_ownership = state.world.province_get_landowners_share(p) + state.world.province_get_capitalists_share(p);
 	auto sz = state.world.province_get_rgo_max_size_per_good(p, c) * rgo_ownership + base;
+	if(!std::isfinite(sz))
```
schombert, line 890: I don't like including checks such as this in release. The nans / infinite values should be fixed earlier, before they ever get to this point. Ideally in scenario creation, if possible, or if not when the file is loaded

PR #1777, opened 2024-09-02
Military, AI, and UI changes

SneakBug8:


src/ai/ai.cpp
```diff
@@ -294,6 +297,12 @@ bool ai_will_accept_alliance(sys::state& state, dcon::nation_id target, dcon::na
 				}
 				++gp_count;
 			}
+			if(rel.get_are_allied()) {
+				if(alli_count >= 4) {
+					return false;
```
schombert, line 302: This change also needs to be reflected in the offer alliance tooltip so that the player will know why they can't make an alliance

src/military/military.cpp
```diff
@@ -4659,7 +4746,7 @@ void end_battle(sys::state& state, dcon::naval_battle_id b, battle_result result
 				auto em_regs = em.get_army().get_army_membership();
 				while(em_regs.begin() != em_regs.end() && transport_cap < 0) {
 					auto reg_id = (*em_regs.begin()).get_regiment();
-					disband_regiment_w_pop_death(state, reg_id);
+					state.world.delete_regiment(reg_id);
```
schombert, line 4749: Is there a reason that you have reverted "killing" regiments that are lost at sea?

SneakBug8: Reintroduced

src/ai/ai.cpp
SneakBug8: Changed the approach to threat calculation instead

src/ai/ai.cpp
```diff
@@ -75,6 +75,14 @@ void update_ai_general_status(sys::state& state) {
 			self_str += 0.75f * float(subj.get_subject().get_military_score());
 		float defensive_str = estimate_defensive_strength(state, n);
 
+		// Include strength of allies in threat calculation. 0.75f to factor for allies not joining war.
+		for(const auto rel : state.world.nation_get_diplomatic_relation(n)) {
+			auto n2 = rel.get_related_nations(rel.get_related_nations(0) == n ? 1 : 0);
+			if(rel.get_are_allied()) {
+				defensive_str += estimate_defensive_strength(state, n2) * 0.75f;
```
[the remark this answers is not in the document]
SneakBug8: Undone

PR #1789, opened 2024-11-17
Local economy and trade routes.

ineveraskedforthis:
Implements trade routes and local economies.


src/economy/demographics.cpp
```diff
@@ -97,25 +97,31 @@ void set_employment(sys::state& state, dcon::pop_id p, float v) {
 void set_raw_employment(sys::state& state, dcon::pop_id p, float v) {
 	state.world.pop_set_uemployment(p, to_pu8(v));
 }
-float get_life_needs(sys::state const& state, dcon::pop_id p) {
+template<typename P, typename V>
+V get_life_needs(sys::state const& state, P p) {
```
schombert, line 101: Writing it this way would require you to provide the type of V wherever this function is called since it cannot be deduced from the call site (get_life_needs<float>(...) for example). Instead of V, you can write `auto` for the return type which will cause the compiler to deduce the return type from the function body

src/economy/economy.cpp
```diff
@@ -13,33 +13,591 @@
...
+	auto total_cost = 0.f;
+	auto total_supply = 0.f;
+	state.world.for_each_market([&](auto m) {
+		auto local_price = price(state, m, c);
+		auto local_supply = state.world.market_get_supply(m, c) + 0.0001f;
+
+		total_cost += local_price * local_supply;
+		total_supply += local_supply;
+	});
+
```
schombert, line 82: you may want a non-zero check here as in the function above

src/economy/economy.cpp
```diff
@@ -13,33 +13,591 @@
...
+) {
+	auto total = 0.f;
+	auto count = 0.f;
+	state.world.nation_for_each_state_ownership(s, [&](auto soid) {
+		auto sid = state.world.state_ownership_get_state(soid);
+		auto market = state.world.state_instance_get_market_from_local_market(sid);
+
+		auto local_count = state.world.state_instance_get_demographics(
+			sid,
+			demographics::to_key(state, state.culture_definitions.capitalists)
```
schombert, line 284: Since the key is a constant value here, you could move the function that computes it out of the loop.

src/economy/economy.cpp
```diff
@@ -13,33 +13,591 @@
...
+		);
+
+		total += state.world.market_get_luxury_needs_costs(
+			market, state.culture_definitions.capitalists
+		) * local_count / state.defines.alice_needs_scaling_factor;
+		count += local_count;
+	});
 
-	state.world.nation_get_gdp(n) -= amount * price * sat;
+	return total / count;
```
schombert, line 293: Sure that you don't need a guard against dividing by zero here?

src/economy/economy.cpp
```diff
@@ -13,33 +13,591 @@
...
-void register_domestic_supply(sys::state& state, dcon::nation_id n, dcon::commodity_id commodity_type, float amount, economy_reason reason) {
-	state.world.nation_get_domestic_market_pool(n, commodity_type) += amount;
-	state.world.nation_get_gdp(n) += amount * state.world.commodity_get_current_price(commodity_type);
+
+trade_volume_data_detailed export_volume_detailed(
+	sys::state& state,
+	dcon::nation_id s,
+	dcon::commodity_id c
+) {
+	static auto per_nation_data = state.world.nation_make_vectorizable_float_buffer();
```
schombert, line 323: So, this may not work the way you want. By making it static, the buffer is only created once, which means that it will never grow if more nations are added. Now, currently, that probably isn't an issue since we create one nation per tag and just leave it like that, so in practice the the number nations are fixed. However, considering that this is a UI function (I assume) I would just remove the `static` here, paying the cost of allocating it once per call and making it more flexible for the future. On the other hand, this function is wedded to there being a fixed number of nations anyways as my next comment will explain ...

src/economy/economy.cpp
```diff
@@ -13,33 +13,591 @@
...
+			auto s_target = state.world.market_get_zone_from_local_market(target);
+			auto n_origin = state.world.state_instance_get_nation_from_state_ownership(s_origin);
+			auto n_target = state.world.state_instance_get_nation_from_state_ownership(s_target);
+
+			auto distance = state.world.trade_route_get_distance(trade_route);
+
+			auto trade_good_loss_mult = std::max(0.f, 1.f - 0.0001f * distance);
+
+			if(n_origin != n_target) {
+				per_nation_data.get(n_target) += sat * absolute_volume;
```
schombert, line 359: So, the issue here is that if this function is running in parallel with the game and the number of nations can grow, it is possible for a new nation to be created and found in the target list that has a greater id than what the buffer created above was sized to accommodate. If the number of nations can change, this function would have to check that the target id is within the size allocated in the buffer and simply discard it if it is not.

src/economy/economy.cpp
```diff
@@ -13,33 +13,591 @@
...
+	});
+	return total;
+}
+
+trade_volume_data_detailed import_volume_detailed(
+	sys::state& state,
+	dcon::nation_id s,
+	dcon::commodity_id c
+) {
+	static auto per_nation_data = state.world.nation_make_vectorizable_float_buffer();
```
schombert, line 423: By two comments above apply to this function as well.

src/economy/economy.cpp
```diff
@@ -95,24 +653,35 @@ int32_t previous_gdp_record_index(sys::state& state) {
...
+}
+
 float full_spending_cost(sys::state& state, dcon::nation_id n);
 void populate_army_consumption(sys::state& state);
 void populate_navy_consumption(sys::state& state);
 void populate_construction_consumption(sys::state& state);
 
 float commodity_daily_production_amount(sys::state& state, dcon::commodity_id c) {
-	return state.world.commodity_get_total_production(c);
+	// TODO
```
schombert, line 683: Does this need to be filled for any critical bits of the ui?

ineveraskedforthis: It's not used at all.

src/economy/economy.cpp
```diff
@@ -143,80 +712,145 @@ bool has_factory(sys::state const& state, dcon::state_instance_id si) {
...
 				count++;
 			}
 		});
 
+		// remove some of needs completely depending on price
+		float total_weights_adjusted = 0.f;
+		state.world.for_each_commodity([&](dcon::commodity_id c) {
+			if(valid_life_need(state, nation, c)) {
+				auto weight = need_weight(state, n, c);
+				if(weight > total_weights * 0.05f) {
```
schombert, line 839: What happens if there are enough life need items that *all* of them being less than 5% is a possible occurrence?

src/economy/economy.cpp
```diff
@@ -405,63 +1068,171 @@ bool nation_is_constructing_factories(sys::state& state, dcon::nation_id n) {
...
+) {
+	auto alice_output_base = state.defines.alice_output_base_factor_artisans;
+	auto nation_output_mod = 1.f + state.world.nation_get_modifier_values(
+		nations, sys::national_mod_offsets::artisan_input);
+
+	return ve::max(alice_output_base * nation_output_mod, 0.01f);
+}
+
+template<typename T>
+ve::fp_vector artisan_throughput_multiplier(
```
schombert, line 1110: should be able to eliminate the duplicate one-item versions by making the return type `auto`

src/economy/economy.cpp
```diff
@@ -3275,91 +5274,181 @@ void daily_update(sys::state& state, bool initiate_buildings) {
...
-		float decay = 0.5f;
-		float world_pool = state.world.commodity_get_global_market_pool(c) * decay;
-		ve::fp_vector sum;
-		state.world.execute_serial_over_nation([&](auto nids) {
-			sum = sum + state.world.nation_get_domestic_market_pool(nids, c);
-			state.world.nation_set_domestic_market_pool(nids, c, 0.0f);
-		});
-		state.world.commodity_set_global_market_pool(c, world_pool + sum.reduce());
+
+	state.world.execute_serial_over_market([&](auto ids) {
```
schombert, line 5364: I think that this could be done in parallel, or that it could be done in parallel with a few adjustments. If I am reading the code correctly, the only things that block parallelization are the effects on nation stockpiles/treasury that come from buying and selling from the stockpiles. And since that is only done in capital markets, and each nation has only one of those, it shouldn't be possible for this function to try to touch values for the same nation if it was run in parallel.

If my analysis is wrong here, it looks like it should still be relatively simple to parallelize. Simply create one or more buffers that are the size of the number of markets to record effects on the owning nation and instead of doing the addition/subtraction effect in function, write the change to the buffer. Then the changes can be applied in serial after the main work, done in parallel, has been finished.

src/economy/economy.cpp
```diff
@@ -3424,193 +5521,349 @@ void daily_update(sys::state& state, bool initiate_buildings) {
...
 		state.world.nation_get_stockpiles(n, money) += refund;
+	});
 
-		auto const min_wage_factor = pop_min_wage_factor(state, n);
+	sanity_check(state);
+
+	/* now we know demand satisfaction and can set actual satifaction of pops */
+
+	/* prepare needs satisfaction caps */
+	state.world.execute_serial_over_market([&](auto ids) {
```
schombert, line 5744: also looks parallelizable

src/economy/economy.cpp
```diff
@@ -3424,193 +5521,349 @@ void daily_update(sys::state& state, bool initiate_buildings) {
...
+			state.world.market_set_supply(markets, c, ve::fp_vector{});
+			state.world.market_set_import(markets, c, ve::fp_vector{});
+			state.world.market_set_export(markets, c, ve::fp_vector{});
+		});
+	});
+
+	// we can handle each trade good separately: they do not influence each other
+	// 
+	// register trade supply
+	concurrency::parallel_for(uint32_t(0), total_commodities, [&](uint32_t k) {
```
schombert, line 5817: Assuming you don't allow the money commodity to be traded, you can start this parallel loop at `uint32_t(1)`, since it is always id 0 (mods that add additional such commodities mean that the check for the money commodity is still necessary of course)

src/economy/economy.cpp
```diff
@@ -3881,144 +6189,138 @@ void daily_update(sys::state& state, bool initiate_buildings) {
...
+			auto current_price = ve_price(state, ids, cid);
+			auto market_balance = demand - supply;
+			auto oversupply_factor = ve::max(supply / demand - 1.f, 0.f);
+			auto overdemand_factor = ve::max(demand / supply - 1.f, 0.f);
+			auto speed_modifer = (overdemand_factor - oversupply_factor);
+			auto price_speed = 0.0001f * speed_modifer;
+			price_speed = price_speed * current_price;
+			current_price = current_price + price_speed;
+
+			//the only purpose of upper price bound is to prevent float overflow
```
schombert, line 6325: might be useful to have an is finite assert here

src/economy/economy.cpp
```diff
@@ -4299,35 +6650,60 @@ void daily_update(sys::state& state, bool initiate_buildings) {
...
+							state.world.force_create_province_building_construction(best_p.first, n)
+						);
 						new_rr.set_is_pop_project(true);
 						new_rr.set_type(uint8_t(province_building_type::railroad));
 						//found_investment = true;
 					}
 				}
 			}
-			n.set_private_investment(0.0f);
+			//n.set_private_investment(0.0f);
```
schombert, line 6658: Do you clear the private investment amount in some other place? Because without that I believe that investment money will just increase constantly day over day.

ineveraskedforthis:
```cpp
			float pi_total = full_private_investment_cost(state, n);
			float perceived_spending = pi_total * 10.f;
			float pi_budget = state.world.nation_get_private_investment(n);
			auto pi_scale = perceived_spending <= pi_budget ? 1.0f : pi_budget / perceived_spending;
			//cut away low values:
			pi_scale = std::max(0.f, pi_scale - 0.1f);
			state.world.nation_set_private_investment_effective_fraction(n, pi_scale);
			state.world.nation_set_private_investment(n, std::max(0.0f, pi_budget - pi_total * pi_scale));
```

ineveraskedforthis: Instead of clearing it, i reduce it by with actual spending.

schombert: The reason we don't do that previously is because it can result in undesirable boom and bust behavior. If there is ever a situation where the investors don't have anything to build (say, because of laws) the investment pool will build up and up. Then, when they can invest, the huge pool will start many, many projects at once. Which they won't be able to actually afford day after day, and so most will sit idle clogging up spaces.

ineveraskedforthis: Main reason i removed it is because i do not want investors to spend all their money on the projects as it leads to them being ready to pay any price for construction goods. If they are always eager to spend entire budget, it encourages prices to grow  up to the limit. To avoid it, I cut away lower values of private budget scale, so if price is too high and they can afford only 0.1 of goods, they spend nothing. 

Then there are two cases: this price is caused by economic environment and they save up until they can afford it. And if this high price is caused by them, they could save up until they are able to pay for 0.1 of goods and push the price further but the whole process is more smooth and allows producers and trade routes to adapt to the rising prices.

And being able to save up money is crucial for this, otherwise initially poor investors will not be able to construct anything if they decide that the price is too high for them, decide to not invest and then the game clears up their savings.

schombert: I understand the reasoning, but then something needs to be added to prevent them from saving up simply because there is nothing to invest in. You can probably just add a boolean somewhere that detects in the possible investment section whether a no investment result is because of prices being too high, since you are testing for that, and then clear the pool if that flag is not set, meaning that excess money is accumulating for other reasons.

ineveraskedforthis: I can add a process which siphons part of investment pool back to pops (not only investors, but a wide spread of pops, could represent payment for some abstract services or gifts/bribery/charity). This way it will not grow above a certain amount and will add additional money flows which the game currently lacks.

schombert: Sounds fine to me. Just as long as we don't end up with a country full of stuck rail roads being built

src/provinces/province.cpp
```diff
@@ -716,6 +717,51 @@ void change_province_owner(sys::state& state, dcon::province_id id, dcon::nation
 			if(state.world.province_get_building_level(id, uint8_t(economy::province_building_type::naval_base)) > 0)
 				state.world.state_instance_set_naval_base_is_taken(new_si, true);
 
+			auto new_market = state.world.create_market();
+
+			// set prices to something to avoid infinite demand:
+
+			state.world.for_each_commodity([&](auto cid){
+				state.world.market_set_price(new_market, cid, 10.f);
+			});
```
schombert, line 726: Would it make sense to copy the market data from the previous state's market? A common situation will be a whole state (and hence its market) changing owners, but this is implemented as a sequence of province ownership changes. Of course, a single province could also split up the state, so I guess only copy the data if it wouldn't break things if the old market kept existing.

src/nations/nations.cpp
```diff
@@ -140,21 +141,344 @@ void restore_unsaved_values(sys::state& state) {
 	restore_cached_values(state);
 }
 
+void recalculate_markets_distance(sys::state& state) {
```
schombert, line 144: This function probably could be parallelized if it is called sufficiently often / in the hot path. However, there are also a number of places where it could probably run in parallel as a single-threaded function to other global updates (specifically it could probably run in parallel with almost anything that doesn't alter demographics, building construction, and province ownership, from a superficial inspection).

src/nations/nations.cpp
```diff
@@ -140,21 +141,344 @@ void restore_unsaved_values(sys::state& state) {
...
+				p_prev = p_current;
+			}
+			state.world.trade_route_set_land_distance(route, effective_distance / speed);
+		} else {
+			state.world.trade_route_set_land_distance(route, 99999.f);
+		}
+	});
+}
+
+void generate_sea_trade_routes(sys::state& state) {
```
schombert, line 251: This could probably also be parallelized, although with more effort. Basically instead of creating the new routes inside the main loop it would instead stick the information about the new route to be created into a thread-safe collection of some sort as it goes, and then do just the route creation at the end as a single-threaded operation. Again, not sure if that work is necessary or appropriate, however.

PR #1799, opened 2024-11-24
Check army and navy validity

SneakBug8:


src/map/map_state.cpp
```diff
@@ -286,6 +286,9 @@ void update_unit_arrows(sys::state& state, display_data& map_data) {
 	}
 
 	for(auto map_army : state.world.in_army) {
+		if(!state.world.army_is_valid(map_army)) {
+			return;
```
schombert, line 290: If the army possibly being invalid is a problem, this won't actually fix anything, because it can still become invalid after this check. The game is running in parallel with this function

SneakBug8:
Recent changes introduce dynamic infamy calculation based on country/state size.

schombert:
It looks like you have added target_state to the available cbs, but haven't actually done anything with it. It probably shouldn't be added until we have something that needs it.

schombert:
Also, you should run your changes to how infamy work by the players first. It isn't a big code change, but it is a large change to how the game works.

SneakBug8:
Probe of the community in Discord produced zero negative feedback.

PR #1809, opened 2024-11-30
Economy balance tweaks

ineveraskedforthis:
- Less greedy artisans
- Need weights are set to 1 to encourage consumption of actual goods instead of coal
- Cull trade routes more aggressively to avoid 1 Jan freeze
- Tweak some numbers

### DIFF
--- a/src/economy/economy.cpp
+++ b/src/economy/economy.cpp
@@ -995,6 +995,7 @@ void rebalance_needs_weights(sys::state& state, dcon::market_id n) {
 	auto zone = state.world.market_get_zone_from_local_market(n);
 	auto nation = state.world.state_instance_get_nation_from_state_ownership(zone);
 
+	/*
 	{
 		float total_weights = 0.f;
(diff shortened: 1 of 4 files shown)


src/economy/economy.cpp
```diff
@@ -1028,6 +1028,7 @@ void rebalance_needs_weights(sys::state& state, dcon::market_id n) {
 				}
 				auto& w = state.world.market_get_life_needs_weights(n, c);
 				w = ideal_weighting * state.defines.alice_need_drift_speed + w * (1.0f - state.defines.alice_need_drift_speed);
+				w = 1.f;
```
schombert, line 1031: Did you mean to write this? It seems to completely discard the previous calculation. If you want to keep the old calculation around, I suggest commenting it out so that the linter won't complain

ineveraskedforthis: Yes, the plan was to discard it. Will comment out old calculation.

PR #1812, opened 2024-12-10
intel JCC erratum codegen knob

Nivaturimika:
stop trashing the uop cache


schombert:
This doesn't affect enough CPUs to make sense to me

PR #1829, opened 2025-01-02
faster ai decision taking process

Nivaturimika:


schombert:
It looks like the decision PR is setup to run only 1/32nd of the decisions, as if it was being run every day. But, it isn't, so would you mind either making it run every day or run over all the decisions?

it also isn't doing the second validity check, which we added because it was possible for one decision taken on a day to invalidate another decision. If you collect all the ones to be taken first, that problem could reappear

PR #1838, opened 2025-01-04
Add files via upload for chinese too

MeizhouWan:
alice_budget_chart_laxis;? mean


schombert:
The laxis "Sat." stands for (needs) satisfaction. There is very little space for the label, unfortunately. I am not even sure where you will have room for two characters.

MeizhouWan:
6a29c5f is auto merge?

PR #1844, opened 2025-01-06
Update CMakeLists.txt

wonderinglostsoul44:
Factored Repeated Flags

Moved repeated warning disable flags (/wd4100 /wd4189 /wd4065, etc.) into a single BASE_MSVC_FLAGS. Debug vs. Release

Separated DEBUG_FLAGS and RELEASE_FLAGS for clarity, and appended them appropriately. Removed Repetition

Avoided rewriting the set(CMAKE_CXX_FLAGS) lines multiple times for each library, using a consistent approach across both freetype and harfbuzz. Essential Logic Retained

The if(WIN32) block sets flags for MSVC or Clang on Windows. The same flags are still there, just organized more succinctly. Comments and Organization

Added short comments (e.g., # Additional debug flags) to clarify which lines do what, improving readability without altering behavior. No Behavior Changes

The final compiled flags are effectively the same, preserving the original settings for debug vs. release and for Clang vs. MSVC.

### DIFF
--- a/dependencies/freetype/CMakeLists.txt
+++ b/dependencies/freetype/CMakeLists.txt
@@ -1,5 +1,7 @@
-message(STATUS "Fetching freetype ...")
 
+message(STATUS "Fetching freetype...")
+
+# Disable unwanted Freetype features
 set(FT_DISABLE_ZLIB ON)
 set(FT_DISABLE_BZIP2 ON)
 set(FT_DISABLE_PNG ON)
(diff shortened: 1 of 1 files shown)


schombert:
Hey, so we have been dealing with an influx of drive-by AI PRs lately. I encourage you to drop by the discord to say "hi" so that we know that there is a real person behind this. We are very friendly, I promise.

wonderinglostsoul44:
ok will do

PR #1878, opened 2025-01-23
Max loan, savings, reinvestment modifiers

SneakBug8:


src/ai/ai.cpp
```diff
@@ -674,6 +674,17 @@ void update_influence_priorities(sys::state& state) {
 			if(t.get_demographics(demographics::total) > state.defines.large_population_limit)
 				continue;
 
+			auto natid = state.world.nation_get_identity_from_identity_holder(n.nation);
+			for(auto prov_owner : state.world.nation_get_province_ownership(t)) {
+				auto prov = prov_owner.get_province();
+
+				for(auto core : prov.get_core_as_province()) {
+					if(core.get_identity() == natid) {
+						continue; // holds our cores
```
schombert, line 683: @SneakBug8 This will not do what you want. `continue` continues the inner-most containing loop. So this loop doesn't actually do anything.

src/gamestate/modifiers.hpp
```diff
@@ -196,8 +196,16 @@ namespace sys {
...
 	MOD_LIST_ELEMENT(124, flavor_tech_research_bonus, true, modifier_display_type::percent, "flavor_theory_tech_research_bonus") \
 	MOD_LIST_ELEMENT(125, seperatism, false, modifier_display_type::fp_two_places, "seperatism") \
+	MOD_LIST_ELEMENT(126, aristocrat_reinvestment, true, modifier_display_type::percent, "aristocrat_reinvestment") \
+	MOD_LIST_ELEMENT(127, capitalist_reinvestment, true, modifier_display_type::percent, "capitalist_reinvestment") \
+	MOD_LIST_ELEMENT(128, middle_class_reinvestment, true, modifier_display_type::percent, "middle_class_reinvestment") \
+	MOD_LIST_ELEMENT(129, farmers_reinvestment, true, modifier_display_type::percent, "farmers_reinvestment") \
+	MOD_LIST_ELEMENT(130, aristocrat_savings, true, modifier_display_type::percent, "aristocrat_savings") \
+	MOD_LIST_ELEMENT(131, capitalist_savings, true, modifier_display_type::percent, "capitalist_savings") \
+	MOD_LIST_ELEMENT(132, middle_class_savings, true, modifier_display_type::percent, "middle_class_savings") \
+	MOD_LIST_ELEMENT(133, farmers_savings, true, modifier_display_type::percent, "farmers_savings") \
```
schombert, line 206: @SneakBug8 It is really weird to have these specific modifiers for farmers and not for the lower class in general.

SneakBug8: Well, with Homesteading there is a Use Case when we want farmers to become active investors in society, but not any other lower class pop.

src/scripting/script_constants.hpp
```diff
@@ -470,14 +470,16 @@ EFFECT_BYTECODE_ELEMENT(0x01BD, set_culture_pop, 1) \
 EFFECT_BYTECODE_ELEMENT(0x01BE, change_party_name, 3) \
 EFFECT_BYTECODE_ELEMENT(0x01BF, change_party_position, 2) \
 EFFECT_BYTECODE_ELEMENT(0x01C0, diplo_points, 2) \
+EFFECT_BYTECODE_ELEMENT(0x01C1, suppression_points, 2) \
```
schombert, line 473: @SneakBug8 the size of the effect you have defined doesn't match the parsing, which adds only 1 cell

src/gamestate/modifiers.hpp
schombert: I don't understand why you would want that because there seems to be no away for farmers to receive profits, directly or indirectly from investments (i.e. factories). In any case, savings doesn't appear to be linked to actual investment (i.e. the investment pool).

SneakBug8: Savings go into national bank (loans)
While investments are going into investment pool.

PR #1892, opened 2025-02-02
Factory potentials & Economies of scale for factories

SneakBug8:


schombert:
@ineveraskedforthis you probably want to review this

src/economy/economy.cpp
```diff
@@ -1893,7 +1893,8 @@ float factory_throughput_multiplier(sys::state const& state, dcon::factory_type_
 	auto result = 1.f
 		* std::max(0.f, 1.f + national_t)
 		* std::max(0.f, 1.f + provincial_fac_t)
-		* std::max(0.f, 1.f + nationnal_fac_t);
+		* std::max(0.f, 1.f + nationnal_fac_t) *
+		1.f + (levels / 100.f) /* Economies of scale throughput bonus */;
```
ineveraskedforthis, line 1897: weird multiplication by 1
probably brackets are missed?

SneakBug8: Fixed

src/ai/ai.cpp
```diff
@@ -1127,13 +1127,14 @@ void get_craved_factory_types(sys::state& state, dcon::nation_id nid, dcon::mark
...
 	auto m = dcon::fatten(state.world, mid);
+	auto sid = m.get_zone_from_local_market();
 
 	auto const tax_eff = nations::tax_efficiency(state, n);
 	auto const rich_effect = (1.0f - tax_eff * float(state.world.nation_get_rich_tax(n)) / 100.0f);
 
 	if(desired_types.empty()) {
 		for(auto type : state.world.in_factory_type) {
-			if(n.get_active_building(type) || type.get_is_available_from_start()) {
+			if(command::can_begin_factory_building_construction(state, nid, sid, type, false) || command::can_begin_factory_building_construction(state, nid, sid, type, true)) {
```
schombert, line 1137: @SneakBug8 This is changing a very cheap check into two very expensive checks. If the expensive checks are necessary, the code should be written so that they are taken a minimal number of times (for example, it shouldn't be doing duplicate work checking for both upgrades and new construction)

src/gamestate/dcon_generated.txt
```diff
@@ -467,6 +467,11 @@ object {
 		type{ bitfield }
 		tag{ scenario }
 	}
+	property{
+		name{ uses_potentials }
```
schombert, line 471: why is this being stored in the factory type when everything else is tied to the commodity? I don't think that your syntax for this feature allows it to attach to some factory types and not others if they happen to produce the same output

PR #1911, opened 2025-02-14
Devbranch7 into main

SneakBug8:
- Local commodities
- AI upgrades ships
- Ticking score adjustments
- disallow_naval_trade, disallow_land_trade modifiers for mods (uncivs)
- Technologies costing leadership for mods (military doctrines)
- Maneuver as a modifier (Vic2 compat)
- Provid console command


assets/localisation/en-US/alice.csv
```diff
@@ -221,6 +221,7 @@ month;month
 a_port;a port
 national_rank;national rank
 att_technology;technology
+TECHNOLOGY_LEADERSHIP_POINTS;Leadership points:
```
schombert, line 224: the convention is that all PA keys are in lower case

assets/localisation/en-US/alice.csv
```diff
@@ -1323,12 +1325,25 @@ price_origin;Origin price
 price_target;Target price
 trade_distance;Distance
 volume;Volume
-desired_volume; Desired Volume
+desired_volume;Actual Volume
```
schombert, line 1328: if this is correct, then the key itself should also be changed to avoid very confusing C++ code

docs/super_console.md
```diff
@@ -23,6 +23,7 @@ And here is a list of some more specialized console functions that you can use
 - `clear` : removes old text form the console window
 - `spectate` : switches the game to spectator mode (use `change-tag` to resume playing)
 - `true fps` : turns the visible FPS counter on. A value of `false` will instead turn it off
+- `provid`: Turns on and off display of content province ids in tooltips
```
schombert, line 26: this should probably take a boolean parameter like the other options

src/ai/ai.cpp
```diff
@@ -4145,42 +4145,43 @@ bool unit_on_ai_control(sys::state& state, dcon::army_id a) {
...
-	to_delete.clear();
 	for(auto n : state.world.in_nation) {
 		if(n.get_is_player_controlled())
 			continue;
 		if(n.get_is_at_war() == false && nations::is_landlocked(state, n)) {
 			for(auto v : n.get_navy_control()) {
 				if(!v.get_navy().get_battle_from_navy_battle_participation()) {
 					for(auto shp : v.get_navy().get_navy_membership()) {
-						to_delete.push_back(shp.get_ship().id);
+						state.world.delete_ship(shp.get_ship());
```
schombert, line 4176: deleting ships while you are iterating over them is a bug

src/culture/culture.cpp
```diff
@@ -36,6 +36,20 @@ std::string get_tech_category_name(tech_category t) {
 	return "none";
 }
 
+std::vector<culture::tech_category> get_active_tech_categories(sys::state& state) {
```
schombert, line 39: I don't understand why this function exists, since we already have a vector of the tech categories in the form of state.culture_definitions.tech_folders, which you use below

src/economy/economy.cpp
```diff
@@ -3939,10 +3939,18 @@ void daily_update(sys::state& state, bool presimulation, float presimulation_sta
...
 		auto is_land_route = state.world.trade_route_get_is_land_route(trade_route);
+		auto same_nation = controller_capital_A == controller_capital_B;
 
-		is_sea_route = is_sea_route && !is_A_blockaded && !is_B_blockaded;
+		// Ban international sea routes or international land routes based on the corresponding modifiers
+		auto A_bans_sea_trade = state.world.nation_get_modifier_values(n_A, sys::national_mod_offsets::disallow_naval_trade) > 0.f;
+		auto B_bans_sea_trade = state.world.nation_get_modifier_values(n_B, sys::national_mod_offsets::disallow_naval_trade) > 0.f;
+		auto sea_trade_banned = A_bans_sea_trade || B_bans_sea_trade;
+		auto A_bans_land_trade = state.world.nation_get_modifier_values(n_A, sys::national_mod_offsets::disallow_land_trade) > 0.f;
+		auto B_bans_land_trade = state.world.nation_get_modifier_values(n_B, sys::national_mod_offsets::disallow_land_trade) > 0.f;
```
schombert, line 3949: if you are using disallow_land_trade like a boolean rule, it should probably be a boolean rule. Otherwise, it will show up extremely weirdly in the ui

src/gamestate/dcon_generated.txt
```diff
@@ -1350,6 +1355,12 @@ object {
 	tag{ scenario }
 	tag{ save }
 
+	property{
+		name{ provid }
```
schombert, line 1359: I deliberately don't store the original province ids in the scenario since they are meaningless in game. When people say that they want to know the province id with a console command, they generally want a number that they can use with other console commands, not the number form the game files. Giving them the number from the game files will be frustrating since they can't actually use it to do change owner, etc

src/gamestate/modifiers.hpp
```diff
@@ -268,6 +270,7 @@ struct unit_variable_stats {
 	float siege_or_torpedo_attack = 0.0f;
 	float reconnaissance_or_fire_range = 0.0f;
 	float discipline_or_evasion = 0.0f;
+	float maneuver = 0.0f;
```
schombert, line 273: does any unit have both a maneuver value and a support value?

src/gui/gui_modifier_tooltips.cpp
```diff
@@ -44,6 +45,8 @@ std::string format_modifier_value(sys::state& state, float value, modifier_displ
 		return(value >= 0.f ? "+" : "") + text::format_float(value, 2);
 	case modifier_display_type::fp_three_places:
 		return (value >= 0.f ? "+" : "") + text::format_float(value, 3);
+	case modifier_display_type::yesno:
```
schombert, line 48: ok, but it will still be represented internally as a floating point, which means that you could have one modifier with a positive value showing "yes" and another with a negative value showing "no" and no way to know whether the sum will be "yes" or "no"

schombert: If you want to extend the nation's rules in some way, you should extend the rules

src/gamestate/dcon_generated.txt
SneakBug8: I remember people explicitly looking for Content province id to modify events/history files easier. Anyway, tooltip shows both PA id and content ID as two rows.

docs/super_console.md
SneakBug8: Then it still will confuse vanilla V2 players expecting to type just `provid`

src/economy/economy.cpp
SneakBug8: I don't think we have boolean modifiers (like other Pdx games do) and AFAIK modifiers can't change rules, so a true/false modifier type is necessary

src/gamestate/modifiers.hpp
SneakBug8: Why would it matter? We used base unit type values previously.

src/culture/culture.cpp
SneakBug8: Folders are created for all potential categories incl flavour from every mod. Previously scenario generator used hardcoded 5 categories and would crash when 6th was used. Now this method finds which categories actually have some tech in them.

src/gamestate/modifiers.hpp
schombert: Well, if no unit has both, it could double up, like many of the other values do

docs/super_console.md
schombert: since all the other console commands have been changed, I imagine that they are already confused.

src/economy/economy.cpp
schombert: Fundamentally, modifiers are designed as an additive system. But you can't add one or more "bans trade" modifiers together in a meaningful way. I understand that putting it as a modifier is the easiest way to achieve the desired result. However, it isn't a good way to achieve the desired result. If you want to come up with a way to extend the political rules, I encourage you to do that.

src/gamestate/modifiers.hpp
ZombieFreak115: Some units such as artillery (has 1 maneuver and support), and planes (has 2 manuever and support) by default.

assets/localisation/en-US/alice.csv
SneakBug8: Done

assets/localisation/en-US/alice.csv
SneakBug8: Done

docs/super_console.md
SneakBug8: Done

src/ai/ai.cpp
SneakBug8: Fixed

src/economy/economy.cpp
SneakBug8: OK. Then let's keep it AS IS for now and see if we can find any more cases for yes/no modifiers since it'd be quite a sizeable change. I don't expect any mods to use these modifiers in any way other than `disallow_land_trade = 1` and then any number of additions will work as expected. Let me add a disclaimer to the docs.

PR #1918, opened 2025-02-22
Implenting allowing reserves in battle to reinforce/reorg, among other things.

ZombieFreak115:
The new system works as follows:

All units in a battle, which are in the reserves (not in the battle-screen doing damage), will receive org and reinforcement ticks as any other land unit would.
The goals for this would be to allow slower, grindier battles in the lategame (similar to ww1), while still having fast-paced early game battles, as combat width is larger, and thus fewer units would be in reserve than on the battle screen. As a predominantly MP player, it would also emulate the "cycling" and reinforcement of units, but without any micro needed.

The reinforcement rate of units in a battle is the same as regular units, but the location-based reinforcement bonus (by default 200% in owned territory, 100% in friendly, 10% in hostile) will not be dependent on the province the battle is taking place in, but rather any suitable adjacent province. This means that if a battle is surriounded completely by enemy territory, you would receive low reinforcement compared to if it was adjacent to friendly territory. It will pick the highest reinforcement bonus of any of the provinces possible.
Enemy units will block the province from giving a bonus, so a "encircled" battle would not receive any reinforcements whatsoever.
Org regain is currently exempt from the location-based reinforcement bonus, as it is currently for units not in a battle.

The purpose of this is to prevent a battle from being able to continuously reinforce if it is behind enemy lines, or encircled. 
When units reach a strenght or org threshold (currently 0 str, or >10% org, dictated by some new defines) in a battle, they will be kicked back to the reserve pool, where they can reinforce/regain org and be available for deployment again once they go above said threshold. Each day the reserve pool is sorted according to unit strength, and so will try to deploy the units with the highest strength first, which also meets the >10% org requirement.

In addition, some new UI elements have been added to the battle screen to show reinforcement rate, and num of reserve regiments. 
The battle "progress bar" has also been altered to ignore regiments which are in the reserve, but cannot be deployed, and is depentant on the total strength of all units, rather than the raw regiment count. Also changed it so brigades with no pop attached cannot reinforce under any circumstance, and added some more reinforce-location-modifiers which were absent.


src/military/military.cpp
```diff
@@ -5791,24 +5787,101 @@ void apply_regiment_damage(sys::state& state) {
...
+					// if the dead brigade with not enough supporting pop is in the reserves, remove it from the reserves first before deletion
+					if(!controller || state.world.pop_get_size(pop_backer) < state.defines.pop_min_size_for_regiment) {
+						for(uint32_t j = reserves.size(); j-- > 0;) {
+							if(reserves[j].regiment == s) {
+								std::swap(reserves[j], reserves[reserves.size() - 1]);
+								reserves.pop_back();
+								break;
+							}
 						}
+						state.world.delete_regiment(s);
```
schombert, line 5799: unfortunately, deleting a regiment while you are iterating over a set of regiments `auto reg : state.world.army_get_army_membership(a)` will introduce bugs. I suggest simply not adding such a regiment (i.e. zero strength and no backing pop) to the reserves, echoing the original logic.

src/gui/gui_land_combat.hpp
```diff
@@ -631,6 +639,129 @@ class attacker_combat_modifiers : public overlapping_listbox_element_base<lc_mod
...
+		text::add_to_layout_box(state, contents, box, text::prettify(int64_t(reserve_regs)), text::text_color::black);
+		text::close_layout_box(contents, box);
+	}
+	tooltip_behavior has_tooltip(sys::state& state) noexcept override {
+		return tooltip_behavior::variable_tooltip;
+	}
+
+	void update_tooltip(sys::state& state, int32_t x, int32_t y, text::columnar_layout& contents) noexcept override {
+		auto b = retrieve<dcon::land_battle_id>(state, parent);
+		uint32_t reserve_regs = uint32_t(military::get_reserves_by_side(state, b, true).size());
```
schombert, line 735: This is ... a bit silly. You are populating a vector only to discard it just to get its size. I suggest adding a function to count the reserves directly instead without generating the vector.

src/military/military.cpp
```diff
@@ -5791,24 +5787,101 @@ void apply_regiment_damage(sys::state& state) {
...
+	for(uint32_t i = reserves.size(); i-- > 0;) {
+		if(reserves[i].regiment == reg) {
+			return true;
+		}
+	}
+	return false;
+}
+// uses insertion sort; we assume the reserves won't have a huge amount of items, and since reserves are only inserted/popped from the back, they will probably stay partially sorted even when changes happen
+// implementation of what it sorts by can change, for now it sorts by strength and puts the highest str brigades in front of the queue
+void sort_reserves_by_deployment_order(sys::state& state, dcon::dcon_vv_fat_id<reserve_regiment> reserves) {
```
schombert, line 5871: I strongly suggest against reinventing a sorting function. You can use std::sort directly on the reserves set by giving it a custom comparator to sort as desired (https://en.cppreference.com/w/cpp/algorithm/sort overload 3). std::sort is very well optimized

src/military/military.cpp
```diff
@@ -6249,10 +6330,15 @@ void update_land_battles(sys::state& state) {
 			std::swap(def_front[0], def_front[1]);
 		}
 
+		// sorts the reserves by some criteria so the most fit brigades for deployment are pushed to the front of the queue
+		sort_reserves_by_deployment_order(state, reserves);
+
+		// won't use troops from the reserve which are > 10% org, or zero strength
```
schombert, line 6336: presumably < 10% org, although this is doubly out-of-date since you have switched the implementation to using a define

src/military/military.cpp
```diff
@@ -7536,71 +7629,308 @@ float unit_get_strength(sys::state& state, dcon::regiment_id regiment_id) {
...
+	
+
+}
+
+
 // Calculates max reinforcement for units in the army
 float calculate_army_combined_reinforce(sys::state& state, dcon::army_id a) {
 	auto ar = fatten(state.world, a);
-	if(ar.get_battle_from_army_battle_participation() || ar.get_navy_from_army_transport() || ar.get_is_retreating())
+	if(ar.get_navy_from_army_transport() || ar.get_is_retreating() || ar.get_black_flag())
```
schombert, line 7780: In the change you made for recovering org, you specifically prevented units on the battle lines from recovering org, but you make no similar check here for recovering strength. Intentional?

ZombieFreak115: The calculate_army_combined_reinforce function is used to calculate reinforcement at the army level, but some regiments in the same army may be on the battle lines, while others are in reserve, that condition is left out here. It is present in regiment_calculate_reinforcement, which is called by reinforce_regiments together with calculate_army_combined_reinforce.

src/military/military.cpp
```diff
@@ -7536,71 +7629,308 @@ float unit_get_strength(sys::state& state, dcon::regiment_id regiment_id) {
 float unit_get_strength(sys::state& state, dcon::ship_id ship_id) {
 	return state.world.ship_get_strength(ship_id);
 }
+dcon::land_battle_id get_province_battle(sys::state& state, dcon::province_id prov) {
+	for(auto b : state.world.province_get_land_battle_location(prov)) {
+		if(b.get_battle() && b.get_location().id == prov) {
+			return b.get_battle();
```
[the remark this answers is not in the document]
ZombieFreak115: It was an un-used function that should have been removed. I removed it

src/military/military.cpp
ZombieFreak115: Updated the comments

src/military/military.cpp
ZombieFreak115: Did as you suggested as swapped to using std::sort with a comparer.

src/gui/gui_land_combat.hpp
ZombieFreak115: Good point, fixed.

src/military/military.cpp
ZombieFreak115: Im not sure i understand it completly, wouldn't the `state.world.delete_regiment(s);` on line 5808 then also be an issue since it is also deleting the regiment s while iterating over it? Which seems to not be related to the reserves brigades.

schombert: Ah, with things folded away in the context I misread which loop we were in here. Yes, this is fine because the loop is blindly iterating over all regiments, and not over a particular collection of regiments that would be altered by regiment deletion.

PR #1921, opened 2025-02-27
Max organisation implemenation of military units & other military QoL and fixes

ZombieFreak115:
This PR aims to add a simulation of the "max org" stat which is presently not implemented (any +org techs don't have any effect) by decreasing the amount of org damage taken in combat the higher your max org is, while also slowing relative org regen. Max org from tech and from individual generals are considered in the calculation.

In addition, it also adds logic to ensure your backline support regiments are the highest strength possible of all reserves, to avoid annoying micro situations where your art is stuck at low str doing litle damage. Also implemented maneuver working from the backrow, which was missing (this might be intended, but didnt make sense to me, since support units such as planes have noticably high maneuver)

It fixes a bug relating to maneuver in which the maneuver variable stat was not being loaded in properly, resulting in 0 maneuver for all units in combat. Also fixes a small oversight in the maneuver-target picking code, in which meant the unit in position 1 would not be able to attack diagonally to position 0.

Changed implementation of the detection of whether a unit should be blocked from moving over a blockaded sea strait, the previous impl. lead to some inconsistencies, as it relied on the "port_to" province, which can sometimes be diffrent depending on the strait in question.

Fixed a bug in army movement in which if the army was queued to move into "non allowed" territory, it would prevent you from stopping the movement untill it reached its destination. The bug was seemingly caused by the can_move_army function always assuming a "shift-click" input, and calculating the path from the province the army was queued to move to, so i added the "reset" parameters to those functions as well to determine if it should be treated as a "shift click" movement.

Added extra check to prevent a movement order which goes over a blockaded sea strait.


src/gamestate/commands.cpp
```diff
@@ -3599,18 +3599,29 @@ std::vector<dcon::province_id> can_move_army(sys::state& state, dcon::nation_id
...
+	return calculate_army_path(state, source, a, last_province, dest, reset);
 }
 
 
-std::vector<dcon::province_id> calculate_army_path(sys::state& state, dcon::nation_id source, dcon::army_id a, dcon::province_id last_province, dcon::province_id dest) {
-	if(last_province == dest)
-		return std::vector<dcon::province_id>{};
+std::vector<dcon::province_id> calculate_army_path(sys::state& state, dcon::nation_id source, dcon::army_id a, dcon::province_id last_province, dcon::province_id dest, bool reset) {
+	if(last_province == dest) {
+		// if reset is true, it means the last_province is the same as the current province of the army, so create an army path which just keeps the army there
```
schombert, line 3616: how well has this been tested? Stay-in-place is currently represented by not having a path. Moving into the current province was never expected to happen, and so may be a source of bugs

src/gamestate/commands.cpp
```diff
@@ -5831,7 +5842,7 @@ bool can_perform_command(sys::state& state, payload& c) {
 		return can_send_peace_offer(state, c.source);
 
 	case command_type::move_army:
-		return can_move_army(state, c.source, c.data.army_movement.a, c.data.army_movement.dest).size() != 0;
+		return can_move_army(state, c.source, c.data.army_movement.a, c.data.army_movement.dest, c.data.army_movement.reset).size() != 0;
```
schombert, line 5845: The way this works in the ui currently is that a move to the invalid province is interpreted as a command to clear the current movement of the army and leave it in place. Thus, a non-shift movement command is encoded as two commands. The first clears the existing path, and the second creates a new movement path from the now-stopped unit.

src/gamestate/commands.cpp
schombert: Furthermore, I am not sure how this will interact with arrival times. A moving unit has both a path and an arrival time to the next province in the path. What arrival time does moving to the province it is already in get? Is it appropriately in the future so that it can be marked as "arrived" and have its path cleared?

src/provinces/province.cpp
```diff
@@ -1073,6 +1073,32 @@ void change_province_owner(sys::state& state, dcon::province_id id, dcon::nation
 				event::slot_type::state, new_owner, -1, event::slot_type::none);
 	}
 }
+// returns true if a strait between the two provinces are blocked by an enemy navy from the perspective of thisnation
+// Current impl is to check if any sea province which borders both to and from is blockaded by an enemy ship. Will probably want to import the missing adj province data eventually and use that instead
+bool is_strait_blocked(sys::state& state, dcon::nation_id thisnation, dcon::province_id from, dcon::province_id to) {
+	auto path_bits = state.world.province_adjacency_get_type(state.world.get_province_adjacency_by_province_pair(to, from));
+	if((path_bits & province::border::non_adjacent_bit) != 0) { // strait crossing
+		for(auto adj : state.world.province_get_province_adjacency(from)) {
```
schombert, line 1081: Unfortunately, this is extremely inefficient (a loop within a loop within a loop within a loop (the fourth+ level is inside at_war_with)), which would be fine if it was called rarely, but unfortunately it is placed inside some of the path finding code which ends up being called quite often (not just for moving units from one place to other, but for the AI considering whether a unit can reach one place from another, for example). You will need to come up with some faster way of doing this. In particular, if you want to mark a particular sea tile as the one that the strait passes through and check only it (this is what V2 does, and is encoded in the extra adjacency definitions file, although PA currently ignores it), that would be a start. But, at the end of the day, it should ideally come closer to the current very cheap check. For example, you could do something like requiring the strait crossing to be not blockaded and in friendly control from the side the unit is leaving from.

ZombieFreak115: Yeah this is indeed too high complexity for a method which is called this often. I will probably try to implement reading the sea province from the vic2 adjacencies file and use that.

ZombieFreak115: Imported the vic2 adjacencies. The seralization/deseralization should work from my understanding of the code, but if you notice im missing something let me know.

Also refactored some unrelated reinforcement code to fix some bugs caught during testing.

src/gamestate/commands.cpp
ZombieFreak115: This is true, however the problem is, in `selected_units_control` it uses `can_move_army` to check if the command should be queued up in the first place, but the `can_move_army` function does not verify if we are checking for holding shift, or not. The result is that for the player, it always checks as if you are holding shift when it checks if the move is legal or not.

This means eg. that if you are queued to move into someone's territory, and they revoke mil access, then you cannot "un-queue" this action because `can_move_army` checks from the perspective of the province the army is moving to. The purpose of the "reset" param is thus to check if we are to move from the province we are moving to, or from the province we are in.

The reset parameter is being passed here in order to keep consistency with the afrorementioned check.

schombert: That's fair, and I will merge solution if you have worked out that none of the new edge cases I described above can cause a problem. However, it might be simpler to call the pathfinding logic directly in the ui to see whether the appropriate path is possible rather than relying on the can_move function's logic, which wouldn't require changing the command logic

src/gamestate/commands.cpp
ZombieFreak115: I have tested it pretty well, ingame and stepping though the code, and the movement to the current province did work, however it is admittedly abit wonky, and has a minor side-effect of making it look like the unit is moving for one day after you cancel a movement (as it takes one day to "move" to its own province), after which it is cleared. 

I have adressed this oddity in a new commit, which will cancel the movement properly (clearing the path, aswell as resetting its arrival timer) when it detects that a command is intented to cancel the current movement.

src/provinces/province.hpp
```diff
@@ -22,7 +22,7 @@ struct global_provincial_state {
 	std::vector<dcon::province_id> canal_provinces;
 	ankerl::unordered_dense::map<dcon::modifier_id, dcon::gfx_object_id, sys::modifier_hash> terrain_to_gfx_map;
 	std::vector<bool> connected_region_is_coastal;
-
+	ankerl::unordered_dense::map<dcon::province_adjacency_id, dcon::province_id, sys::province_adjacency_hash> sea_adjacencies;
```
schombert, line 25: I don't understand why this is necessary, an extra province could be added to the adjacency itself (as a new property) and just left un-populated for most of them. This would be a guaranteed O(1) lookup, and would be much easier to serialize as well.

src/military/military.cpp
```diff
@@ -8026,8 +8029,8 @@ bool get_allied_prov_adjacency_reinforcement_bonus(sys::state& state, dcon::prov
 		if(province_has_enemy_unit(state, prov, our_nation)) {
 			return false;
 		}	
-		// checks if the province controlled by us, or is controlled by someone who is at war with the owner of location and it is not rebels
-		else if(prov_controller == our_nation || (!prov_controller && are_at_war(state, location_controller, prov_controller))) {
```
schombert, line 7880: !prov_controller also covers uncolonized land. I think this means that your logic treats uncolonized land as enemy controlled, except possibly for rebels? not sure if this matters

enc_temp_folder/2f4f608e1339c8eb6d9bb3293d9dfca8/commands.cpp
```diff
@@ -0,0 +1,6442 @@
+#include "commands.hpp"
```
schombert, line 1: you have accidentally committed a temp file here

src/provinces/province.hpp
ZombieFreak115: This was my initial idea aswell, to add another property to each adjacency. I thought at the time that adding it to each adjacency would be wasteful, since it wouldnt be used for most adjacencies, however i now do realize that using a map with the sea adjacencies is prehaps even more wasteful, and overcomplicates things. I will see to simplify it as you suggest when i can.

enc_temp_folder/2f4f608e1339c8eb6d9bb3293d9dfca8/commands.cpp
ZombieFreak115: My bad, will fix

src/military/military.cpp
ZombieFreak115: Good catch. I failed to account for the fact that uncolonized and rebels return the same nation id by both ownership and controllership. I added some checks to take account for that.

PR #1935, opened 2025-05-02
Small adjustments to migration and bureacracy.

ineveraskedforthis:
- Add additional multiplier to increase a chance of migration to province with high wage
- Add base chance for primary culture bureaucracy to migrate to problematic regions (resolves issues with far away regions with diverging administration costs)
- Use state capital bureaucracy instead of provincial one (has to be done as bureaucracy does not migrate to random provinces)
- Minor fix for trade: allow trade if at least one party unlocked related commodity

### DIFF
--- a/src/economy/demographics.cpp
+++ b/src/economy/demographics.cpp
@@ -8,6 +8,7 @@
 #include "triggers.hpp"
 #include "ve_scalar_extensions.hpp"
 #include "container_types.hpp"
+#include "economy_stats.hpp"
 
 // #define CHECK_LLVM_RESULTS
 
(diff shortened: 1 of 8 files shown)


src/economy/demographics.cpp
```diff
@@ -3264,7 +3282,37 @@ dcon::province_id get_province_target_in_nation(sys::state& state, dcon::nation_
 	auto weights_buffer = state.world.province_make_vectorizable_float_buffer();
 	float total_weight = 0.0f;
 
+	auto origin = state.world.pop_get_province_from_pop_location(p);
 	auto pt = state.world.pop_get_poptype(p);
+
+	auto labor_type = economy::labor::no_education;
```
schombert, line 3288: This calculation also appears below. You may want to factor it out into a common function that goes from pop to labor type, or something of that sort

schombert:
does any of the description information in gui/alice_gui.cpp need to be updated? There are describe_... functions in there that populate tooltips with migration numbers, etc. Also, is the labor type weighting reflected in the province weight in the pop details window and associated tooltip?

PR #1937, opened 2025-05-05
Topbar Trade panel changes

Skuldran:
-Display trade volume as value of goods, instead of abstract goods units -Changed to only consider net-export/import to each nation


schombert:
FYI, this doesn't build

Skuldran:
I noticed that yesterday. Did some cleanup in the code and somehow missed to match the variable names. Should have time to fix it today.

PR #1938, opened 2025-05-07
Add local control.

ineveraskedforthis:
- Nations hire bureaucracy in the capital or one of the local administrations to produce "control"
- Control is propagated along market connections and provinces adjacency
- Control decays in provinces depending on local conditions
- Population consumes control. Ratio of satisfied control decides a lot of local effects: building costs, effective tax rates and local tariffs
- Rework public works and social spending sliders to use a percentage of the budget which simplifies related balance and reduces the need to fiddle with these sliders.


src/ai/ai.cpp
```diff
@@ -3090,23 +3104,53 @@ void update_cb_fabrication(sys::state& state) {
...
 			possible_targets.clear();
+			float total_weight = 0.f;
 			for(auto i : state.world.in_nation) {
 				if(valid_construction_target(state, n, i)
 				&& !military::has_truce_with(state, n, i)) {
-					possible_targets.push_back(i.id);
-					if(!i.get_is_civilized())
-						possible_targets.push_back(i.id); //twice the chance!
+
+					auto weight = 1.f / province::sorting_distance(
```
schombert, line 3114: sorting_distance preserves the relative order between provinces, but is not proportional to their actual distance (it is just the dot product without the acos)

PR #1947, opened 2025-05-15
War exhaustion in AI peace offer evaluation

SneakBug8:


SneakBug8:
1) In willingness_factor, every month of the war beyond the first year adds 0.82 points. War exhaustion is capped at a 100 and with blockades only (without actual occupations) one won't get more than +0.5 war exhaustion per month.

Considering that at a 100 WE country is basically dead with constant rebellions; I believe every AI should be willing to white peace out at some reasonable amount of WE as it wasn't able to overtake the enemy in this time period too.

2) Reviewing all mentions of "war_exhaustion" for getters, I couldn't find any mentions of WE in AI war declarations. Added with the latest commit.

PR #1948, opened 2025-05-15
Assert to help map makers

SneakBug8:


src/map/map.cpp
```diff
@@ -2725,7 +2725,7 @@ emfx::xac_pp_actor_material_layer get_diffuse_layer(emfx::xac_pp_actor_material
 	}
 	return mat.layers.empty()
 		? emfx::xac_pp_actor_material_layer{}
-		: mat.layers[0];
+	: mat.layers[0];
```
schombert, line 2728: and it has deleted the indentation here

src/map/map.cpp
```diff
@@ -590,8 +590,8 @@ void display_data::render(sys::state& state, glm::vec2 screen_size, glm::vec2 of
 						province::border::coastal_bit
 						| province::border::national_bit
 						| province::border::impassible_bit
+						)
```
schombert, line 593: and this would incorrectly match the indentation of the corresponding open parenthesis

PR #1952, opened 2025-05-16
Fixes to crisis mechanics

SneakBug8:


SneakBug8:
Yeah, claim crises need further work for sure. However, so far I don't see why liberation crises won't fire for claims and behave in the same way (although not exactly working as planned).

PR #1954, opened 2025-05-17
change_factory_limit effect

SneakBug8:


src/scripting/effects.cpp
```diff
@@ -5331,6 +5331,19 @@ uint32_t ef_change_party_position(EFFECT_PARAMTERS) {
 	return 0;
 }
 
+uint32_t ef_change_factory_limit(EFFECT_PARAMTERS) {
+	auto com = trigger::payload(tval[1]).com_id;
+	auto change = trigger::payload(tval[2]).value;
+
+	if(com) {
+		auto p = trigger::to_prov(primary_slot);
+		ws.world.province_get_factory_max_size(p, com) += (int)change;
```
schombert, line 5340: what happens if the max size becomes negative or if it is reduced to less than the size of factories that are already present?

SneakBug8: 1) if the max size becomes negative - no new buildings/upgrades will be available after that point.
2)  if it is reduced to less than the size of factories that are already present - there is no validation for that. Existing levels will operate in the state.

src/scripting/effects.cpp
```diff
@@ -5332,10 +5332,14 @@ uint32_t ef_change_party_position(EFFECT_PARAMTERS) {
...
 
 uint32_t ef_change_factory_limit(EFFECT_PARAMTERS) {
+	/*
+	Use Case: increase the factory limit for the commodity during runtime (e.g. natural resources discovery).
+	*/
 	auto com = trigger::payload(tval[1]).com_id;
 	auto change = trigger::payload(tval[2]).value;
 
 	if(com) {
+		assert(change > 0 && "Decreasing factory limit is not implemented.");
```
schombert, line 5342: if you want it to always be positive, you can presumably rule out 0 and negative in parsing by emitting an error

SneakBug8:
Moved the assert to a parsing warning.

PR #1956, opened 2025-05-17
Smarter AI

SneakBug8:
1. AI properly research doctrines costing leadership points
2. AI won't ally weak neighbours
3. A bit smarter construction
4. AI army size estimation
5. Remove extra break
6. AI won't upgrade units if shortages

### DIFF
--- a/src/military/military.cpp
+++ b/src/military/military.cpp
@@ -156,8 +156,28 @@ bool is_big_ship_better(sys::state& state, dcon::nation_id n, dcon::unit_type_id
 	return false;
 }
 
-dcon::unit_type_id get_best_infantry(sys::state& state, dcon::nation_id n, bool primary_culture) {
-	dcon::unit_type_id curbest;
+// Evaluate if the nation can reasonably build this unit type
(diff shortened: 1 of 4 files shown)


src/military/military.cpp
```diff
@@ -3544,15 +3607,9 @@ void implement_war_goal(sys::state& state, dcon::war_id war, dcon::cb_type_id wa
...
-		else {
-			for(auto n : state.world.in_nation) {
-				auto rel = state.world.nation_get_overlord_as_subject(n);
-				auto overlord = state.world.overlord_get_ruler(rel);
-
-				if(overlord == target) {
-					state.world.overlord_set_ruler(rel, from);
-				}
+		} else {
+			for(auto sub : state.world.nation_get_overlord_as_ruler(target)) {
```
schombert, line 3611: this will create a bug since it is altering the set of subjects of target as it is iterating over them

SneakBug8:
Pushed a fix. I wonder if that caused problems as well then:
```
for(auto sub : state.world.nation_get_overlord_as_ruler(target)) {
				nations::release_vassal(state, sub);
			}
```

PR #1963, opened 2025-05-21
Province tiles UI

SneakBug8:
UI window on province level displaying an 8x8 grid of tiles to display all the province buildings. For now, the list is generated from existing feature structure:
  - A tile per every RGO commodity with sizeable max employment.
  - A tile per every built factory (all levels into one).
  - A tile per every commodity that has free (unused) resource potential.
  - Every province building takes a tile.
  - Regiment barracks take tiles last.
  
In tooltips, we display the info relevant to this building:
  - RGO
  - Factory
  - Regiment
  - Province building

When clicked:
- Factory opens context menu allowing upgrade the said factory
- RGO opens economy scene allowing to view where similar goods are produced/consumed (the only way to open the scene beyond the hotkey).
- Regiment selects the army to which the regiment belongs

With context windows player can:
  - [Done] Upgrade factories if Rules allow
  - [To Do] Delete factories if Rules allow


SneakBug8:
Resolved merge conflicts

src/economy/construction.hpp
```diff
@@ -73,4 +73,11 @@ float calculate_factory_refit_money_cost(sys::state& state, dcon::nation_id n, d
 
 float estimate_construction_spending_from_budget(sys::state& state, dcon::nation_id n, float current_budget);
 float estimate_construction_spending(sys::state& state, dcon::nation_id n);
+
+// Check rules for factories in colonies
+bool can_build_in_colony(sys::state& state, dcon::province_id p);
```
schombert, line 78: these function should be renamed to something like "can build factory in province" since that is what they are testing

src/ai/ai.cpp
```diff
@@ -1340,10 +1340,12 @@ void get_state_craved_factory_types(sys::state& state, dcon::nation_id nid, dcon
 	auto treasury = n.get_stockpiles(economy::money);
 	auto wage = state.world.province_get_labor_price(pid, economy::labor::basic_education) * 2.f;
 
-
 	if(desired_types.empty()) {
 		for(auto type : state.world.in_factory_type) {
 			if(n.get_active_building(type) || type.get_is_available_from_start()) {
+				if(!economy::can_build_in_colony(state, sid)) {
```
schombert, line 1346: this suggests that you are allowing this function to be called in colonial states and then stopping it from doing anything here. This is inefficient -- if the game rules preventing factory construction in colonies are in place, the function should probably never be called in the first place. And, if you did want to prevent the function from proceeding here, you would want to put this test outside the loop, since it is invariant over factory types.

src/ai/ai.cpp
```diff
@@ -1368,6 +1370,11 @@ void get_state_desired_factory_types(sys::state& state, dcon::nation_id nid, dco
 	if(desired_types.empty()) {
 		for(auto type : state.world.in_factory_type) {
 			if(n.get_active_building(type) || type.get_is_available_from_start()) {
+				// Check rules for factories in colonies
+				if(!economy::can_build_in_colony(state, pid, type)) {
+					continue;
```
schombert, line 1375: see previous comment

src/ai/ai.cpp
```diff
@@ -1536,8 +1543,10 @@ void update_ai_econ_construction(sys::state& state) {
 			static std::vector<dcon::province_id> ordered_provinces;
 			ordered_provinces.clear();
 			for(auto p : n.get_province_ownership()) {
-				if(p.get_province().get_state_membership().get_capital().get_is_colonial() == false)
+				// Check rules for factories in colonies
+				if(economy::can_build_in_colony(state, p.get_province())) {
```
schombert, line 1547: Note that this is not strictly the same as the check it is replacing, since that check consults the *capital* of the containing state. There have been weird cases where colonial provinces get placed in non-colonial states, or vice versa

src/economy/economy_production.cpp
```diff
@@ -2003,7 +2003,58 @@ float rgo_income(sys::state& state, dcon::commodity_id c, dcon::province_id id)
 		* state.world.market_get_supply_sold_ratio(mid, c);
 }
 
+commodity_set rgo_inputs_actual(sys::state& state, dcon::nation_id n, dcon::market_id m, dcon::province_id pid, dcon::commodity_id c, float mobilization_impact) {
```
schombert, line 2006: This function doesn't appear to be used. What is it trying to accomplish? it's name suggests that it would be returning a list of inputs to the rgo production method, but since those don't have inputs ... I am thus unclear on what the function does.

src/gamestate/dcon_generated.txt
```diff
@@ -560,6 +560,11 @@ object {
 		type{ float }
 		tag{ scenario }
 	}
+	property{
+		name{ can_be_built_in_colonies }
```
schombert, line 564: So I am guessing you want to support some factory types appearing in colonies? But none of the build checks take factory type into consideration.

src/gui/gui_element_types.hpp
```diff
@@ -29,7 +29,7 @@ void render_text_chunk(
 );
 
 template<typename T, typename ...Params>
-std::unique_ptr<element_base> make_element_by_type(sys::state& state, std::string_view name, Params&&... params) { // also bypasses global creation hooks
+std::unique_ptr<T> make_element_by_type(sys::state& state, std::string_view name, Params&&... params) { // also bypasses global creation hooks
```
schombert, line 32: Why does the type need to be preserved here?

src/gui/gui_province_tiles_tiles.hpp
```diff
@@ -0,0 +1,278 @@
...
+		dcon::unit_type_id utid = state.world.regiment_get_type(target.regiment);
+		return utid;
+	}
+
+	// Done
+	void button_action(sys::state& state, economy::province_tile target, ui::element_base* parent) noexcept override {
+		auto army = state.world.regiment_get_army_from_army_membership(target.regiment);
+		state.selected_armies.clear();
+		state.selected_navies.clear();
+		state.select(army);
```
schombert, line 160: It would probably be wise to check that army is non null, alive, owned by player, etc here at this point

src/economy/province_tiles.cpp
```diff
@@ -0,0 +1,123 @@
+#include "province_tiles.hpp"
+#include "container_types.hpp"
+#include "dcon_generated.hpp"
+#include "commands.hpp"
+#include "economy_production.hpp"
+
+namespace economy {
```
schombert, line 7: as this is purely ui logic, it should be moved to the ui namespace and the gui directory

src/economy/economy_production.cpp
SneakBug8: It duplicates the calculation of RGO efficiency inputs consumption for the RGO tooltip.

src/gamestate/dcon_generated.txt
SneakBug8: economy::can_build_factory_in_colony(state, sid, type) check in command::can_begin_factory_building_construction. When type is provided - it checks particular type. Otherwise - it checks colonial construction in principle

src/gamestate/commands.cpp
```diff
@@ -522,11 +523,12 @@ bool can_begin_factory_building_construction(sys::state& state, dcon::nation_id
 
 	if(!state.world.nation_get_active_building(source, type) && !state.world.factory_type_get_is_available_from_start(type))
 		return false;
-	if(state.world.province_get_is_colonial(state.world.state_instance_get_capital(sid)))
```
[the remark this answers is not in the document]
SneakBug8: Moved into assert

src/economy/economy_production.cpp
SneakBug8: Done

src/ai/ai.cpp
SneakBug8: Fixed

src/ai/ai.cpp
SneakBug8: Fixed

src/economy/construction.hpp
SneakBug8: Renamed

SneakBug8:
Just a friendly reminder on the PR.

PR #1968, opened 2025-06-02
Split AI files, utility calculations for CB, allow addition of CB midwar for AI

ineveraskedforthis:
Before introduction of local administrative efficiency taking random states in the middle of nowhere was a fine decision. But currently it's more optimal to avoid exclaves as you can't collect taxes/tariffs there without establishment of local administration.
In this PR I do the following:
- Alleviate the issue of AI capturing exclaves (RIP Prussian Khiva) by adjusting the AI logic of selecting CBs for fabrication and adding mid-war.
- Split ai.cpp file per category
- Allow AIs to liberate countries and add wargoals during usual wars as long as it is beneficial to them or harms their rivals.
- Implement utility system which exposes (currently inside cpp code) several variables which could influence aggressiveness of AI.


src/gamestate/serialization.cpp
```diff
@@ -1047,6 +1047,17 @@ void write_save_file(sys::state& state, save_type type, std::string const& name)
 
 	state.save_list_updated.store(true, std::memory_order::release); // update for ui
 
+	// log count of pressed wargoals
+	{
```
schombert, line 1051: I imagine that you didn't mean to leave this in (or you forgot to hide it behind a condition.

src/ai/ai_war.cpp
```diff
@@ -0,0 +1,2139 @@
...
+	if(military::cb_requires_selection_of_a_liberatable_tag(state, t))
+		return false;
+	if(military::cb_requires_selection_of_a_valid_nation(state, t))
+		return false;
+	return true;
+}
+
+void prepare_and_sort_list_of_desired_states(
+	sys::state& state,
+	std::vector<ai::weighted_state_instance>& result,
```
schombert, line 481: assuming that you don't intend to call this function repeatedly with the same vector in order to append results to a single vector, this function (and the one below) could be written to return the std::vector result without any loss of efficiency because of what is known as the return value optimization, aka copy elision. Basically, if a function is always returning a given object (i.e. you don't construct two different vectors in the function and then conditionally return one or the other), then the compiler is allowed to implement a function returning an object *as if* it took a pointer to uninitialized memory for that object, in which it constructs the object that it will ultimately return, meaning that no copying has to be done, and so it can be just as efficient as what you have written here, but with a clearer expression of what is the input and the outputs. You can read the notes here https://en.cppreference.com/w/cpp/language/return.html#Notes if you are curious about tracking down the details of when copy elision can be expected.

src/ai/ai_war.cpp
```diff
@@ -0,0 +1,2139 @@
...
+							diplomatic_message::post(state, m);
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+bool will_accept_crisis_peace_offer(sys::state& state, dcon::nation_id to, bool is_concession, bool missing_wg) {
```
schombert, line 1140: If the logic for these functions has changed, then there may be some ui elements that need to be changed as well (if the tooltip on why the ai will/won't accept gives details)

src/ai/ai_war.cpp
ineveraskedforthis: The function below is used repeatedly, but this one could be made into a function which returns the vector, thanks. Actually these functions were already like this, I just edited them to use utility.

src/ai/ai_war.cpp
ineveraskedforthis: I haven't changed the functions which influence AI agreement, my changes are focused on changing behavior of selecting a wargoal in different contexts

PR #1969, opened 2025-06-04
Fix lockstep multiplayer save loading, reloading and oos recovery

ZombieFreak115:
This PR seeks to do the following:

Fix synchronicity in MP when loading or reloading a save, and ensuring that all clients&host end with the same exact gamestate no matter what lingering data there may be. This is implemented by completly resetting the gamestate first before a reload (using dcon's .reset() method). And then reloading from there.

As a result of the hard-reload, i have added a "ui_lock" mutex which ensures no data is read by the ui rendering while the gamestate is being reset&reloaded, as such cases lead to crashes in my testing.
Fixed/refactored a bunch of possible invalid writes to zero'd out IDs, quite a few likely remain using getters, and is a likely source for many of the oos'es. Further fixing of those will be a seperate PR.

Resyncing is now an active action the host must take when one or more players oos, rather than happening automatically as soon as an oos is detected, which can be activated from the tab menu by the host only.

Refactored many methods used in fill_unsaved_data to NOT change saved data in any way, and made sure they are still used the same everywhere else (eg in daily tick), Together with some tests to confirm it.

I also refactored the "save_slock" to be a mutex aswell, in order to avoid a race condition.

Issues/deficiencies:

Oos'es are still somewhat frequent, and i intend to try to solve them as i find the source (likely many are from invalid writes), but since the oos-recovery mechanicsm ensures an equal "start" state now, it is less of an issue.

Currently, anytime a player joins it will force the entire lobby to reload their own save. This is technically redundant, but is needed currently as a consequence of how unsaved data is handled, as without a full reset, lingering unsaved data could remain unchanged and alter the state on the next tick. A better solution to this might be found eventually, but currently it is my attempt to "futureproof" it against further changes to unsaved data which could break it again.

The "ui_lock" and "save_slock" are now mutexes, and it is prehaps possible for a very clever person to archieve a similar locking-style mechanism with just atomics with better performance, however i was unable to.

Lastly, i had one client crash when trying to join when i didnt have the debugger on, and further attempts to replicate said crash was unsucessful.


schombert:
Is the `keep_after_state_reload` tag in dcon necessary, or is it just a leftover from debugging?

src/gamestate/system_state.cpp
```diff
@@ -3782,7 +3849,7 @@ void state::fill_unsaved_data() { // reconstructs derived values that are not di
 
 	pop_demographics::regenerate_is_primary_or_accepted(*this);
 
-	nations::update_administrative_efficiency(*this);
+	nations::update_national_administrative_efficiency(*this);
```
schombert, line 3852: below in the update you call update_administrative_efficiency in addition to this. Do both need to be called here as well?

schombert:
It is generally safer to use something like https://en.cppreference.com/w/cpp/thread/scoped_lock.html instead of manually calling lock and unlock, since it prevents an early return from a function, exception, or other unforeseen control flow from skipping the unlock and hence deadlocking later

ZombieFreak115:
> Is the `keep_after_state_reload` tag in dcon necessary, or is it just a leftover from debugging?

Yes it is nessecary, it a tag put on "local only" dcon fields, so these are specifically NOT reset when the save has to be reloaded, whereas everything else is. The get_mp_state_checksum() method also does not take these fields into account when calculating the checksum for oos checks.

src/gamestate/system_state.cpp
ZombieFreak115: update_administrative_efficiency() does not need to be called here. The reason is is because both of these methods used to be just one method (update_administrative_efficiency), however the part that updates national admin did not modify any dcon "save" fields, whereas the rest of the method did. Modifiying saved fields inside of fill_unsaved_data can be problematic for MP syncing for a number of reasons, and it is also uneeded (since the saved data will be transferred with the save beforehand).

Therefore refactored the national admin update part into its own method, and the rest which changes saved data (update_administrative_efficiency) is used everywhere it was previously except in fill_unsaved_data (as it is not unsaved). The national admin update is then used everywhere it was previously, as it only updates unsaved data.

PR #1973, opened 2025-06-06
Adjust urban sprawl to work better with different map sizes

ineveraskedforthis:


src/map/map.hpp
```diff
@@ -195,6 +195,7 @@ class display_data {
 	std::vector<uint8_t> terrain_id_map;
 	std::vector<uint8_t> median_terrain_type;
 	std::vector<uint32_t> province_area;
+	std::vector<float> province_area_km2;
```
schombert, line 198: it might be worthwhile to think about moving some of these values into dcon if they have to be serialized anyways. I believe that when Erik originally wrote this section he has unfamiliar with dcon and so avoided putting anything into it for that reason, not out of some principled choice.

PR #1974, opened 2025-06-08
Less micromanagement in colonies and stockpile_spoilage define

SneakBug8:


SneakBug8:
@schombert  Sorted out merge conflict here. Should be good to merge

src/provinces/province.cpp
```diff
@@ -1998,6 +1998,16 @@ void update_colonization(sys::state& state) {
 
 			d.set_colonization_stage(uint8_t(3));
 			(*colonizers.begin()).set_last_investment(state.current_date);
+		} else if(num_colonizers == 1 && (*colonizers.begin()).get_colonizer().get_is_player_controlled() &&
+			command::can_finish_colonization(state, (*colonizers.begin()).get_colonizer(), d)) {
+			command::execute_finish_colonization(state, (*colonizers.begin()).get_colonizer(), d);
+			/* Since situations where you initiate a colony and don't want to colonize it completely are so rare,
+			* it makes sense to automatically invest points for players to reduce micromanagement. */
+		} else if(num_colonizers == 1 && (*colonizers.begin()).get_colonizer().get_is_player_controlled() &&
```
schombert, line 2006: Except if there were two players in this game with this logic they would be forced to tie up all their points in contesting a single colony, possibly even to the point of a colonial crisis, whether they wanted to or not. Plus, slowing down AI colonization by investing just once to force them to go for all 4 levels is a viable way to slow down their colonization efforts while you out race them elsewhere. @SneakBug8

PR #1976, opened 2025-06-08
AI embargoes

SneakBug8:


src/gamestate/system_state.cpp
```diff
@@ -4294,7 +4294,6 @@ void state::single_game_tick() {
 			ai::make_attacks(*this);
 			ai::update_ships(*this);
 		}
-		ai::take_ai_decisions(*this);
```
schombert, line 4297: @SneakBug8 won't removing this prevent the AI from taking decisions?

SneakBug8: Brought the call back >_>

PR #1978, opened 2025-06-08
Refactoring of naval battles, update to naval combat formula and other minor additions

ZombieFreak115:
This refactors naval battles and the combat formula to work more like what would expect from vic2, such as taking unit strength & evasion into account, and to not always retreat when below 15% org or str. It also adds some additional naval stacking penalties to atleast try to make spamming low-quality ships less of a good strategy.

Also adds some defines for build time multipliers, instead of having it always be hardcoded at 0.1f. I set the land unit & naval unit build time mult define to 1.0f to avoid being able to instantly rebuild lost ships/regiments by default.


src/military/military.cpp
```diff
@@ -6183,13 +6192,14 @@ float unit_get_effective_default_org(sys::state& state, dcon::regiment_id reg) {
...
 	auto navy = state.world.ship_get_navy_from_navy_membership(ship);
+	auto owner = state.world.navy_get_controller_from_navy_control(navy);
 	auto type = state.world.ship_get_type(ship);
 	auto base_org = state.world.nation_get_unit_stats(state.world.navy_get_controller_from_navy_control(navy), type).default_organisation;
 	auto leader = state.world.navy_get_admiral_from_navy_leadership(navy);
 	auto leader_bg = get_leader_background_wrapper(state, leader);
 	auto leader_per = get_leader_personality_wrapper(state, leader);
 	return base_org * ((1.0f + state.world.leader_trait_get_organisation(leader_bg) + state.world.leader_trait_get_organisation(leader_per)) *
-		   (1.0f + state.world.leader_get_prestige(leader) * state.defines.leader_prestige_to_max_org_factor));
+		   (1.0f + state.world.leader_get_prestige(leader) * state.defines.leader_prestige_to_max_org_factor) * (1.0f + state.world.nation_get_modifier_values(owner, sys::national_mod_offsets::land_organisation)));
```
schombert, line 6202: should we be using land org for a ship?

ZombieFreak115: Nope, missed that, it should indeed be naval_organisation. will fix it

PR #1980, opened 2025-06-09
Refactor many getters being used as setters into proper setters

ZombieFreak115:
This PR refactors nearly all getters being used as setters (with += or similar assignments to the returned references) into setters in the given files for ease of debugging invalid writes. From my testing it compiles into the same assembly in release, so there should be no performance penalty.
I'll slowly do them a couple of files at a time in seperate PRs to avoid merge conflicts, and make it less overwhelming to review.

### DIFF
--- a/src/provinces/province.cpp
+++ b/src/provinces/province.cpp
@@ -479,20 +479,20 @@ void set_province_controller(sys::state& state, dcon::province_id p, dcon::natio
 		auto rc = state.world.province_get_rebel_faction_from_province_rebel_control(p);
 		auto owner = state.world.province_get_nation_from_province_ownership(p);
 		if(rc && owner) {
-			state.world.nation_set_rebel_controlled_count(owner, state.world.nation_get_rebel_controlled_count(owner) - uint16_t(1));
+			state.world.nation_set_rebel_controlled_count(owner, uint16_t(state.world.nation_get_rebel_controlled_count(owner) - uint16_t(1)));
 			if(!is_overseas(state, p)) {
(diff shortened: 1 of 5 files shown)


src/provinces/province.cpp
```diff
@@ -642,7 +642,7 @@ void update_blockaded_cache(sys::state& state) {
 		if(owner && owner == controller) {
 			if(!is_overseas(state, pid)) {
 				if(military::province_is_blockaded(state, pid)) {
-					state.world.nation_get_central_blockaded(owner) += uint16_t(1);
+					state.world.nation_set_central_blockaded(owner, state.world.nation_get_central_blockaded(owner) + uint16_t(1));
```
schombert, line 645: So unfortunately this (I have picked this one at random, but the same applies in many places) is going to be a problem for me later. The issue is the rules regarding https://en.cppreference.com/w/c/language/conversion.html#Integer_promotions . What will happen here is that both the left and right side of the addition will be promoted to unsigned integers, making the result of the addition an unsigned integer. This is wider than the type expected by the call to set, so the compiler implicitly performs a truncation here. This is all fine, functionally. The issue is that I compile the release in Clang with the warnings turned way up, to the point where it will not allow implicit truncating conversions. Thus, when I compile the release, I would need to add an explicit cast for the result of the addition back into uint16_t *everywhere*. I will leave it up to you how to solve this problem, whether that is by adding casts or adjusting build flags: if you assure me that it will build under the Release, Clang (profile) build (which has the same flags as the optimize option which is what actually makes the final release) then I will be satisfied.

schombert: Here is another, more readable, explanation of promotion rules https://wiki.sei.cmu.edu/confluence/display/c/INT02-C.+Understand+integer+conversion+rules

PR #1984, opened 2025-06-10
Regiment & ship construction time application

SneakBug8:


schombert:
you have added a date to naval constructions, but it doesn't seem to be used

SneakBug8:
Included ship building as well

PR #1995, opened 2025-06-22
MP player network refactor, support co-op, and some minor military changes

ZombieFreak115:
This PR refactors much of the MP-related code to make the "mp_player_id" the main entity representing a player in MP, instead of the nation_id being the identifier. The nickname is used for identification when sent from client to server, or vice versa.
In the process refactors many things such as ingame messages, nation picking, and joining/leaving. 

As a result, coop is now supported. Events get sent to both players on the same nation, and its first-come-first-server on who clicks the event option first. I also added some server-level checks in commands to ensure no event is executed twice. I also added a server-side check on player_pick_nation to avoid a bad actor abusing coop to silently get on someone else's nation.

Also refactored code around the "swap player" effects to be able to work with the possibility of having multiple players on the same nation.

Added a non-console way of playing as the spectator nation (rebels) with a button in the nation picker screen. It can be used for both SP and MP, and if there are no nations left in MP new players will be put on the spectator tag. Currently FOW is still on for the spectator tag in MP. This could become a host setting later, but for now im leaving it as-is.

Made it so regiments that are in a battle or retreating when the source province changes owner will get safely deleted (from any battles aswell). Previously its strength was set to 0, but could still reinforce and still "stole" the brigade slot&pops from the new owner untill deleted manually.

Added a blackflag update when a nation joins a war, so that army collitions are immedately detected, instead of the two armies being on the same tile with no battle. Also set the organiation of a regiment that has changed owner to be the previous org instead of 0, as in my testing it was leading to some pretty loopsided fights in secessions.

### DIFF
--- a/src/gamestate/commands.cpp
+++ b/src/gamestate/commands.cpp
@@ -43,12 +43,13 @@ void add_to_command_queue(sys::state& state, payload& p) {
 	case command_type::notify_player_fully_loaded:
 	case command_type::notify_player_is_loading:
 	case command_type::chat_message:
+	case command_type::change_ai_nation_state:
 		// Notifications can be sent because it's an-always do thing
 		break;
 	case command_type::notify_start_game:
 	case command_type::notify_stop_game:
 		// do not allow starting or stopping the game whilst clients are loading!
-		if(state.network_state.num_client_loading == 0) {
+		if(!network::check_any_players_loading(state)) {
 			break;
 		}
 		else {
@@ -56,7 +57,7 @@ void add_to_command_queue(sys::state& state, payload& p) {
 		}
 	default:
 		// Normal commands are discarded iff we are not in the game, or if any other client is loading
-		if(!state.current_scene.game_in_progress || state.network_state.num_client_loading != 0)
+		if(!state.current_scene.game_in_progress || network::check_any_players_loading(state))
 			return;
 		state.network_state.is_new_game = false;
 		break;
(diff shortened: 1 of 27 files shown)


src/gamestate/commands.cpp
```diff
@@ -1016,44 +1010,37 @@ void execute_make_vassal(sys::state& state, dcon::nation_id source, dcon::nation
...
 	nations::liberate_nation_from(state, t, source);
 	auto holder = state.world.national_identity_get_nation_from_identity_holder(t);
 	nations::remove_cores_from_owned(state, holder, state.world.nation_get_identity_from_identity_holder(source));
 
-	if(state.world.nation_get_is_player_controlled(source)) {
-		network::switch_player(state, holder, source);
-	} else if(state.world.nation_get_is_player_controlled(holder)) {
-		network::switch_player(state, source, holder);
+	if(state.network_mode == sys::network_mode_type::single_player) {
+		network::switch_all_players(state, holder, source);
```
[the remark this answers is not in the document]
schombert: Also, would it make sense to now move the remove ai data function inside it as well?

src/gamestate/commands.cpp
```diff
@@ -5352,27 +5399,29 @@ void execute_notify_player_joins(sys::state& state, dcon::nation_id source, sys:
...
+	//	}
+	//	// update mp player with the joining players loading state
+	//	state.world.mp_player_set_fully_loaded(p, !needs_loading);
+	//	state.world.mp_player_set_is_oos(p, false);
+	//}
+	//else {
+	//	p = network::create_mp_player(state, name, password, !needs_loading, false);
+	//}
+ //	state.world.nation_set_is_player_controlled(source, true);
+	//state.world.force_create_player_nation(source, p);
```
schombert, line 5424: I know that many of us, myself included, have the habit of leaving dead code around as comments, but this, and the following case, are probably safe to remove. They will always be in the git history if you need to find them again.

src/provinces/province.cpp
```diff
@@ -1265,13 +1265,20 @@ void change_province_owner(sys::state& state, dcon::province_id id, dcon::nation
 				&& !src.get_regiment().get_army_from_army_membership().get_battle_from_army_battle_participation()
 				&& !src.get_regiment().get_army_from_army_membership().get_controller_from_army_rebel_control()) {
 					auto loc = src.get_regiment().get_army_from_army_membership().get_location_from_army_location();
+					auto old_army = src.get_regiment().get_army_from_army_membership();
 					auto new_u = fatten(state.world, state.world.create_army());
 					new_u.set_controller_from_army_control(new_owner);
 					src.get_regiment().set_army_from_army_membership(new_u);
-					src.get_regiment().set_org(0.01f);
+					// if the previous army is now empty, clean it up early so incoming collitions with enemy armies on the same day can be handled properly
```
schombert, line 1272: "collisions"

src/scripting/events.cpp
```diff
@@ -13,10 +13,7 @@ bool is_valid_option(sys::event_option const& opt) {
 
 void take_option(sys::state& state, pending_human_n_event const& e, uint8_t opt) {
 	for(auto i = state.pending_n_event.size(); i-- > 0;) {
-		if(state.pending_n_event[i].date == e.date && state.pending_n_event[i].e == e.e &&
-				state.pending_n_event[i].from_slot == e.from_slot && state.pending_n_event[i].n == e.n &&
-				state.pending_n_event[i].primary_slot == e.primary_slot && state.pending_n_event[i].r_hi == e.r_hi &&
-				state.pending_n_event[i].r_lo == e.r_lo) {
+		if(state.pending_n_event[i] == e) {
```
schombert, line 16: This is not exactly the same as the previous comparisons. The new check also compares things such as slot_type and the contents of the padding bytes that were previously ignored. Double check that this doesn't matter.

schombert: never mind; I see that you have added operator == instead

src/provinces/province.cpp
ZombieFreak115: Fixed

src/scripting/events.cpp
ZombieFreak115: Added custom definitions for "!="  aswell

PR #1998, opened 2025-06-23
Expand tiles UI

SneakBug8:
Admin tiles,

market tiles

Expand RGO tooltips

Expand province building tooltips

### DIFF
--- a/src/economy/economy_government.cpp
+++ b/src/economy/economy_government.cpp
@@ -291,7 +291,7 @@ std::vector<employment_record> explain_capital_administration_employment(sys::st
 	auto capital = state.world.nation_get_capital(n);
 	auto capital_state = state.world.province_get_state_membership(capital);
 	auto capital_of_capital_state = state.world.state_instance_get_capital(capital_state);
-	auto record = employment_record{ economy::labor::high_education_and_accepted };
+	auto record = employment_record{ economy::labor::high_education_and_accepted, 0.f, 0.f, 0.f };
 
 	record.target_employment = state.world.nation_get_administration_employment_target_in_capital(n);
 	record.satisfaction = state.world.province_get_labor_demand_satisfaction(capital, economy::labor::high_education_and_accepted);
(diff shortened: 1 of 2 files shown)


SneakBug8:
@ineveraskedforthis pls review the decomposition of economic functions

src/economy/economy_trade_routes.cpp
```diff
@@ -76,6 +152,57 @@ tariff_data explain_trade_route(sys::state& state, dcon::trade_route_id trade_ro
...
+
+			auto B = state.world.trade_route_get_connected_markets(route, 1);
+			if(B == market) {
+				auto multiplier = -1.f;
+				imports_volume += prediction.current_volume;
+				imports_value += multiplier * prediction.current_volume * prediction.import_price[1];
+				trade_value += multiplier * prediction.current_volume * prediction.import_price[1];
+			}
+			else {
+				exports_volume += prediction.current_volume;
```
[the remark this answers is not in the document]
SneakBug8: Switched to a different function.

src/economy/economy_government.cpp
```diff
@@ -268,4 +268,30 @@ tax_information explain_tax_income(sys::state& state, dcon::nation_id n) {
 	return result;
 }
 
+// Calculate employment of local administrations in the province.
+// When using, check for capital administration separately
+float explain_administration_employment(sys::state& state, dcon::province_id p) {
```
schombert, line 273: I don't understand the names of these functions, since they don't explain anything

src/economy/economy_government.cpp
```diff
@@ -268,4 +268,30 @@ tax_information explain_tax_income(sys::state& state, dcon::nation_id n) {
...
+	auto n = state.world.province_get_nation_from_province_ownership(p);
+
+	for(auto admin : state.world.nation_get_nation_administration(n)) {
+		if(admin.get_administration().get_capital() == p) {
```
schombert, line 277: this loop seems unnecessary, as you could just go directly from province -> state instance -> administration without needing to peek in every administration in the nation

schombert: the dcon file still contains
```
	name{ local_administration }
	tag{ save }
	link{
		object{state_instance}
		name{state_instance}
		type{unique}
	}
	link{
		object{administration}
		name{administration}
		type{many}
	}
```
is this no longer used anywhere?

PR #2000, opened 2025-06-25
Fix disabled WG options on adding wargoals

SneakBug8:
- Fix disabled WG options on adding wargoals
- Count 50% of the occupation score in later wars towards earlier wars


schombert:
this pr includes some random temporary file

src/military/military.cpp
```diff
@@ -19,34 +19,39 @@ namespace military {
 
 // this function should be used
 // only for provinces owned by a war participant
-bool does_province_count_for_war_occupation(sys::state& state, dcon::war_id w, dcon::province_id p) {
+// US5AC4. What % of the province score should be counted towards occupation. [0.0f;1.0f].
+float share_province_score_for_war_occupation(sys::state& state, dcon::war_id w, dcon::province_id p) {
```
schombert, line 23: this seems like it would make it possible for wars to get stuck again, with both sides unable to get enough occupation score to actually get what they want

SneakBug8: We agreed this change with peter beforehand as I showcased a test game recording with AI just grabbing full annexation from the player doing no work by itself. I haven't seen any stuck wars so far in my tests.

To prevent wars getting stuck we have ticking warscore, battles, blockades score. AI not peacing out for white peace after a prolonged conflict (where it continuously pays for army upkeep) is a bug in the AI logic itself and should be handled separately.

SneakBug8: Author of the feature has not put a clear UI alert for the "conflicting war goals" when declaring a war. Such behaviour would for sure lead to negative feedback if it's kept AS IS in the release.

SneakBug8:
@schombert removed the temp file

PR #2001, opened 2025-06-26
Popgrowth update & fix, and also fix ui-related oos

ZombieFreak115:
This PR changes the popgrowth formula to more vic2-like, and in the process get more reasonable pop numbers compared to before. It also updates the ui to reflect the new calculation. Changed the display of popgrowth modifiers to be a percentages with 2 decimal places.

While i tried my best to read and understand the ve-library and use it to implement the new formula, and it also works as expected in my tests, it is possible i have made some mistakes as i haven't used it before, so probably check that part to make sure i haven't made any.

In addition, i also fixed a ui-related oos due to military_rank, prestige_rank and industry_rank being part of the mp_checksum, despite being ui properties. In the process i created a new tag specifically for exempting objects/properties from the checksum. I then added a doc file with a short explanation of each dcon serialization tag, since we have quite a few now.


src/economy/demographics.cpp
```diff
@@ -2598,11 +2598,13 @@ void update_growth(sys::state& state, uint32_t offset, uint32_t divisions) {
...
 	define:MIN_LIFE_RATING_FOR_GROWTH, subtract define:MIN_LIFE_RATING_FOR_GROWTH from it, and then multiply by
 	define:LIFE_RATING_GROWTH_BONUS. If it is less than define:MIN_LIFE_RATING_FOR_GROWTH, treat it as zero. Now, take that value
-	and add it to define:BASE_POPGROWTH. This gives us the growth factor for the province.
+	and add it to define:BASE_POPGROWTH and add to all pop growth modifiers. This gives us the growth factor for the province.
 
-	The amount a pop grows is determine by first computing the growth modifier sum: (pop-life-needs -
-	define:LIFE_NEED_STARVATION_LIMIT) x province-pop-growth-factor x 4 + province-growth-modifier + tech-pop-growth-modifier +
-	national-growth-modifier x 0.1. Then divide that by define:SLAVE_GROWTH_DIVISOR if the pop is a slave, and multiply the pop's
+	Then compute the starvation modifier:
+	MIN((pop_life_needs * 2) / define:LIFE_NEED_STARVATION_LIMIT - 1, 1.0) which is multiplied to all *postitive* pop growth modifiers in the pop growth factor.
```
schombert, line 2604: I don't think that this formula is going to work in general if mods have changed the define. Prior to the change, it was the point at which pops switched from shrinking via starvation to growing. So, if it was 0, for example, pops would always grow. It is going to make a weirdly shaped curve now, where if it goes below 0.5 the growth factor is going to be clamped to 1 for much of the range. Similarly, suppose that it was set to 1 (pops should always starve), well, pops meeting all their life needs here would still come up with a positive growth rate.

src/economy/demographics.cpp
```diff
@@ -2616,44 +2618,102 @@ void update_growth(sys::state& state, uint32_t offset, uint32_t divisions) {
...
+		ve::fp_vector positive_modifiers = lr_factor + state.defines.base_popgrowth +
+			ve::select(state.world.province_get_modifier_values(loc, sys::provincial_mod_offsets::population_growth) >= 0.0f, state.world.province_get_modifier_values(loc, sys::provincial_mod_offsets::population_growth), 0.0f) +
+			ve::select(state.world.nation_get_modifier_values(owner, sys::national_mod_offsets::pop_growth) >= 0.0f, state.world.nation_get_modifier_values(owner, sys::national_mod_offsets::pop_growth), 0.0f);
+	
+
+		positive_modifiers = ve::select(state.world.pop_get_poptype(ids) == state.culture_definitions.slaves, positive_modifiers / state.defines.slave_growth_divisor, positive_modifiers);
+
+		ve::fp_vector negative_modifiers = ve::select(state.world.province_get_modifier_values(loc, sys::provincial_mod_offsets::population_growth) < 0.0f, state.world.province_get_modifier_values(loc, sys::provincial_mod_offsets::population_growth), 0.0f) +
+			ve::select(state.world.nation_get_modifier_values(owner, sys::national_mod_offsets::pop_growth) < 0.0f, state.world.nation_get_modifier_values(owner, sys::national_mod_offsets::pop_growth), 0.0f);
```
schombert, line 2629: If you want to separate a value into a positive (or zero) and a negative (or zero) values for some formula, it is better to do the following: `select(values >= 0.0f, positive formula, negative formula)`, which is much cleaner to read.

PR #2004, opened 2025-06-29
Update and expand Chinese localisation for alice.csv

MeizhouWan:
Removed three obsolete factory stats strings and added multiple new localisation entries related to economy, employment, trade centers, administration, and building modifiers. This update enhances the coverage and accuracy of in-game text for new and existing features.

@schombert Is there a tool to check for duplicate key columns for CSVs? :(


schombert:
yes, there are many tools that can do that for you. Here is one of them for example: https://www.csvexplorer.com/blog/remove-duplicates-from-a-csv-file/ @MeizhouWan

PR #2005, opened 2025-06-29
Reorganize labour UI code & ledger columns tooltips

SneakBug8:


src/gui/gui_ledger_window.hpp
```diff
@@ -81,11 +81,16 @@ class ledger_generic_sort_button : public button_element_base {
 		send(state, parent, current_sort);
 	}
 	tooltip_behavior has_tooltip(sys::state& state) noexcept override {
-		return std::holds_alternative<dcon::pop_type_id>(type) ? tooltip_behavior::tooltip : tooltip_behavior::no_tooltip;
+		return tooltip_behavior::variable_tooltip;
```
schombert, line 84: Don't see why this isn't `tooltip_behavior::tooltip` here

src/gui/labour_details.cpp
```diff
@@ -0,0 +1,39 @@
+#include "gui_common_elements.hpp"
+
+namespace ui {
+
+// Return text key for tooltips on the provided labour type
+std::string labour_type_to_employment_type_text_key(int32_t type) {
```
schombert, line 6: these functions should return string views; there is no need to allocate for static strings

src/gui/gui_ledger_window.hpp
SneakBug8: What is the difference between tooltip_behavior::tooltip and tooltip_behavior::variable_tooltip?

PR #2012, opened 2025-07-04
Even more getter refactoring, fix invalid writes, land attrition refactoring & some fixes

ZombieFreak115:
- Refactors economy_government getter-setters to use setters

- Fixes a bunch of invalid writes, mostly from using state instances from uncolonized provinces.

- Changes the behavoir when a pop attached to a brigade dissapears for any reason, now it will immediately remove the brigade and not allow any brigade to exist w/o a pop being attached to it. This is to solve several annoyances arising from brigades existing w/o any pop attached, and also makes abit more sense.

- Refactored the land attrition code, in the process fixing varius bugs (such as attrition being capped at 2% no matter what, and attrition previously could lead to negative brigade strengths). Also changed the formula to make abit more sense (previously enemy forts could reduce the amount of attrition taken), and made it so armies take a tick of possible attrition each time they move to a new province (vanilla behavour). Lastly, added at-sea attrition which is also a vanilla feature, but made it a static amount decided by a define instead of the wierd vanilla behavour (where it would use the supply limit of the sea prov, which isnt ideal).

- Some UI fixes, such as the army& navy arrival time tooltip now being accurate (and fixing an inconsistency in the arrival times in general), made the attrition tooltip accurate, and added a confirmation window for disbanding military units (it is really easy to accidentially do this, as the split unit button is right next to it)

- Lastly, changed the equation of soldier-to-pop damage to a version closer to vanilla, as it ended up with much lower pop damage than it should. (For example, it would always divide the result by 3.0f no matter what). Dividing by the tech modifier would not make much sense, as it would at most divide it by 1.04f with all tech modifiers, hardly a big change.


src/economy/demographics.cpp
```diff
@@ -4305,6 +4305,10 @@ void remove_size_zero_pops(sys::state& state) {
 	for(auto last = state.world.pop_size(); last-- > 0;) {
 		dcon::pop_id m{dcon::pop_id::value_base_t(last)};
 		if(state.world.pop_get_size(m) < 1.0f) {
+			//safely delete any regiment which has this pop as its source
+			for(auto reg : state.world.pop_get_regiment_source(m)) {
+				military::delete_regiment_safe_wrapper(state, reg.get_regiment());
```
schombert, line 4310: You cannot delete units from a list while iterating over it, because deleting reorders the list and undesirable things will happen. This also goes for the change below

schombert: The most idiomatic way to do this is to delete the first element in the list and then repeat until the list is empty

src/gui/gui_unit_panel.hpp
```diff
@@ -126,17 +126,182 @@ class unit_selection_split_in_half_button : public button_element_base {
 	}
 };
 
+
+class disband_units_title : public simple_text_element_base {
```
schombert, line 130: I really, really hate having to click "yes" on popups after I have already clicked to do an action, partly because of my carpal tunnel syndrome. If you want to add extra "did you really mean this?" confirmations, then I would like them to be behind and settings option and off by default.

src/gui/map_tooltip.cpp
```diff
@@ -137,22 +106,38 @@ void country_name_box(sys::state& state, text::columnar_layout& contents, dcon::
 			auto army = dcon::fatten(state.world, a);
 			auto path = command::calculate_army_path(state, state.local_player_nation, a, army.get_location_from_army_location(), prov);
 			auto curprov = army.get_army_location().get_location().id;
+			auto current_path = army.get_path();
 
-			if (path.size() == 0) { /* No available route */ }
+			if (path.empty()) { /* No available route */ }
 
-			else if(army.get_arrival_time() && *(army.get_path().end()) == prov) {
+			else if(army.get_arrival_time() && army.get_path().size() > 0 && *(army.get_path().end() - 1) == prov) {
```
schombert, line 113: you don't need to check size here because it is already covered by the previous if check

src/military/military.hpp
```diff
@@ -463,7 +463,7 @@ bool will_recieve_attrition(sys::state& state, dcon::army_id a);
 float attrition_amount(sys::state& state, dcon::navy_id a);
 float attrition_amount(sys::state& state, dcon::army_id a);
 float relative_attrition_amount(sys::state& state, dcon::navy_id a, dcon::province_id prov);
-float relative_attrition_amount(sys::state& state, dcon::army_id a, dcon::province_id prov);
+float relative_attrition_amount(sys::state& state, dcon::army_id a, dcon::province_id prov, const std::optional<std::vector<dcon::army_id>>& extra_army_weight = std::optional<std::vector<dcon::army_id>>{ } );
```
schombert, line 466: This is going to be way more overhead than taking a possibly-null pointer to a vector, because to call this function you will almost certainly be constructing a new optional which will create a copy of the vector, even if you are passing the option itself by reference.

src/military/military.cpp
```diff
@@ -6012,51 +6029,34 @@ float peacetime_attrition_limit(sys::state& state, dcon::nation_id n, dcon::prov
...
 
 float local_army_weight(sys::state& state, dcon::province_id prov) {
 	float total_army_weight = 0;
 	for(auto ar : state.world.province_get_army_location(prov)) {
-		if(ar.get_army().get_black_flag() == false && ar.get_army().get_is_retreating() == false &&
-				!bool(ar.get_army().get_navy_from_army_transport())) {
+		if(ar.get_army().get_black_flag() == false && ar.get_army().get_is_retreating() == false) {
 			for(auto rg : ar.get_army().get_army_membership()) {
-				total_army_weight += 3.0f * rg.get_regiment().get_strength();
+				total_army_weight += (state.defines.pop_size_per_regiment / 1000.0f) * rg.get_regiment().get_strength();
```
schombert, line 6059: pop size per regiment is always assumed to be 3000 in many place in the code

src/military/military.cpp
```diff
@@ -6080,52 +6079,53 @@ float local_enemy_army_weight_max(sys::state& state, dcon::province_id prov, dco
...
+	First we calculate (total-strength / define:POP_SIZE_PER_REGIMENT) x MAX(national-attrition-modifier + province-attrition-modifier + 1, with a maximum value of 0) -
+	effective-province-supply-limit (rounded down to the nearest integer). We then reduce that value to at most the max-attrition
+	modifier of the province, and finally we add (define:SEIGE_ATTRITION + fort-level * define:ALICE_FORT_SIEGE_ATTRITION_PER_LEVEL) if the army is conducting a siege. Units taking
+	attrition lose max-strength x attrition-value x 0.01 points of strength. This strength loss is treated just like damage
+	taken in combat, meaning that it will reduce the size of the backing pop.
+	*/
 
-	auto ar = fatten(state.world, a);
+	float total_army_weight = local_army_weight(state, prov);
+	if(extra_army_weight.has_value()) {
```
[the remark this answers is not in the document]
schombert: While it is nicer to have fewer functions, you are adding additional complexity to a "workhorse" function that goes inside the daily update for the sake of making one line of ui code easier to write. This is not a great tradeoff.

src/military/military.cpp
ZombieFreak115: This is true, however it would make more sense for these places to use the pop_size_per_regiment define, as it would keep the logic consistent even in mods that change the pop size per regiment.

PR #2020, opened 2025-07-07
Special army orders: chained siege, strategic redeployment, pursuit and engage

SneakBug8:
Relevant people's vote on the feature: https://discord.com/channels/1064284959034060820/1406964179428053092


src/military/military.cpp
```diff
@@ -8296,15 +8302,50 @@ void navy_arrives_in_province(sys::state& state, dcon::navy_id n, dcon::province
...
+
+		// US7AC1 Handle "move to siege" order
+		if (path.size() > 0 && army_owner && a.get_special_order() == military::special_army_order::move_to_siege) {
+			// Army was ordered to chain siege and it has not yet finished siege
+			auto province_controller = state.world.province_get_nation_from_province_control(from);
+
+			// Must be able to siege the province the army is in
+			if(siege_potential(state, army_owner, province_controller) && state.world.province_get_nation_from_province_control(from) != army_owner) {
+				// Delay the army until it finishes siege
+				a.set_arrival_time(sys::date{});
```
ZombieFreak115, line 8321: Id recommend to use the `military::stop_army_movement ` or `military::stop_navy_movement` wrappers to stop movement of an army/navy. It cleans up other movement-related things aswell. 

And also here you probably want to check command::can_stop_army_movement, as i can't see any checks beforehand which prevents it from stopping movement in a invalid state (ie while retreating)

src/military/military.cpp
```diff
@@ -8366,9 +8407,36 @@ void update_movement(sys::state& state) {
 				}
 			}
 
+			
+			if(!a.get_battle_from_army_battle_participation() && a.get_special_order() == military::special_army_order::pursue_to_engage && a.get_pursuit_target()) {
+				auto reg = a.get_pursuit_target();
+				dcon::army_id target_army;
+				// Find the current army of the target regiment
+				for(auto am : state.world.in_army_membership) {
```
ZombieFreak115, line 8415: Here you can use `regiment_get_army_from_army_membership` to retrieve the army it is part of from the relationship, instead of having to iterate over them all

src/military/military.cpp
```diff
@@ -8366,9 +8407,36 @@ void update_movement(sys::state& state) {
...
+					if(am.get_regiment() == reg) {
+						target_army = am.get_army();
+					}
+				}
+				// Update the path
+				auto npath = command::can_move_army(state, army_owner, a, state.world.army_get_location_from_army_location(target_army), true);
+				auto new_next_dest = npath.at(npath.size() - 1);
```
ZombieFreak115, line 8422: Wont this cause an out of bounds error if the returned path is empty? Since initial size would be 0, and the index would be -1

src/gamestate/commands.cpp
```diff
@@ -3940,19 +3941,63 @@ void execute_move_army(sys::state& state, dcon::nation_id source, dcon::army_id
 		}
 	}
 
+	auto existing_path = state.world.army_get_path(a);
+
+	// Invalid destination province: reset existing path
```
ZombieFreak115, line 3946: Im pretty sure this is old code from the earlier merge which was removed as the "stop movement" was refactored to be its own command. This should be able to be safely deleted.

src/gamestate/commands.cpp
```diff
@@ -3940,19 +3941,63 @@ void execute_move_army(sys::state& state, dcon::nation_id source, dcon::army_id
...
+	auto old_first_prov = existing_path.size() > 0 ? existing_path.at(existing_path.size() - 1) : dcon::province_id{};
+	if(reset) {
+		existing_path.clear();
+		//if the movement is to be reset, and the destination province is the same province as the unit is currently in
+		//we assume that the intention is to cancel the current move command. Cancel the existing path and return w/o assigning a new one
+		if(state.world.army_get_location_from_army_location(a) == dest) {
+			state.world.army_set_arrival_time(a, sys::date{});
+			return;
+		}
+	}
```
ZombieFreak115, line 3962: Same case with this, old code from previous merge, should also be able to be safely deleted

src/military/military.cpp
```diff
@@ -4772,7 +4772,12 @@ float effective_army_speed(sys::state& state, dcon::army_id a) {
 	auto bg = get_leader_background_wrapper(state, leader);
 	auto per = get_leader_personality_wrapper(state, leader);
 	auto leader_move = state.world.leader_trait_get_speed(bg) + state.world.leader_trait_get_speed(per);
-	return min_speed * (state.world.army_get_is_retreating(a) ? 2.0f : 1.0f) *  (leader_move + 1.0f) * state.defines.land_speed_modifier;
+	// US8AC1 Army can be commanded to "Strategic Redeployment" increasing its speed but limiting org to 10%
+	auto special_order_mod = state.world.army_get_special_order(a) == military::special_army_order::strategic_redeployment ? 3.f : 1.f;
+	return min_speed * (state.world.army_get_is_retreating(a) ? 2.0f : 1.0f) *
+		(1.0f + state.world.province_get_building_level(state.world.army_get_location_from_army_location(a), uint8_t(economy::province_building_type::railroad)) *
+								state.economy_definitions.building_definitions[int32_t(economy::province_building_type::railroad)].infrastructure) *
+		(leader_move + 1.0f) * special_order_mod * state.defines.land_speed_modifier;
```
ZombieFreak115, line 4780: Railroads' movement_cost modifier is already taken into account when calculating unit movement (it is applied to the distance, before unit speed is computed) so this looks uneeded. 

Also, is 3x flat movement speed not a bit excessive?

docs/features/out-of-battle.md
```diff
@@ -0,0 +1,37 @@
...
+# US12 Regiment supplies
+
+# US13 Regiments organization regain
+
+**As a Nation,**
+**I want to replenish the organization of my regiments,**
+**So that they can fight in the next battles.**
+
+**Acceptance Criteria:**
+| AC1 | Organization of regiments is replenished monthly |
```
ZombieFreak115, line 12: Org is actually replenished daily.

src/military/military.cpp
SneakBug8: Addressed

src/gamestate/commands.cpp
SneakBug8: Addressed

docs/features/out-of-battle.md
SneakBug8: Addressed

src/military/military.cpp
SneakBug8: Addressed

SneakBug8:
@schombert could please you take a look

src/gamestate/commands.cpp
```diff
@@ -3940,19 +3941,49 @@ void execute_move_army(sys::state& state, dcon::nation_id source, dcon::army_id
...
+			// Target the first regiment of the first army in the target province
+			// Targeting regiments allows more stable pursuit that can't be easily reset by nullifying army id
+			auto al = (*state.world.province_get_army_location(dest).begin());
+
+			if(al) {
+				auto army = al.get_army();
+				auto membership = (*army.get_army_membership().begin());
+
+				if(membership) {
+					state.world.army_set_pursuit_target(a, membership.get_regiment());
```
schombert, line 3971: This is extremely fragile. Regiments can be deleted and new ones created with the same id. This pursuit order could easily result in the army chasing an entirely different army.

schombert:
to work properly, a pursuit order would have to be an army-to-army relationship managed by dcon so that it could be properly updated when armies are deleted

SneakBug8:
@schombert updated the thing

src/gamestate/dcon_generated.txt
```diff
@@ -2216,6 +2221,23 @@ object {
...
+
+	link{
+		object{army}
+		name{source}
+		type{unique}
+	}
+	link{
+		object{army}
+		name{target}
+		type{unique}
```
schombert, line 2236: if they are both unique then it will be impossible for two armies to pursue the same target at the same time.

PR #2035, opened 2025-07-17
optimize launcher header file

stevezxc:


src/launcher/launcher_main_win.cpp
```diff
@@ -1,878 +1,23 @@
-#define ALICE_NO_ENTRY_POINT 1
-
-#include "common_types.cpp"
-#include "system_state.cpp"
-#ifndef INCREMENTAL
```
schombert, line 5: if you have removed all the files it needs to build the non incremental configuration on windows, does it still work?

PR #2051, opened 2025-08-03
Refactor ai code, some unit movement code, other refactors and minor fixes

ZombieFreak115:
Refactors a bunch of code in ai.cpp to use common helper-functions which are also used by other parts og the game.

Also refactors many other common behavoir in many files to use common helper functions.

Splits "unit movement" commands into two new commands: The current "move_army/navy", and "stop_army/navy", and created wrappers to do either one (move_or_stop_army/navy). The reason for changing this is that the old method of passing a "fake" empty vector to signal that a movement was valid when it was to stop movement (using "can_move_army/navy") was very awkward, and didn't work well for expanding rules for unit movement/retreating.

Fixes some egde cases in whích retreating units could bypass blocked straits or hostile canals. Now they are forcefully stopped instead, and enemy unit collision is checked.

Overall, this PR shouldn't affect actual ingame behavior other than the fixes.


src/nations/nations.cpp
```diff
@@ -18,6 +18,7 @@
 #include "economy_government.hpp"
 #include "economy_production.hpp"
 #include "economy_stats.hpp"
+#include <gui_effect_tooltips.hpp>
```
schombert, line 21: the general c++ convention is that angle brackets are used for standard library headers and quotation marks for includes that are part of the project

src/military/military.cpp
```diff
@@ -9545,23 +9533,141 @@ void move_land_to_merge(sys::state& state, dcon::nation_id by, dcon::army_id a,
...
-			return;
+		if(move_army_fast(state, a, path, by)) {
+			state.world.army_set_moving_to_merge(a, true);
+		}
+	}
+}
+
+
+// moves navies for ai nations. skips most player-movement checks and assumes the move command is legitimate. Will return false if there is no valid path.
+bool move_navy_fast(sys::state& state, dcon::navy_id navy, const std::vector<dcon::province_id>& naval_path, bool reset) {
```
schombert, line 9544: It probably doesn't matter here, but it would be more flexible in general to take the path as a std::span (by value), which would allow you to take the path from a number of different sources instead of requiring that the path come from a vector

src/ai/ai.cpp
```diff
@@ -787,18 +741,7 @@ bool merge_fleet(sys::state& state, dcon::navy_id n, dcon::province_id p, dcon::
...
-		auto reg = (*regs.begin()).get_ship();
-		reg.set_navy_from_navy_membership(merge_target);
-	}
-
-	auto transported = state.world.navy_get_army_transport(n);
-	while(transported.begin() != transported.end()) {
-		auto arm = (*transported.begin()).get_army();
-		arm.set_navy_from_army_transport(merge_target);
-	}
+	command::execute_merge_navies<command::execute_cmd_as::ai>(state, owner, merge_target, n);
```
schombert, line 744: This is going to cause a problem. You will notice that the full merge navies command does a number of things that the stripped down version here doesn't do, and one of those is run the naval cleanup function. And merge_fleet doesn't do that because it is called inside a loop that is iterating over the navies, and thus should not be altering the list of which navies exist. The previous version of merge fleet instead relied on the military "garbage collector" to dispose of the empty fleet later outside of the loop. The command version could not rely on this because it could be invoked by the player while the game was paused, and hence outside of the daily update where the "garbage collector" would run, and I didn't want to leave an empty fleet sitting there until the game was unpaused.

src/nations/nations.cpp
ZombieFreak115: I see, fixed.

src/military/military.hpp
```diff
@@ -287,6 +287,13 @@ enum class regiment_dmg_source {
 	combat, attrition
 };
 
+struct ai_path_length {
+	uint32_t length;
```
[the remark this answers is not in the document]
ZombieFreak115: My bad, simply forgot to initialize it, fixed.

src/military/military.cpp
ZombieFreak115: Good idea, i haven't seen any place where a path isn't in a vector, but it could be useful later so added it.

PR #2062, opened 2025-08-24
diplo fixes and additions

ZombieFreak115:
This PR aims to make some small fixes and changes to diplo behavior which is not consistent with the intended functions. If something is intentionally diffrent from the expected behavoir of Vic2 and i didn't see that it was purposeful, let me know.

- Subjects are forced to accept demands of military access and call to arms. Call to arms is mostly a safeguard incase the subject is not in the overlord's wars (puppeted mid-war or similar). Military access is to ensure the overlord can always have troops in the subject nation, without getting undue supply limit penalties from having no access.

- Cannot be called into a war if you are truced with someone on the other side of the war. A automatic defensive call overrides this restriction, allowing you to join even if truced with someone on the other side. This is consistent with Vic2 behavior.

- Implemented "command units" diplo option which was present in Vic2. This option allows you to command an AI subjects units as long as you are both at war. Control is relinquished when the subject is no longer in any wars, or if control is given back manually. This is also kindof a mod support thing, as some mods rely on this command units option to make "puppet-unions" playable.

- Post-battle reports will appear for anyone involved in the battle, not just the lead attacker/defender.

- Add diplo request loc to free trade agreements, and change their logic to work how they are described to work: Free trade agreements can only be signed if both parties are not in a sphere and not a subject. The free trade agreement will then apply to both spheres of the signatories, meaning anyone in one sphere can trade tariff-free with anyone in the other sphere.

- Similar refactoring for embargos: They are sphere-wide, although a non-sphereling can embargo a sphereling in a diffrent sphere.

- Added checks to prevent spherelings from declaring wars against spherelord to be consistent with expected behavoir. Joining wars also have some sphere restrictions, but not as strict to be compliant with vic2 behavoir.

- Added check to ensure nations cannot declare war on eachother if one has units inside the target nation or their overlord.


schombert:
So there are some issues here

> Subjects are forced to accept demands of military access and call to arms.
In PA subjects are automatically brought into wars, and at least when I wrote the code, an overlord can always move through subject territories. Thus there was no need to take those diplomatic actions at all.

> Cannot be called into a war if you are truced with someone on the other side of the war
I see some potential issues here with the order people are called into wars and player reaction time. If A and B have a truce, C declares war on D, A is allied with C and B is allied with D, then which of A and B end up in the war may depends on who is faster on the draw, because if A is faster to respond to the cta then they can join even if B responds defensively later, but they can't if B is fast enough. I believe that the intended "solution" is for A not to be able to add wargoals against B without suffering the "breaking a truce" infamy penalty. If that isn't in the game, adding it is probably a better remedy.

> Implemented "command units" diplo option 
This was left out because it makes it even more unlikely for a strong subject to ever be able to gain independence, since you can just command their troops to die of attrition or be stationed in an island in the middle of nowhere.

> free trade agreements can only be signed if both parties are not in a sphere and not a subject.
You will have to double check that these agreements are appropriately broken then when the conditions change

> Added checks to prevent spherelings from declaring wars against spherelord to be consistent with expected behavoir
I wanted sphere members to be able to declare war on their sphere leader if they wanted to. If you want to make that impossible then I believe that this should be turned into an optional and/or moddable game rule.

schombert:
> And the reason for the military access is because armies get a bonus to supply limit for their units if they have military access in the country, rather than just being in a common war. This probably could just be coded to also trigger if said nation is a subject of X.

I think that the supply limit bonus should probably automatically be given if you are in a common war (or if the nation is a subject). This removes the busy work of having to ask for access that you really ought to be granted. I am not opposed to the idea that subjects have to say "yes" to things, but I think that if they have to say "yes" then you shouldn't even have to ask and should get those benefits automatically.

RE command units:: if you think it is important for MP, you could have it as just an MP feature. If you want to bring it into SP, I just want there to be some guardrail against the obvious abuses.

RE agreements and embargos etc: whenever we add conditions that make something possible only under conditions X it is almost certainly a good idea to make sure that it ends when those conditions no longer hold. Mostly this is just for intuitive consistency; it is weird to say that a sphere member can't do something unless it is grandfathered in.

RE declaring war on your sphere leader: I didn't mean this as a second way to leave the sphere (also isn't that decision only available to secondary powers?), but rather to cover all the other possible cases. Having a force shield against war declarations feels weird to me, especially when other agreements, like truces, don't provide that shield

ZombieFreak115:
Adressed the points in the following manner:

Supply limit bonus from military access is automatically applied when nations are allied in war, or subject/sphereling.

Made command units MP-Only. I can't think of a way to have reasonable guardrails whilst still retaining player control, as even with disbanding puppet units being disallowed one can still intentionally attrition them.

Relavent embagoes & free trade agreements are now removed when a nation becomes sphered.

I implemented a define to have the "vanilla" sphere war block, with it being turned off by default.

PR #2073, opened 2025-09-06
Randomize AI colonization

SneakBug8:


src/ai/ai.cpp
```diff
@@ -222,35 +222,50 @@ void update_ai_colony_starting(sys::state& state) {
...
-					if(p.get_province().get_is_coast() && !coastal_target) {
-						coastal_target = p.get_province();
-					}
-					if(p.get_province().id.index() < state.province_definitions.first_sea_province.index())
-						has_unowned_land = true;
+		states.push_back(sd);
+	}
+	// Shuffle the vector
+	for(int i = (int) states.size() - 1; i >= 0; i--) {
+		int j = rng::get_random(state, state.current_date.to_raw_value()) % (i+1);
```
schombert, line 232: This `rng::get_random(state, state.current_date.to_raw_value())`, of course, will produce the same number every time you call it (you could call it just once outside of the loop) and then you are taking its remainder via successively higher divisors. If the i values were all prime, maybe that wouldn't matter, but they aren't, and so there are strong correlations between what happens at different values. For example, if the remainder mod 8  (i == 7) is, say, 5, then we know that the remainder mod 4 (i == 3) will be 1 and the remainder mod 2 (i == 1) will also be 1. Similarly, for example, the remainder mod 6 (i == 5) also dictates the remainder mod 2 (i == 1) and 3 (i == 2)

src/ai/ai.cpp
```diff
@@ -228,8 +228,9 @@ void update_ai_colony_starting(sys::state& state) {
 		states.push_back(sd);
 	}
 	// Shuffle the vector
+	auto rnd = rng::get_random(state, state.current_date.to_raw_value());
 	for(int i = (int) states.size() - 1; i >= 0; i--) {
-		int j = rng::get_random(state, state.current_date.to_raw_value()) % (i+1);
+		int j = rnd % states.size();
```
schombert: this is now the same number for the entire iteration, and so will swap the ith position with the same jth position for the entire loop, which is even less random

schombert:
You are getting close to a fisher-yates shuffle (https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle); you should probably just implement that. Also, if you look at the implementation of the random number generator, you will see that the current date is already one of the factors that goes into making the result, so you should probably pick a different seed value.

PR #2084, opened 2025-09-21
GFM fixes, trade/price adjustments and more rigorous needs satisfaction calculation

ineveraskedforthis:


src/economy/economy_trade_routes.cpp
```diff
@@ -135,6 +135,11 @@ void make_trade_volume_tooltip(
 		multiplier = -1.f;
 	}
 
+	int this_id = 0;
+	if(B == point_of_view) {
```
schombert, line 139: debug code left in by accident?

ineveraskedforthis: used later in `text::add_line(state, contents, "trade_route_volume_profit_score_export",`

PR #2091, opened 2025-09-27
Add gamerules & other small fixes and touchups

ZombieFreak115:
Adds gamerules, which can be toggled in-game. There are hardcoded gamerules which refer to base game logic (For now there is just one, which is whether or not spherelings can declare war on their overlord, but i plan on adding more later), and scripted ones  which mods can add. The current setting of any gamerule can also be checked with a trigger.
My motivation for adding these is that there wasen't a good way of changing gameplay settings on-the-go without creating a new scenario. 

Also did some minor fixups, such as:

-  moving the in-game chat messages to the "back" instead of the front, so that they won't block the view of other gameplay elements (which i found annoying).
- Sort the list of mod localization files by filename to ensure compatabillity with mods that use duplicate loc keys (a suprising amount).

- Changed a few places in which a "+ 1" was added to avoid DBZ problems into "std::max()" instead, due to it making more sense in those places to not add 1 to the total as it would affect the result.


src/culture/culture.cpp
```diff
@@ -699,7 +699,7 @@ void apply_invention(sys::state& state, dcon::nation_id target_nation, dcon::inv
 				++total;
 			}
 		}
-		nations::adjust_prestige(state, target_nation, p / float(total + 1));
```
schombert, line 702: This was correct the way it was before. For the first inventor, the total is 0 (no one else has invented it), so they get 100% prestige. For the second person, the total is 1, and they get  / 1 + 1 = 50% prestige, etc (your version would give the second person 100% prestige)

src/scripting/fif_triggers.cpp
```diff
@@ -3281,6 +3281,12 @@ TRIGGER_FUNCTION(tf_relation_from_province) {
 TRIGGER_FUNCTION(tf_check_variable) {
 	return "dup " + std::to_string(trigger::payload(tval[3]).natv_id.index()) + " >national_variable_id variables @ " + std::to_string(read_float_from_payload(tval + 1)) + " " + compare_values(tval[0]);
 }
+
+TRIGGER_FUNCTION(tf_check_gamerule) {
+	return "dup " + std::to_string(trigger::payload(tval[1]).gr_id.index()) + " >gamerule_id current_setting @ " + std::to_string(tval[2]) + " " + compare_values(tval[0]);
```
schombert, line 3286: you want just `std::to_string(trigger::payload(tval[1]).gr_id.index()) + " >gamerule_id current_setting @ " + std::to_string(tval[2]) + " " + compare_values(tval[0]);` because you don't need to copy the value on the top of the stack.

src/parsing/trigger_parsing.hpp
```diff
@@ -233,6 +233,44 @@ struct tr_check_variable {
...
+
+	}
+	void option(association_type assoc, std::string_view value, error_handler& err, int32_t line, trigger_building_context& context) {
+		auto option_it = context.outer_context.map_of_gamerule_options.find(std::string(value));
+		if(option_it == context.outer_context.map_of_gamerule_options.end()) {
+			err.accumulated_errors += "Gamerule option with name " + std::string(value) + " was referenced but dosen't exist (" + err.file_name + "), line " + std::to_string(line) + "\n";
+		} else {
+			option_id = option_it->second.option_id;
+		}
+		if(assoc != association_type::eq && assoc != association_type::eq_default && assoc != association_type::ne) {
```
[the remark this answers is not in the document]
schombert: Also, this won't work as you expect, because `association_to_trigger_code` (see the implementation of check_gamerule) turns `association_type::eq_default` into greater-than-or-equal-to. If you really want to handle only the matching/not matching case, you want `association_to_bool_code` and then in the implementation of the triggers explicitly compare the result with `==` and then call `compare_to_true` on that result (or use the `compare_values_eq` helper). See the `primary_culture` trigger as an example of that. @ZombieFreak115

PR #2099, opened 2025-10-05
UI cache and fps counter improvements

ineveraskedforthis:


src/gamestate/commands.cpp
```diff
@@ -6504,6 +6504,7 @@ bool can_perform_command(sys::state& state, payload& c) {
 bool execute_command(sys::state& state, payload& c) {
 	if(!can_perform_command(state, c))
 		return false;
+	state.tick_start_counter.store(state.tick_start_counter + 1, std::memory_order::release);
```
schombert, line 6507: I think you want to use `fetch_add` for these, and probably with the sequential consistency memory order (or, maybe not for the first, but definitely for the second -- the way I think of it is that release prevents code written below from being executed before the atomic point and release prevents code written above from being executed before, and seq_consist. prevents both directions. -- that's just a rule of thumb, not the full rules. Since the second one needs one direction and the first needs the other, I would given them both seq_consist (as well as their associated reads elsewhere). rather than looking up the full rules.

src/gamestate/system_state.cpp
```diff
@@ -508,6 +508,433 @@ int state::get_edit_y(){
...
+		export_volume.resize(state.world.nation_size());
+	}
+	if(import_volume.size() < state.world.nation_size()) {
+		import_volume.resize(state.world.nation_size());
+	}
+	if(consumption_volume.size() < state.world.nation_size()) {
+		consumption_volume.resize(state.world.nation_size());
+	}
+	if(production_volume.size() < state.world.nation_size()) {
+		production_volume.resize(state.world.nation_size());
```
schombert, line 527: but of course, since the game is running, in theory the number of nations could increase between now and the end of the function (although, in practice, we don't currently bring new nations into existence). To get around this, you would have to resize these arrays to the maximum number of objects that dcon can store, not just the current number.

ineveraskedforthis: Would moving them toward the end to ensure that at least it's safe to write current item into them work?

PR #2100, opened 2025-10-06
Protect resizes in cached vectors with mutex and locks

ineveraskedforthis:


src/gui/trade_dashboard.cpp
```diff
@@ -1754,15 +1754,18 @@ void  trade_dashboard_main_top_production_t::on_create(sys::state& state, layout
...
+	if(!size.has_value()) {
 		return;
 	}
 	values.clear();
 	size_t counter = 0;
-	while(counter < state.ui_cached_data.commodity_per_province.sorted_by_production.size() && values.size() < 8) {
-		auto pid = state.ui_cached_data.commodity_per_province.sorted_by_production[counter];
-		if(!main.nation_pov || state.world.province_get_nation_from_province_ownership(pid) == main.nation_pov) {
-			add_province_row_production(pid);
+	while(counter < size.value() && values.size() < 8) {
```
schombert, line 1763: and you can also write `size.value()` as `*size`, as if it was a pointer

PR #2101, opened 2025-10-07
Winter world textures and extra attrition

SneakBug8:


schombert:
This appears to make winter in the northern and summer hemispheres the same. It also appears to give winter siege attrition to all provinces regardless of climate, including tropical jungles and equatorial deserts.

PR #2102, opened 2025-10-08
Allow addition of on-map general icons with directions

SneakBug8:


src/parsing/defines.hpp
```diff
@@ -761,7 +761,7 @@
 	LUA_DEFINES_LIST_ELEMENT(alice_rebel_reduction_after_reoccupation, 7.0) \
 	LUA_DEFINES_LIST_ELEMENT(alice_can_add_constructable_cbs_as_wargoals, 1.0) \
 	LUA_DEFINES_LIST_ELEMENT(alice_can_goto_war_against_spherelord_default_setting, 1.0) \
-
+	LUA_DEFINES_LIST_ELEMENT(alice_render_on_map_generals, 0.0) \
```
schombert, line 764: why is this a mod dependent setting rather than a graphics option?

SneakBug8: A mod needs to add graphics for the component to work. I could make a runtime calculation whether a `map_general` element has been added, but I think that a define would be more appropriate in this situation

PR #2114, opened 2025-10-14
Gamerule improvements&hotfix

ZombieFreak115:
Adds persistent saving of gamerule settings. Whenever the player as either the host or in singleplayer changes a gamerule, the current gamerules will be saved as the default ones. When the player launches the game, or loads the startdate those gamerules will be loaded again. Proper saves will keep their own gamerule settings when loaded.

Added gamerules for Fog of war, the auto-peace when conceding mechanic and for partial retreats. 

Furthermore fix an oversight in socket_recv_command, and make sure it is keeping track of which part of a command (healer or payload) is being expected inbetween calls. Also minor fix to change_controller, as it was possible for uncolonized land to change controller from some effects (eg GFM)

### DIFF
--- a/assets/alice.gui
+++ b/assets/alice.gui
@@ -278,22 +278,6 @@ guiTypes = {
 				position = { 167 99 }
 				quadTextureSprite = "GFX_checkbox_default"
 			}
-			# Fog of War (Y/N)
-			instantTextBoxType = {
-				name = "fow_label"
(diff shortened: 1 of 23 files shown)


schombert:
Given player feedback, I would like "can reject" to be the default when the user hasn't touched the gamerules window yet, but I would still like partial retreats to default to off

PR #2134, opened 2025-11-01
Fixes to AI factory construction & extra info in factory construction UI

SneakBug8:


src/ai/ai.cpp
```diff
@@ -1396,11 +1396,34 @@ float estimate_balanced_composition_factor(sys::state& state, dcon::army_id a) {
...
+	// average army quality
+	float total_str = 0.f;
+	auto owner = state.world.army_control_get_controller(state.world.army_get_army_control(a));
+	for(const auto reg : regs) {
+		auto type = reg.get_regiment().get_type();
+		auto stats = state.world.nation_get_unit_stats(owner, type);
+		auto& atk = (stats.discipline_or_evasion > 0.0f) ? stats.attack_or_gun_power : state.military_definitions.unit_base_definitions[type].attack_or_gun_power;
+		auto& def = (stats.discipline_or_evasion > 0.0f) ? stats.defence_or_hull : state.military_definitions.unit_base_definitions[type].defence_or_hull;
+
+		total_str += atk * def * reg.get_regiment().get_strength();
```
schombert: and why is the units strength estimated by the product of attack and defense rather than the sum of them?

src/ai/ai.cpp
```diff
@@ -1396,11 +1396,34 @@ float estimate_balanced_composition_factor(sys::state& state, dcon::army_id a) {
...
+	auto regs = state.world.army_get_army_membership(a);
+	if(regs.begin() == regs.end())
+		return 0.0f;
+	// average army quality
+	float total_str = 0.f;
+	auto owner = state.world.army_control_get_controller(state.world.army_get_army_control(a));
+	for(const auto reg : regs) {
+		auto type = reg.get_regiment().get_type();
+		auto stats = state.world.nation_get_unit_stats(owner, type);
+		auto& atk = (stats.discipline_or_evasion > 0.0f) ? stats.attack_or_gun_power : state.military_definitions.unit_base_definitions[type].attack_or_gun_power;
```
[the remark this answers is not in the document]
schombert: if they haven't unlocked the unit, presumably it shouldn't appear in their armies, but I do understand its purpose now.

src/ai/ai.cpp
schombert: I would like to see some evidence of that, if so. I am inclined to sum them. Peter thought that min or max might be the right choice. Zombie also thinks the sum is more likely to work, although in the form of `(attack + defense + some constant) * strength`. So, if you want to defer to the masses, I suggest going with the sum. If not, I would like to see what exactly in the combat math makes the product a better estimate.

PR #2135, opened 2025-11-02
savefile now remembers which decisions were ignored, but only for SP. 

GeorgeBush-420:
Pretty much the title.


schombert:
First, if you need to serialize/de-serialize data and there isn't an existing suitable serialize/deserialize/serialize_size set, then you should add new functions. Secondly, if you need to serialize a variable number of things, you should first serialize the count of those things and then serialize/deserialize using that count. Using the remaining space in the file to figure out how many are left prevents anything else from ever being serialized after it in the file. Finally, all data stored in dcon should be serialized and de-serialized using the existing dcon facilities provided. You can read the documentation for that here: https://github.com/schombert/DataContainer/blob/master/serialization.md

GeorgeBush-420:
@schombert Let me know if there are further issues 
Edit: in diffs it will show changes that arent mine, i think thats because my branch is a couple commits behind main, i manually replaced my serialization.cpp / hpp files with the latest one from main which has zombie's recent additions in it.

src/gamestate/serialization.cpp
```diff
@@ -876,10 +882,46 @@ uint8_t const* read_mp_data(uint8_t const* ptr_in, uint8_t const* section_end, s
...
+
+    for(auto idx : hidden) {
+        if(idx < state.world.decision_size()) {
+            dcon::decision_id did{ dcon::decision_id::value_base_t(idx) };
+            state.world.decision_set_hide_notification(did, true);
+        }
+    }
+    state.game_state_updated.store(true, std::memory_order_release);
+    return ptr_in;
+}
```
ZombieFreak115: Since this is called on  `write_save_section` and `read_save_section` this will also be serialized over the network, and clients will effectively get a host' ignore settings tranferred to them when joining or resync'ing a game.

The way to fix this would be to either create a new DCON serializing tag and implement it so that it exempts it from the network seralization (i would refer to the dcon documentation that Schombert linked), or pick an easier solution such as storing the ignored decisions per-mod in a local file on the players' filesystem, and read data from it on startup (Look at `load_gamerule_settings()`, `save_gamerule_settings()` and `get_or_create_gamerules_directory()` for an example of how it is implemented for other constructs.

ZombieFreak115: My bad i missed some of the changes in the latest commit. In the latest commit it will still send the data over network, but it won't read it in MP. This may also be an issue because the serialization logic expects to read exactly as much data as `sizeof_save_section` returns, and the amount of data read will vary between MP and SP sessions. While another if statement on the network mode may work, it will still say that implementing it though DCON would be much cleaner and maintainable

PR #2136, opened 2025-11-02
Improve budget prediction and clean up some code

ineveraskedforthis:
- Add refund to administration and education spending
- Split predictions aggregator to its own file
- Rename `direct_satisfaction` to `actual_probability_to_`
- Rename `satisfaction` to `expected_probability_to_`
- Make their usage consistent around the files
- Add function to predict probability to buy commodity after adding new demand.
- Make consistent usage of controller of market in trade routes calculation
- Use owner as fallback to controller when controller is rebels in trade routes calculation

### DIFF
--- a/src/economy/construction.hpp
+++ b/src/economy/construction.hpp
@@ -27,6 +27,7 @@ economy::commodity_set calculate_factory_upgrade_goods_cost(
 	bool is_pop_project
 );
 
+float estimate_private_construction_spendings(sys::state& state, dcon::nation_id nid);
 void populate_construction_consumption(sys::state& state);
 
 struct unit_construction_data {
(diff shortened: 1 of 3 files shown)


src/ai/ai_economy.cpp
```diff
@@ -212,7 +213,7 @@ void retrieve_list_of_provinces_for_national_construction(sys::state& state, dco
 }
 
 inline bool province_has_available_workers(sys::state& state, dcon::province_id p) {
-	return state.world.province_get_labor_supply_sold(p, economy::labor::no_education) >= 0.95f;
+	return state.world.province_get_labor_supply_sold(p, economy::labor::no_education) <= 0.95f;
```
schombert: I think I have seen this comparison change back and forth in a recent PR; probably worth double checking

ineveraskedforthis: If this condition is false then AI drops attempt to build a factory. This condition was false if workers were able to sell less than 95% of their labor. Being able to sell less than 95% of produced labor means that unemployment of this type of labor is higher than 5%. Dropping attempt to build a factory because of a high unemployment was undesired behavior

PR #2203, opened 2025-12-08
Reorganize and expand LUA scripting APIs

SneakBug8:


src/gamestate/lua_alice_api.cpp
```diff
@@ -181,7 +187,219 @@ void remove_battle_end(const char function_name[]) {
...
+	lua_remove(state.lua_ui_environment, -1); // [
+
+	assert(lua_gettop(state.lua_ui_environment) == 0);
+}
+
 void set_state(sys::state* state_ptr) {
 	alice_state_ptr = state_ptr;
 }
+
+bool has_named_function(sys::state& state, const char function_name[]) {
```
schombert: The problem with the way this is written is that if the function doesn't exist it will ask the lua interpreter each time to try to find it, which is fairly inefficient. I suggest collecting the list of registered functions once at load time rather.

PR #2225, opened 2026-02-09
Adjust AI constructions algorithms

SneakBug8:


src/common_types/stackedcalculation.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <vector>
+#include <string>
+#include <iomanip>
+#include <sstream>
+#include <memory>
+
+namespace sys {
+
+	// US49 StackedCalculationWithExplanations allows uniting number calculations for backend with explanation tooltips for the UI
+	class StackedCalculationWithExplanations {
```
schombert: Unfortunately, there are some issues with this design. Turning a formula in the code into an instance of this type is a non-trivial performance sacrifice. I do understand the impulse behind it, however, and the idea can be rescued, even if this implementation isn't on the right track. What you want to do is develop a system of expression templates (the technique even has its own wikipedia page: https://en.wikipedia.org/wiki/Expression_templates ) that encodes the formula and any explanatory strings into a type at compile time, which in turn allows the calculation of that expression to be realized such that it can be compiled to the same end result as if it had been written out directly in the source code.

SneakBug8:
It builds now w no merge conflicts

src/common_types/stackedcalculation.hpp
```diff
@@ -0,0 +1,150 @@
+#pragma once
```
schombert: most of this file still doesn't conform to the capitalization and naming conventions of the project

src/ai/ai_economy.cpp
```diff
@@ -119,6 +122,121 @@ void update_factory_types_priority(sys::state& state) {
 	});
 }
 
+// US50AC1 Each factory type is evaluated by its profitability, payback time, and a number of synergies
+// Unprofitable factories have negative scores
+factory_evaluation_stack evaluate_factory_type(sys::state& state,
```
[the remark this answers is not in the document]
SneakBug8: The same function is called from the back end to then retrieve the number and from the UI to display full tooltip over the explanations, which is the point of the change in paradigm.

```cpp
auto factory_type_score = ai::evaluate_factory_type(state, n, mid, p, content, false, 0.3f, 0.5f, 200.f, rich_effect);

		for(auto line : factory_type_score.getSteps()) {
			if(line.operation == sys::Operation::ADD || line.operation == sys::Operation::SUBTRACT) {
				text::add_line(state, contents, line.explanation, text::variable_type::value, text::fp_one_place{ line.value }, 15);
			} else if(line.operation == sys::Operation::DIVIDE) {
				text::add_line(state, contents, line.explanation, text::variable_type::value, text::fp_one_place{ 1 / line.value }, 15);
			} else {
				text::add_line(state, contents, line.explanation, text::variable_type::value, text::fp_one_place{ line.value }, 15);
			}
		}
```

PR #2241, opened 2026-04-01
Mod general fixes & battle improvements

ZombieFreak115:
The PR aims to fix some mod compatabillity issues:

- Change parsing of triggers so that it is more strict and won't confuse invention triggers (invention_name = yes) with reform options (reform_category = reform_option) synaxes.

- Add support for the "land" strait type used by some mods. It acts like a "shortcut" between two land provinces, and cannot be blockaded by ships nor give a strait crossing penalty.

- Update events when the game starts, so that pending day1 events pop up appropriately. Also move the pending decision execution to the end of scenario generation to prevent an egde case.

- Increase max_issue_options constant from 6 to 8, as quite a few mods have 7 issue options per category.

- Change the way OOB files are loaded: Only the oob files referenced in a country history file will be read, instead of assuming that the first three letters of the filename is the TAG it is meant to be read for. Even vanilla uses this feature as Manchuria (MCK) uses china's oob file in the 1861 startdate.

- Fix edge case if soldiers were state-capital-only, and regiments would be deleted when merging pops to capital

- Slight change to which oob regiments are valid to accomondate mods.

The PR also adds some battle improvements:

- Change crossing penalties (river & strait) to be unit-specific roll modifiers. When a unit enters a battle on the attacking side, a crossing penalty may be placed on the specific units entering, which will persist until the unit leaves the battle.

- Same thing has been added to dig-in. Previously dig-in was battle-wide, but reinforcing with any units on the defending side would cancel the dig-in for the entire battle. Now the defending units get to keep their dig-in as a unit-specific attribute, and any additional units reinforcing won't cancel the dig-in of the existing units, but instead receive no dig-in bonus themselves. The dig-in bonus for units will be removed when they are moved to the reserve, or get outrolled as usual.

- Modify the unit deployment inside of a battle considerably. It no longer relies on unit types to make a decision on where to deploy. Instead it calculates a heuristic per unit of how effective they will be in the given spot (attacking vs defending and backline vs frontline). Reduce the heuristic value if the unit can't target anything, to move high-maneuver units to the egdes.

- Refactored land battle code in the process of doing the above.

- Adjust combat width of a battle when an army leaves or joins the battle, or any nation reseaches a technology or invention which changes combat width.

_ Adjust org and str damage equation abit to fix some mistakes. Change terrain modifiers to be a penalty on the attacker, rather than a bonus for the defender.

_ Add some defines relating to fort effects in battle and dice roll ranges for modders


src/gamestate/commands.cpp
```diff
@@ -6234,6 +6234,8 @@ void execute_notify_start_game(sys::state& state, dcon::nation_id source) {
 		state.yield_game_state_resetting_lock = false;
 		lock.unlock();
 		state.game_state_resetting_cv.notify_all();
+		// update up any events which were queued, so that they appear for players as soon as the game has started
+		event::update_future_events(state);
```
schombert: does this command also run in single player, or is this change also duplicated somewhere else? (If so, should we de-duplicate that code?)

PR #2251, opened 2026-04-11
More mod compat and bug fixes, add QoL Ui and refactor some command stuff

ZombieFreak115:
- Increase the max event options and max national mod to be able to cope events and modifiers from GFM and TGC.

_ Change factory bonuses to be an array instead of 3 diffrent fields. Up the max factory bonuses to 6 to add support for more mods' bonuses.

- Up the dcon max factory count from 20k to 100k, it is quite easy to reach the 20k cap (GFM reached it in 1898).

- Add legacy is_colonial_crisis trigger, which was missing previously.

- Move "update_events" call to after "ai_take_decisions" so that events popped as a result of a decision are popped immediately.

- Add ai_take_decisions call to "execute_start_game", so that ai make take decisions at game start.

- Move many calls which writes to "state.selected_armies" and "state.selected_navies" to execute on UI thread to prevent race conditions with UI reading it while the update thread is writing.

- Fix bug in crisis peacedeals in which it would always send a peacedeal with all wargoals checked, even though they weren't checked in the UI.

- Fix tax_efficiency national modifier being 100x less effective than it should be, as it used the same function as the "tax_eff" technology modifier.

- Add icons for colonization, which were previously missing.

- Add guard against invalid sea trade routes which become invalid due to a state being split up and losing its coast.

- Refactor some commands such as splitting units to be more uniform and AI being able to use the same functions as the player for checking if they are allowed to do it, but with a diffrent template parameter. Eventually this is the method i will use to try to centralize "can ai do this thing" check in ai code directly, with a vectorized version if needed.


src/common_types/constants_dcon.hpp
```diff
@@ -45,6 +45,7 @@ enum class province_building_type : uint8_t {
 	railroad, fort, naval_base, bank, university, last, factory, province_selector, province_immigrator
 };
 constexpr inline int32_t max_building_types = 5;
+constexpr inline size_t MAX_PRODUCTION_TYPE_BONUSES = 6;
```
schombert: generally, ALL_CAPS names are reserved for defines and macros; I know that I have been letting this slip a bit, but ...

In any case, you should touch base with peter regarding the performance of this change. The issue with increasing the number of bonuses is that the all involve running triggers as part of the econ update, and the econ update is already heavy.

src/common_types/container_types.hpp
```diff
@@ -306,6 +306,136 @@ class tagged_vector {
...
+		storage_size = 0;
+	}
+
+	constexpr fixed_size_vector(const std::initializer_list<data_type> initializer) {
+		assert(initializer.size() <= capacity);
+		storage_size = initializer.size();
+		std::copy(initializer.begin(), initializer.end(), data());
+	}
+
+	fixed_size_vector& operator=(fixed_size_vector<data_type, capacity> const& other) noexcept {
```
schombert, line 341: you probably need the appropriate move and copy constructors as well

src/common_types/container_types.hpp
```diff
@@ -306,6 +306,136 @@ class tagged_vector {
...
+		if(size() != capacity) {
+			_storage[size()] = v;
+			storage_size++;
+			return true;
+		} else {
+			return false;
+		}
+	}
+	data_type& back() {
+		return _storage.back();
```
schombert: these will return the end of the storage array, presumably, and not the last "live" element

src/common_types/container_types.hpp
```diff
@@ -306,6 +306,136 @@ class tagged_vector {
...
+	auto begin() {
+		return _storage.begin();
+	}
+	auto end() const {
+		return const_iterator(data(), size());
+	}
+	auto end() {
+		return iterator(data(), size());
+	}
+	auto rbegin() {
```
schombert: these functions are supposed to return *reverse* iterators; not only should rbegin return an iterator that points to the last item in the collection, but incrementing it should advance to the previous element, etc. It is perfectly fine to omit these functions if you don't want to implement.

src/common_types/container_types.hpp
```diff
@@ -306,6 +306,136 @@ class tagged_vector {
...
+	}
+	auto rend() {
+		return iterator(begin());
+	}
+	auto size() const {
+		return storage_size;
+	}
+	void resize(size_t new_size) {
+		if(new_size < size()) {
+			std::fill_n(_storage[new_size - 1], size() - new_size, data_type{ });
```
schombert: if new_size is 0, this would attempt to fill_n starting from position -1, which is probably not the intended behavior.

src/nations/nations.cpp
```diff
@@ -196,6 +196,10 @@ void recalculate_markets_distance(sys::state& state) {
 		state.world.market_set_max_throughput(markets, throughput);
 	});
 
+	static tagged_vector<fixed_bool_t, dcon::trade_route_id> to_delete;
+	to_delete.resize(state.world.trade_route_size());
+	std::fill(to_delete.begin(), to_delete.end(), false);
```
schombert, line 201: While this isn't inefficient, it doesn't feel very efficient either, because in our expected use case here we *probably* won't be deleting routes very often, but this is going to allocate a vector full of bools, clear then, and then eventually iterate over all the routes to check if one should be deleted. For the expected situation in which deletions are rare, my feeling is that a normal vector holding the ids to delete that you push to, and then sort prior to the deletion loop ( so you can call delete on just the is in the vector ) would be more efficient in the typical cases, since in most cases nothing will ever be added, so it will remain empty with no memory allocated, the sort will exit immediately, and the loop over its contents to do the deletions will exit immediately.

schombert: That said, I see that the routes to be deleted are discovered in parallel, which complicates things ...

schombert: Perhaps it would be better to delete them in advance outside the parallel work here. Finding the coastal "capitals" will fail for the sea routes in the case that state_is_coastal fails for either of them (that is cheaper than the capital finding function which compares populations). That may be a cheapish loop to run in reverse order on the trade routes. Then they could be deleted as found without having to make a separate storage container for them.

schombert: On third thought, maybe your first approach here is as good as it gets, since even doing the above you still end up looping over all the trade routes one additional time (in the pre-delete loop instead of the post delete loop).

... so ignore all my comments here I guess

src/common_types/constants_dcon.hpp
ZombieFreak115: I see, i changed all the ALL_CAPS constants that i remember i added to normal snake_case.

src/common_types/container_types.hpp
ZombieFreak115: Added appropriate constructors which should work

src/common_types/container_types.hpp
ZombieFreak115: Yup, i changed it

src/common_types/container_types.hpp
ZombieFreak115: I replaced them with proper reverse iterators from std::array, should work now i think

src/common_types/container_types.hpp
ZombieFreak115: Yep, fixed it

src/nations/nations.cpp
ZombieFreak115: Peter and i were also discussing an alternative solution which was to basically run a full trade route update on a daily tick if any provinces changed owners in the previous tick (via a flag being set on province ownership changing). I went with this solution instead cause i recall there was a reason to not update it that often (iirc its a very heavy update).

PR #2255, opened 2026-05-03
Technology Headers FIx

Sassy33893:
When switching to and from classic fonts, some text breaks which is usually fixed by just switching back to the original font you started with, but the headers in the technology tab get stuck as folder category. 

Screenshot 2026-05-03 143052

This PR just makes it so that the headers keep the proper name when switching to and from classic fonts.


schombert:
thanks for the PR. I suggest making yourself known on the discord so that you can get the appropriate tag and coordinate with other contributors.

PR #2260, opened 2026-05-11
Convert several serial loops to parallel

ineveraskedforthis:
Also adds entry point for profiling


src/economy/advanced_province_buildings.cpp
```diff
@@ -62,8 +62,8 @@ void reset_price(sys::state& state) {
 
 void match_supply_and_demand(sys::state& state) {
 	for(int32_t i = 0; i < list::total; i++) {
-		//state.world.execute_serial_over_province([&](auto pids) {
-		state.world.for_each_province([&](auto pids) {
+		//state.world.for_each_province([&](auto pids) {
+		state.world.execute_parallel_over_province([&](auto pids) {
```
schombert: is there a reason not to limit this to just land provinces?

src/economy/economy.cpp
```diff
@@ -2215,96 +2215,84 @@ void daily_update(sys::state& state, bool presimulation, float presimulation_sta
...
+				state.world.province_set_factory_bank(province, current_money + last_token_price * effective_output);
+			}
+		});
+	});
 
+	state.world.for_each_commodity([&](auto cid) {
+		auto base_output = state.world.commodity_get_rgo_amount(cid);
+		if(base_output == 0.f) return;
+
+		province::for_each_nation_owned_province_parallel_over_nation(state, [&](dcon::nation_id nation, dcon::province_id province) {
```
schombert: It would be better to interchange these loops (i.e. do the for each commodity loop inside the for each nation loop). This creates more work inside the parallel loop and means there just has to be one spawn and then join for the whole thing instead of one per commodity

src/economy/economy.cpp
```diff
@@ -2215,96 +2215,84 @@ void daily_update(sys::state& state, bool presimulation, float presimulation_sta
...
+				auto current_money = state.world.province_get_rgo_bank(province);
+				auto last_token_price = state.world.nation_get_subsidy_token_price(nation);
+				state.world.province_set_rgo_bank(province, current_money + last_token_price * effective_output);
+			}
+		});
+	});
 
-			break;
-		}
+	state.world.for_each_commodity([&](auto cid) {
```
schombert: The same goes for these nested loops as well.

src/economy/economy.cpp
```diff
@@ -4289,13 +4282,15 @@ void daily_update(sys::state& state, bool presimulation, float presimulation_sta
 	// # TAXES AND TARIFFS #
 	// #####################
 
-	for(auto n : state.world.in_nation) {
+
+	concurrency::parallel_for(int32_t(0), int32_t(state.world.nation_size()), [&](int32_t index) {
+		auto n = dcon::nation_id{ dcon::nation_id::value_base_t(index) };
 		/* advance construction */
 		advance_construction(state, n, spent_on_construction_buffer.get(n));
```
schombert, line 4302: Is advance construction safe to call in parallel? I know that it isn't for military constructions, because they can produce new entities. If this includes factory construction, it presumably isn't safe either.

src/economy/economy.cpp
schombert: I'm not sure if go bankrupt is safe to run in parallel. If not, the nations that are due to go bankrupt can be added to a a buffer and then make to go bankrupt in serial after the loop is complete.

schombert: After checking, at the very least it invokes "fire_fixed_event" which cannot be done in parallel.

PR #2272, opened 2026-05-24
Various fixes

ineveraskedforthis:
- Store updates to total satisfaction in an additional field due to lack of precision in 1 byte fields
- Fix issue when trade route was attempting to move around the continent due to state being at coast of two oceans by allowing sea routes to move around the original and target state instance
- Unify advanced buildings employment changes with production employment changes
- Allow land trade to path through 1 province lakes or seas and connect regions around them
- Improve handling of public and private education spending
- A few minor balance changes

### DIFF
--- a/src/economy/economy.cpp
+++ b/src/economy/economy.cpp
@@ -557,9 +557,7 @@ void initialize(sys::state& state) {
 
 	state.world.for_each_pop([&](dcon::pop_id p) {
 		auto fp = fatten(state.world, p);
-		pop_demographics::set_life_needs(state, p, 1.0f);
-		pop_demographics::set_everyday_needs(state, p, 0.1f);
-		pop_demographics::set_luxury_needs(state, p, 0.0f);
+		state.world.pop_set_satisfaction(p, 0.4f);
 		fp.set_savings(fp.get_size() * expected_savings);
 	});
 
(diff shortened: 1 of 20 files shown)


src/economy/economy.cpp
```diff
@@ -3763,13 +3763,30 @@ void daily_update(sys::state& state, bool presimulation, float presimulation_sta
...
+			//ln = ve::min(1.f, ve::max(0.f, ln + 2.f * (potential_ratio_life.get(ids) * (ln_satisfaction + satisfaction_from_subsistence.get(ids)) - 1.f) * pop_demographics::pop_u8_scaling));
+			//en = ve::min(1.f, ve::max(0.f, en + 2.f * (potential_ratio_everyday.get(ids) * en_satisfaction - 1.f) * pop_demographics::pop_u8_scaling));
+			//lx = ve::min(1.f, ve::max(0.f, lx + 2.f * (potential_ratio_luxury.get(ids) * lx_satisfaction - 1.f) * pop_demographics::pop_u8_scaling));
+
+			satisfaction = satisfaction * 3.f;
+			auto ln = ve::min(1.f, satisfaction);
+			satisfaction = satisfaction - 1.f;
+			auto en = ve::max(0.f, ve::min(1.f, satisfaction));
+			satisfaction = satisfaction - 1.f;
+			auto lx = ve::max(0.f, ve::min(1.f, satisfaction));
```
schombert: Since the three values are derived in such a direct way, the properties can be removed from the pop object completely and instead the 

```
float get_life_needs(sys::state const& state, dcon::pop_id p);
float get_everyday_needs(sys::state const& state, dcon::pop_id p);
float get_luxury_needs(sys::state const& state, dcon::pop_id p);
```

functions (from the demographics files) can have their implementation changed to load the single satisfaction value and compute the derived need directly. All references to those values should be going through those functions already, so dropping the dcon properties should thus be pretty painless.

PR #2286, opened 2026-07-21
insert_or_find_glyph() instead of sep. get_glyph/make_glyph

xinitrcn1:


src/graphics/stateless_render.cpp
```diff
@@ -84,11 +84,11 @@ void text_render(
 			pixel_x_off = trunc_pixel_x_off + 1.0f;
 		}
 
-		font_instance.make_glyph(uint16_t(glyphid), subpixel);
-		auto& gso = font_instance.get_glyph(uint16_t(glyphid), subpixel);
+		font_instance.insert_or_find_glyph(uint16_t(glyphid), subpixel);
```
schombert: ... but you didn't remove the double lookup ...

src/text/fonts.cpp
```diff
@@ -587,7 +584,7 @@ void font_at_size::make_glyph(uint16_t glyph_in, int32_t subpixel) {
 		auto err = FT_Get_Glyph(font_face->glyph, &g_result);
 		if(err != 0) {
 			glyph_positions.insert_or_assign((uint32_t(glyph_in) << 2) | uint32_t(subpixel & 3), gso);
-			return;
+			return std::nullopt;
```
schombert: If there is now an assigned slot, shouldn't that be returned? In which case the return type doesn't have to be an optional.

PR #2287, opened 2026-07-22
Rendering performance improvements

ineveraskedforthis:


ineveraskedforthis:
Main changes:
- Total rework of internal logic of unit counters to allow lazy and more conservative update of counters data.
- Borders generation is reworked to ensure consistent orientation.
- - Consistent orientation allows to simplify borders weaving logic and make it more robust
- Split borders into distinct groups: Province, state, national.
- - Allows usage of more performant OpenGL API related to indirect rendering. In the result map rendering time is reduced to nothing
- - During map update only data related to national borders has to be uploaded back to GPU, the rest of the border meshes are static
- Improve management of colonization labels by splitting them into groups. Group is rendered and updated only if there is some colonization going on locally.

PR #2295, opened 2026-08-16
Advanced consumption

ineveraskedforthis:
Adds alternative way to define population needs via consumption categories defined with lua.
Allows to implement substitution in mods.
Calls specific lua function during the start of scenario creation to allow intervention into the process of building the scenario. Currently could be used for creation and setting properties of commodities, factories and consumption categories. Example is provided in comments in related lua file.

### DIFF
--- a/src/gui/pop_budget_details.cpp
+++ b/src/gui/pop_budget_details.cpp
@@ -392,6 +392,15 @@ void  pop_budget_details_main_needs_t::update(sys::state& state, layout_window_e
 			add_consumption(cid, 2, cost, state.world.market_get_actual_probability_to_buy(market, cid));
 		}
 	});
+
+	state.world.for_each_consumption_category([&](auto cat) {
+		state.world.for_each_commodity([&](dcon::commodity_id cid) {
+			auto cost = economy::pops::estimate_pop_spending_category(state, cat, main.for_pop, cid);
+			if(cost > 0.001f) {
+				add_consumption(cid, (uint8_t)(cat.index()) + 3, cost, state.world.market_get_actual_probability_to_buy(market, cid));
+			}
+		});
+	});
 // END
 }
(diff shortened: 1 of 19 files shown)


src/parsing/parsers_declarations.hpp
```diff
@@ -412,6 +412,7 @@ struct scenario_building_context {
 	ankerl::unordered_dense::map<std::string, dcon::commodity_id> map_of_commodity_names;
 	ankerl::unordered_dense::map<std::string, dcon::factory_type_id> map_of_production_types;
 	ankerl::unordered_dense::map<std::string, dcon::factory_type_id> map_of_factory_names;
+	bool lua_factories;
```
schombert, line 415: This should probably be moved down to the both booleans below. It also probably needs to be given an initial value.

Edit: ah, I see that it gets set unconditionally, so it not having an initial value isn't a correctness issue.

src/gui/pop_budget_details.cpp
```diff
@@ -1915,6 +1932,8 @@ void pop_budget_details_consumption_name_t::on_update(sys::state& state) noexcep
 		modifier = text::produce_simple_string(state, "alice_pop_details_en_short");
 	} else if(consumption.category == 2) {
 		modifier = text::produce_simple_string(state, "alice_pop_details_lx_short");
+	} else {
+		modifier = std::to_string(consumption.category - 3);
```
schombert, line 1936: To make it possible for people to give names to the additional categories, maybe something like `text::produce_simple_string(state, "alice_pop_details_cat" +  std::to_string(consumption.category - 3) + "_short");` assuming that the category ids are stable and predictable. Otherwise, a category id to string name map could be stored at creation time and this name could be inserted instead of the number to form a localization key.

